PR #83, opened 2026-03-01
feat: channel creation modal (#44)

### LINKED ISSUE #44: Build channel creation modal
## Description
Create a modal for creating new channels within a server.

## Tasks
- [x] Create `CreateChannelModal` triggered from a "+" icon next to channel category headers
- [x] Form fields: channel name, channel type (Text/Voice), visibility (default: PRIVATE), topic (optional)
- [x] Channel name validation: lowercase, no spaces (auto-convert spaces to hyphens)
- [x] Preview: show how the channel name will appear (e.g., "#my-new-channel")
- [x] Submit calls mock `channelService.createChannel()` and adds to channel list
- [x] Loading state during creation, success toast on completion
- [x] Close modal on success or cancel

## Acceptance Criteria
- Modal opens from channel list
- Form validates channel name
- New channel appears in sidebar after creation
- Loading and success states work
- Modal is keyboard accessible

## Dependencies (blocked by)
- #18 — Channel creation calls createChannel service
- #21 — Channel creation modal triggered from ChannelSidebar

FardeenI:
## Summary

- **`CreateChannelModal`** — form dialog with channel name (auto-normalises to slug with live preview), type toggle (Text / Voice), visibility radio group (default: Private), and optional topic field; includes focus trap, Escape-to-close, inline validation, loading spinner, and success toast
- **`createChannelAction`** — `'use server'` action that validates enum values and slug format before delegating to `channelService.createChannel()`
- **`ChannelSidebar`** — adds a hover-revealed `+` button beside each category header (Text Channels / Voice Channels), only rendered for owner/admin users; clicking pre-selects the matching channel type in the modal
- **`HarmonyShell`** — introduces `localChannels` state (same optimistic pattern as `localMessages`) so the new channel appears in the sidebar immediately after creation without a page reload

## Test plan

- [x] Log in as an owner or admin — hover over "Text Channels" / "Voice Channels" headers and confirm the `+` button appears
- [x] Click `+` next to Text Channels → modal opens with Text pre-selected; click `+` next to Voice Channels → Voice pre-selected
- [x] Type a channel name — verify spaces convert to hyphens, uppercase becomes lowercase, special chars are stripped, and the slug preview updates live
- [x] Submit with an empty name — inline error "Channel name is required." appears
- [x] Submit a duplicate name — inline error "A channel with this name already exists." appears
- [x] Create a valid channel — loading spinner shows, success toast fires, modal closes, new channel appears in the sidebar immediately
- [x] Press Escape or click the backdrop → modal closes without creating a channel
- [x] Tab through all fields and buttons — focus stays trapped inside the modal; Shift+Tab wraps correctly
- [x] Log in as a regular member — confirm no `+` buttons are visible
- [x] View as guest — confirm no `+` buttons are visible

Closes #44


harmony-frontend/src/app/actions/createChannel.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ChannelType, ChannelVisibility, type Channel } from '@/types';
+import { createChannel } from '@/services/channelService';
+
+export interface CreateChannelInput {
+  serverId: string;
+  /** Normalised slug — must be [a-z0-9-], no leading/trailing hyphens. */
+  slug: string;
+  name: string;
+  type: ChannelType;
+  visibility: ChannelVisibility;
```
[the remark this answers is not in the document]
FardeenI: Addressed in 9dea47c.

harmony-frontend/src/components/channel/CreateChannelModal.tsx
```diff
@@ -0,0 +1,435 @@
...
+  onCreated,
+  onClose,
+}: CreateChannelModalProps) {
+  const titleId = useId();
+  const containerRef = useRef<HTMLDivElement>(null);
+  const nameInputRef = useRef<HTMLInputElement>(null);
+
+  const [rawName, setRawName] = useState('');
+  const [type, setType] = useState<ChannelType>(
+    defaultType === ChannelType.VOICE ? ChannelType.VOICE : ChannelType.TEXT,
```
AvanishKulkarni: 🟡 **[Medium] `ANNOUNCEMENT` type is silently coerced to `TEXT`** — The ternary `defaultType === ChannelType.VOICE ? ChannelType.VOICE : ChannelType.TEXT` maps any value that isn't `VOICE` (including `ANNOUNCEMENT`) to `TEXT`. If `ChannelSidebar` ever passes `ChannelType.ANNOUNCEMENT` as the default, the modal will open with the wrong type selected and the user won't know. Simplify to just `useState<ChannelType>(defaultType)` — it already has a default prop value of `ChannelType.TEXT`, so the fallback is handled at the call site.

harmony-frontend/src/components/channel/CreateChannelModal.tsx
```diff
@@ -0,0 +1,435 @@
...
+    setError(null);
+    setIsLoading(true);
+    try {
+      const newChannel = await createChannelAction({
+        serverId,
+        slug,
+        type,
+        visibility,
+        topic: topic.trim() || undefined,
+        position: existingChannels.filter(c => c.serverId === serverId).length,
```
AvanishKulkarni: 🔴 **[High] Race condition — `position` computed from stale client snapshot** — Two users creating a channel at the same time will both compute the same `position` value (e.g. both see 5 existing channels → both send `position: 5`). The resulting duplicate position must be resolved server-side. `position` should be assigned by the server (e.g. `SELECT MAX(position) + 1`) rather than accepted from the client. Note this is separate from @acabrera04's sanitization comment above — the integer is valid, but the value itself is wrong under concurrency.

harmony-frontend/src/components/channel/CreateChannelModal.tsx
```diff
@@ -0,0 +1,435 @@
...
+      const focusable = Array.from(
+        container.querySelectorAll<HTMLElement>(
+          'button:not([disabled]), input:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])',
+        ),
+      );
+      if (focusable.length === 0) return;
+      const first = focusable[0];
+      const last = focusable[focusable.length - 1];
+      if (e.shiftKey) {
+        if (document.activeElement === first) {
```
AvanishKulkarni: 🟡 **[Medium] Focus trap breaks when focus leaves and returns to the browser window** — Comparing `document.activeElement === first/last` only works while the window has focus. If the user Alt+Tabs away and back, `document.activeElement` may be `document.body`, causing the trap to stop working until they Tab into the modal again. A more robust approach is to call `e.relatedTarget` in a `focusout` listener, or use the native `<dialog>` element (which has a built-in focus trap in modern browsers) / a library like `@radix-ui/react-dialog`. At minimum, add a guard: if `document.activeElement` is not inside `container`, force-focus `first` before applying the wrap logic.

harmony-frontend/src/components/layout/HarmonyShell.tsx
```diff
@@ -162,6 +181,16 @@ export function HarmonyShell({
         isOpen={isSearchOpen}
         onClose={() => setIsSearchOpen(false)}
       />
+
+      {isCreateChannelOpen && (
+        <CreateChannelModal
+          serverId={currentServer.id}
+          existingChannels={localChannels}
+          defaultType={createChannelDefaultType}
+          onCreated={newChannel => setLocalChannels(prev => [...prev, newChannel])}
```
AvanishKulkarni: 🟢 **[Low] Newly created channel always appended to end regardless of type** — `setLocalChannels(prev => [...prev, newChannel])` places the new channel after all existing channels (text + voice), so a new text channel will appear below the voice channels section until the next full reload. Consider inserting at the correct position: e.g. place text channels before voice channels, or sort by `position`. This only affects the optimistic local state — the server list will be correct after `revalidatePath` resolves.

acabrera04 (approved):
## ✅ All comments resolved — approved

Verified all 4 issues from the previous review are fixed in `537fbc7`:

| Issue | Status |
|---|---|
| 🔴 Missing `revalidatePath` after channel creation | ✅ Fixed — `revalidatePath` called for `/channels`, `/c`, `/settings` after `createChannel()` resolves |
| 🟡 Unsanitized `serverId`, `topic`, `position` | ✅ Fixed — `serverId` validated as non-empty string; `topic` trimmed and clamped to 1024 chars; `position` removed from client input entirely and computed server-side via `getChannels(serverId).length` (bonus: eliminates the concurrent-create race condition) |
| 🟡 `+` button SVG in `ChannelSidebar` missing `aria-hidden` + `focusable="false"` | ✅ Fixed |
| 🟡 All 7 icon SVGs in `CreateChannelModal` missing `focusable="false"` | ✅ Fixed — all 7 SVGs (lines 37, 45, 53, 62, 71, 80, 88) now have both `aria-hidden="true"` and `focusable="false"` |

Also noted the bonus fix in `HarmonyShell.tsx`: new text/announcement channels are now inserted before voice channels in `localChannels`, so the sidebar grouping is correct immediately on creation.

No new issues found. Good to merge. 🚢

PR #131, opened 2026-03-06
Add server service & tRPC API (#99)

### LINKED ISSUE #99: Server Service & API
## Description
Implement server (guild) management service with tRPC endpoints.

## Tasks
- Server CRUD via tRPC: getServers, getServer(slug), createServer, updateServer, deleteServer
- Auto-slug generation from name
- Member count tracking
- Owner relationship to users
- is_public flag for server-level publicity
- Include minimum tests for server service happy paths

## Due Date
March 5, 2026

## Dependencies
Depends on: #95 #96

AvanishKulkarni:
## Summary
- Add `owner_id` FK to `servers` table with Prisma migration
- Implement `ServerRepository` (M-D1: CL-C-D1.3) with full CRUD, slug existence check, and member count increment
- Implement `ServerService` with auto-slug generation from name (unique collision handling) and owner-only authorization for update/delete
- Add `server` tRPC router with 5 procedures: `getServers`, `getServer(slug)`, `createServer`, `updateServer`, `deleteServer` — all with Zod input validation
- Wire server router into app router
- Add 10 tests (6 unit tests for slug generation, 4 integration tests for auth guards)

Closes #99

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npx jest tests/server.test.ts` — 10/10 pass
- [x] `npx jest tests/app.test.ts` — existing 5/5 still pass
- [ ] Run `docker compose up -d && npx prisma migrate deploy` to verify migration applies cleanly
- [ ] Verify full CRUD flow with authenticated session against running DB

### DIFF
--- a/harmony-backend/tests/visibility.service.test.ts
+++ b/harmony-backend/tests/visibility.service.test.ts
@@ -373,11 +373,14 @@ describe('visibilityService.setVisibility — VISIBILITY_CHANGED event', () => {
     await new Promise((resolve) => setTimeout(resolve, 200));
 
     expect(receivedPayloads.length).toBeGreaterThanOrEqual(1);
-    const payload = receivedPayloads[receivedPayloads.length - 1] as Record<string, unknown>;
-    expect(payload.channelId).toBe(textChannelId);
-    expect(payload.oldVisibility).toBe(ChannelVisibility.PRIVATE);
-    expect(payload.newVisibility).toBe(ChannelVisibility.PUBLIC_INDEXABLE);
-    expect(payload.actorId).toBe(userId);
+    const payload = (receivedPayloads as Record<string, unknown>[]).find(
+      (p) => p.channelId === textChannelId,
+    );
+    expect(payload).toBeDefined();
+    expect(payload!.channelId).toBe(textChannelId);
+    expect(payload!.oldVisibility).toBe(ChannelVisibility.PRIVATE);
+    expect(payload!.newVisibility).toBe(ChannelVisibility.PUBLIC_INDEXABLE);
+    expect(payload!.actorId).toBe(userId);
 
     unsubscribe();
   });


declanblanc (changes requested):
## Review

The PR closes issue #99 and covers all the required tasks (CRUD, auto-slug, member count, owner FK, is_public, tests). The core logic is sound, but there are several correctness issues and convention violations that need to be addressed before merge.

---

### Critical Issues

**1. Duplicate Prisma singleton**

The project already has `src/db/prisma.ts` (used by `channel.service.ts`). The PR introduces a second singleton at `src/lib/prisma.ts` with a slightly different implementation. Two separate `PrismaClient` instances will result in wasted connections and potential inconsistency. The `ServerRepository` should import from the existing `src/db/prisma.ts`:

```ts
// src/repositories/server.repo.ts
import { prisma } from '../db/prisma';
```

The new `src/lib/prisma.ts` file should be deleted.

**2. Wrong file location for `server.router.ts`**

The channel router lives at `src/trpc/routers/channel.router.ts` and is imported in `router.ts` as `'./routers/channel.router'`. The PR places `server.router.ts` at `src/trpc/server.router.ts` (one level up, outside the `routers/` subdirectory). This breaks the established convention.

Move to `src/trpc/routers/server.router.ts` and update the import in `router.ts` to `'./routers/server.router'`.

**3. Schema/migration mismatch — nullable vs. non-nullable**

The migration adds `owner_id` as nullable (`UUID` with no `NOT NULL`):
```sql
ALTER TABLE "servers" ADD COLUMN "owner_id" UUID;
```

But the Prisma schema declares `ownerId` as required (`String`, non-optional). This means:
- Existing rows will have `NULL` in the column, which Prisma will fail to deserialize since it expects a non-null string.
- The DB constraint allows future null inserts that bypass Prisma.

The migration needs a second step to add the `NOT NULL` constraint after backfilling, or the PR must document that this only targets a fresh database with no existing rows. At minimum, the comment "nullable first for safety" is misleading — the schema demands the value is always present, so the safety argument doesn't hold.

---

### Significant Issues

**4. Inconsistent service pattern**

The existing `channel.service.ts` exports a plain object (`export const channelService = { ... }`) and directly accesses `prisma`. The PR introduces a class-based service + a separate repository class layer, a pattern that doesn't exist anywhere else in the codebase. While classes aren't inherently wrong, introducing two different service patterns in the same project creates inconsistency. The server service should follow the same object-literal pattern as the channel service.

**5. `getServers` returns private servers to unauthenticated callers**

`getServers` is a `publicProcedure` that calls `repo.findAll()`, which returns every server with no visibility filter. An unauthenticated user can enumerate all private servers. Either:
- Apply an `isPublic: true` filter on the public procedure, or
- Change it to `authedProcedure`

**6. `updateServer` has the same slug race condition as the original (unfixed)**

The retry logic in `createServer` catches `P2002` and retries slug generation. However, `updateServer` also calls `generateUniqueSlug` when the name changes, and that path has no equivalent retry protection. The `P2002` on slug conflict could bubble up as an unhandled error.

---

### Minor Issues

**7. `getPublicInfo` and `incrementMemberCount` are untested dead code (from the service perspective)**

`getPublicInfo` and `incrementMemberCount` are exposed on `ServerRepository` but not called by `ServerService` or any router. If they're intended for future use by a REST controller, that should be noted. Otherwise they can be removed to keep the scope tight.

**8. `generateSlug` is `public` unnecessarily**

It's only public to enable direct unit testing. A better approach is to export it as a standalone utility function so the class interface isn't widened. This also enables testing without instantiating `ServerService`.

---

### Architecture Compliance

The architecture doc (§6.8, CL-C-D1.3) specifies `ServerRepository` should expose `findBySlug` and `getPublicInfo` only — both are present. The additional CRUD methods are required by Issue #99, so this is acceptable. The tRPC placement under `M-B1` and the use of `authedProcedure` for mutations are correct per spec.

PR #145, opened 2026-03-08
feat(messages): message service & API (#101)

### LINKED ISSUE #101: Message Service & API
## Description
Implement message service with tRPC endpoints, including cursor-based pagination and message pinning.

## Tasks
- Message CRUD via tRPC: getMessages(channelId, pagination), sendMessage, editMessage, deleteMessage (soft delete)
- Cursor-based pagination (20 per page default, configurable)
- Author snapshot embedding
- Attachment metadata support
- Message pinning: pinMessage, unpinMessage, getPinnedMessages(channelId) — add `pinned` boolean + `pinnedAt` timestamp to message schema
- Include minimum tests for message service happy paths

## Due Date
March 6, 2026

## Dependencies
Depends on: #95 #96

FardeenI:
## Summary
- Implements full message service & tRPC API for Issue #101
- Adds `pinned`/`pinnedAt` fields to the `messages` table via a new Prisma migration
- Adds `message:pin` action to the RBAC permission matrix (MODERATOR+)
- Syncs the Prisma client (`prisma generate`) — fixes pre-existing type errors in auth/permission/user services

### `prisma/schema.prisma` + migration `20260308000000_add_message_pinning`
- `pinned Boolean @default(false)` and `pinnedAt DateTime?` on `Message`
- Partial index `idx_messages_channel_pinned` for fast pinned-message lookups

### `src/services/message.service.ts`
| Operation | Notes |
|---|---|
| `getMessages` | Cursor-based pagination (default 20, max 100); Redis SWR cache |
| `sendMessage` | Optional attachment metadata; author snapshot via Prisma include |
| `editMessage` | Ownership enforced in service; sets `editedAt` |
| `deleteMessage` | Soft delete; cross-user delete checks `message:delete_any` in service |
| `pinMessage` | Sets `pinned=true` + `pinnedAt`; CONFLICT if already pinned |
| `unpinMessage` | Clears `pinned` + `pinnedAt`; CONFLICT if not pinned |
| `getPinnedMessages` | Returns pinned non-deleted messages, ordered `pinnedAt DESC` |

### `src/trpc/routers/message.router.ts`
All 7 procedures gated with `withPermission`:
- `getMessages` / `getPinnedMessages` → `message:read` (GUEST+)
- `sendMessage` → `message:create` (MEMBER+)
- `editMessage` → `message:update_own` (MEMBER+)
- `deleteMessage` → `message:delete_own` (MEMBER+; elevated in service for cross-user)
- `pinMessage` / `unpinMessage` → `message:pin` (MODERATOR+)

### `src/services/permission.service.ts`
- Added `message:pin` to `MessageAction`; granted to MODERATOR+

### `tests/message.service.test.ts`
- Happy-path coverage for all 7 operations
- Error cases: FORBIDDEN, NOT_FOUND, CONFLICT

## Test plan
- [ ] Run `npm test -- --testPathPattern=message.service` against a live Postgres instance to verify all happy paths pass
- [ ] Confirm `pinMessage` / `unpinMessage` round-trip correctly
- [ ] Confirm `deleteMessage` by a non-author MEMBER throws FORBIDDEN
- [ ] Confirm `deleteMessage` by a MODERATOR+ succeeds on another user's message
- [ ] Run `npm run lint` and `npx tsc --noEmit` (both clean on this branch)
- [ ] Apply migration `20260308000000_add_message_pinning` to staging DB before deploying

Closes #101


acabrera04 (changes requested):
The overall structure is solid — cursor pagination, SWR caching, RBAC gating, and test coverage are all well done. Two issues need to be resolved before merge; the cross-server bypass raised by @declanblanc is the top priority.

---

### HIGH — `BigInt` in attachment schema causes JSON serialization crash

`AttachmentInputSchema` uses `z.bigint()` and the Prisma `Attachment` model returns `sizeBytes: BigInt`. `JSON.stringify` throws on `BigInt` values, so **any `sendMessage` call with attachments and any `getMessages`/`getPinnedMessages` response containing attachments will crash at the tRPC serialization layer**.

Fix options:
- Accept `z.number().int().positive()` in the router input and cast to `BigInt` before the Prisma `create`, **or**
- Add a `superjson` transformer to the tRPC init to handle `BigInt` end-to-end.

---

### HIGH — Cache key is not scoped to server/user, leaking channel content across authorization boundaries

`getMessages` caches on `channel:msgs:{channelId}:cursor:{c}:limit:{n}`. There is no server or user scope. After the cross-server ownership check is added (per @declanblanc), a valid read by User A on Server X will populate a cache entry that User B — with `message:read` on Server Y but no access to that channel — can immediately hit. For private channels this is a content leak.

The cache key should include at minimum `serverId`.

---

### MEDIUM — `pinMessage` / `unpinMessage` still lack `serverId` validation in the service layer

`deleteMessage` already threads `serverId` through to the service for its `delete_any` check. `pinMessage` and `unpinMessage` accept only `messageId` and perform no channel-to-server ownership validation. Even after the fix for channel-based methods, these two will remain cross-server bypasses unless `serverId` is added to their signatures and validated.

---

### MEDIUM — Module-level `channelId` reassignment in tests breaks suite isolation

The `getMessages` describe block's inner `beforeAll` reassigns the module-level `channelId` variable. All subsequent suites (`editMessage`, `deleteMessage`, `pinMessage`) inherit the pagination channel instead of the one created in the outer `beforeAll`. Tests pass coincidentally but are not isolated. Scope `channelId` locally per describe block.

acabrera04 (approved):
All previously raised issues have been addressed — approving.

**BigInt serialization** ✓ — `AttachmentInputSchema` now uses `z.number().int().positive()`, cast to `BigInt` only at the Prisma layer. `sizeBytes` is excluded from `ATTACHMENT_SELECT` so BigInt never hits the tRPC serializer.

**Cache key scoping** ✓ — `msgCacheKey()` now includes `serverId` as a segment, and all `invalidatePattern` calls were updated to match.

**`pinMessage`/`unpinMessage` ownership validation** ✓ — Both methods now accept `serverId` and perform the channel-to-server check inside a `prisma.$transaction`, which also fixes the TOCTOU race noted by Copilot.

**Test suite isolation** ✓ — `getMessages` block now uses a locally scoped `paginationChannelId` instead of reassigning the module-level variable.

The rebase onto `main` is clean with no new issues introduced.

PR #161, opened 2026-03-09
feat: Visibility Audit Log Service (closes #106)

### LINKED ISSUE #106: Visibility Audit Log Service
## Description
Implement audit logging for all channel visibility changes per dev spec.

## Tasks
- AuditLogService: logVisibilityChange(channelId, actorId, oldValue, newValue, ipAddress)
- AuditLogRepository with pagination
- getVisibilityAuditLog(channelId, { limit, offset, startDate })
- Store IP address and user agent for compliance
- 7-year retention policy notation in schema
- Include minimum tests for audit log operations

## Due Date
March 10, 2026

## Dependencies
Depends on: #105

Aiden-Barrera:
## Summary
Implements the audit log service for channel visibility changes per issue #106.

### New: `src/services/auditLog.service.ts`
- `logVisibilityChange(input, tx?)` — creates a `VisibilityAuditLog` entry; accepts an optional Prisma transaction client so it can be called atomically inside `visibility.service`'s `$transaction`
- `getVisibilityAuditLog(channelId, { limit, offset, startDate })` — paginated query ordered newest-first, returns `{ entries, total }`
- Stores IP address and user agent for compliance

### Updated: `src/services/visibility.service.ts`
- Delegates audit log creation to `auditLogService.logVisibilityChange()`, passing `tx` to keep it in the same atomic transaction

### Updated: `prisma/schema.prisma`
- Added 7-year retention policy doc comment to the `VisibilityAuditLog` model

### New: `tests/auditLog.service.test.ts`
- 11 tests covering `logVisibilityChange` (correct fields, userAgent default, IPv6, transaction support) and `getVisibilityAuditLog` (ordering, total count, limit, offset, startDate filter, empty/unknown channel)

## Test Results
All 271 tests pass.

### DIFF
--- a/harmony-backend/tests/trpc.permission.middleware.test.ts
+++ b/harmony-backend/tests/trpc.permission.middleware.test.ts
@@ -28,7 +28,7 @@ const testRouter = router({
 const createCaller = createCallerFactory(testRouter);
 
 function callerAs(userId: string | null): ReturnType<typeof createCaller> {
-  const ctx: TRPCContext = { userId, ip: '127.0.0.1' };
+  const ctx: TRPCContext = { userId, ip: '127.0.0.1', userAgent: 'test-agent' };
   return createCaller(ctx);
 }
 


harmony-backend/src/services/auditLog.service.ts
```diff
@@ -0,0 +1,101 @@
...
+        action: 'VISIBILITY_CHANGED',
+        oldValue: input.oldValue as Prisma.InputJsonValue,
+        newValue: input.newValue as Prisma.InputJsonValue,
+        ipAddress: input.ipAddress,
+        userAgent: input.userAgent ?? '',
+      },
+    });
+  },
+
+  /**
```
acabrera04: **Medium:** `limit` has no upper bound. If `getVisibilityAuditLog` is ever called from an HTTP handler, a caller could pass an arbitrarily large value and exhaust memory. Consider capping it here defensively:

```ts
const { limit = 20, offset = 0, startDate } = options;
const safeLimist = Math.min(limit, 500); // or whatever the product max is
```

Even as an internal service method today, adding a guard here is cheap insurance.

Issue #113, opened 2026-03-04
Frontend Integration — Authentication


acabrera04:
## Description
Replace mock authService with real backend API calls.

## Tasks
- Replace mock authService with real API calls
- JWT token storage (httpOnly cookies)
- Auto-refresh token logic
- Update AuthContext to use real endpoints (login, register, logout, getCurrentUser, updateCurrentUser)
- Wire UserSettingsPage profile editing (displayName, status) + logout flow
- Redirect flows on 401

## Due Date
March 12, 2026

## Dependencies
Depends on: #97

acabrera04:
**Blocks:** #153 (Wire up pin button and implement pinned messages panel) — the pin panel makes authenticated tRPC calls that require real auth/JWT to be in place before it can be implemented.

PR #163, opened 2026-03-09
feat: Next.js Auth Middleware — Server-Side Route Protection (closes #119)

### LINKED ISSUE #71: Fix non-admin redirect spinner delay on /settings/* routes
## Summary
When a non-admin user navigates directly to `/settings/:serverSlug/:channelSlug`, they see a loading spinner for ~3–4 seconds before being redirected back to the channel view.

## Root Cause
The auth guard in `ChannelSettingsPage` is client-side only — it reads from `sessionStorage` via `AuthContext`. The component mounts, shows a spinner while `isLoading: true`, then fires the redirect once auth state resolves.

## Steps to Reproduce
1. Log in as `dave_42` (role: `member`)
2. Navigate directly to `/settings/harmony-hq/general`
3. Observe: ~3–4 second spinner before redirect to `/channels/harmony-hq/general`

## Expected
Immediate redirect (or at minimum a very brief flash).

## Fix Direction
Requires moving the auth check server-side. Options:
- Add server-side session (cookie/JWT) so the settings page server component can redirect before rendering
- Or use Next.js middleware to intercept `/settings/*` routes and redirect non-admins immediately

## Context
Found during dogfood testing of issue #29 (ChannelSettingsPage). Accepted as known limitation for initial ship — client-side auth is the pattern used throughout this app.

### LINKED ISSUE #119: Next.js Auth Middleware — Server-Side Route Protection
## Description
Add Next.js middleware for server-side auth, fixing the client-side redirect spinner (issue #71).

## Tasks
- Add Next.js middleware to protect /settings/* and /channels/* routes server-side
- Verify JWT from httpOnly cookie before rendering protected pages
- Redirect unauthenticated users to /auth/login immediately (no client-side spinner)
- Redirect non-admin users away from /settings/* routes
- Fixes GitHub issue #71 (3–4s spinner delay before redirect)

## Due Date
March 13, 2026

## Dependencies
Depends on: #97 #113

declanblanc:
## Summary

- **`src/middleware.ts`** — Next.js Edge middleware that intercepts every request to `/channels/*` and `/settings/*` before the page renders. Checks for the `auth_token` httpOnly cookie; if absent or malformed, immediately redirects to `/auth/login?returnUrl=<path>` (no more 3-4s client-side spinner, fixes #71). For `/settings/*` routes it also decodes the cookie payload and blocks non-admin/owner roles, redirecting them to `/channels`.
- **`src/app/actions/session.ts`** — two server actions (`setSessionCookie` / `clearSessionCookie`) that write/delete an httpOnly, SameSite=Lax `auth_token` cookie containing a base64url-encoded `{ sub, username, role }` payload. The cookie name and payload format are compatible with real JWT cookies (the middleware also handles standard JWT `xxxxx.yyyyy.zzzzz` format by decoding the middle segment), so the handoff when #113 ships real backend JWTs is seamless.
- **`src/context/AuthContext.tsx`** — `login`, `register`, and `logout` now call the session server actions to keep the cookie in sync. The mount-time session restore also refreshes the cookie in case it expired while `sessionStorage` still held valid user data.

## Design notes

- The middleware **does not verify the JWT signature** — that is intentionally left to the backend on every API/tRPC call. The cookie is used only for routing decisions (presence check + role decode).
- When issue #113 (real backend JWT auth) merges, `setSessionCookie` in `AuthContext` is replaced by the backend's `Set-Cookie` header on the `/api/auth/login` response. No middleware changes required.

## Test plan
- [ ] Visiting `/channels/...` while logged out redirects immediately to `/auth/login?returnUrl=/channels/...` with no spinner
- [ ] Visiting `/settings/...` while logged out redirects to `/auth/login`
- [ ] After login, `/channels/*` and `/settings/*` are accessible for admin users
- [ ] Non-admin user visiting `/settings/*` is redirected to `/channels`
- [ ] Logout clears cookie; subsequent navigation to `/channels` redirects to login
- [ ] Frontend `npm run lint` passes
- [ ] Frontend `npm run build` passes

Closes #119

### DIFF
--- a/harmony-frontend/src/app/actions/session.ts
+++ b/harmony-frontend/src/app/actions/session.ts
@@ -2,18 +2,8 @@
 
 import { cookies } from 'next/headers';
 
-/**
- * Cookie name used for auth session.
- * Must stay in sync with the local constant in middleware.ts.
- *
- * NOTE: When issue #113 (real backend JWT auth) is merged, this server action
- * will be replaced by the backend setting an httpOnly cookie directly via
- * Set-Cookie on the /api/auth/login response.
- *
- * Not exported — 'use server' files may only export async functions.
- * middleware.ts maintains its own copy of this constant.
- */
-const AUTH_COOKIE_NAME = 'auth_token';
+// Import from the shared constants module — can't export from 'use server' files.
+import { AUTH_COOKIE_NAME } from '@/lib/auth-constants';
 
 /** Maximum cookie age: 7 days (matches backend refresh token TTL). */
 const MAX_AGE_SECONDS = 7 * 24 * 60 * 60;
--- a/harmony-frontend/src/lib/auth-constants.ts
+++ b/harmony-frontend/src/lib/auth-constants.ts
@@ -1,0 +1,7 @@
+/**
+ * Shared auth constants used by both the Next.js middleware (Edge runtime)
+ * and server actions. Kept in a plain module (no 'use server' directive) so
+ * both consumers can import without restriction.
+ */
+
+export const AUTH_COOKIE_NAME = 'auth_token';
--- a/harmony-frontend/src/middleware.ts
+++ b/harmony-frontend/src/middleware.ts
@@ -6,7 +6,7 @@
  *
  * Protected routes:
  *   /channels/*  — require authentication
- *   /settings/*  — require authentication + admin/owner role
+ *   /settings/*  — require authentication (role enforcement is page-level)
  *
  * The middleware reads the `auth_token` httpOnly cookie set by the
  * `setSessionCookie` server action (or, after #113, by the backend directly).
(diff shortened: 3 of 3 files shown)


declanblanc:
**[AI Agent — In Progress]**

Rebased PR #163 onto current main to address all issues from the second-pass review:

1. **CRITICAL — Open redirect via `returnUrl`**: Already safe. The login page validates `returnUrl` with `raw.startsWith('/') && !raw.startsWith('//')` before following it ([`login/page.tsx`](../blob/main/harmony-frontend/src/app/auth/login/page.tsx#L28-L31)).

2. **HIGH — `ADMIN_ROLES.has(session.role)` always false**: Fixed in commit `69ebbf2` — removed the role check from middleware entirely. Settings pages enforce access page-level via `isAdmin(serverOwnerId)`.

3. **HIGH — AuthContext.tsx based on stale version**: Fixed in this rebase. The PR now diffs cleanly against main's current `AuthContext.tsx` (email-based, `getCurrentUser()` session restore). The only additions are the `setSessionCookie`/`clearSessionCookie` calls wired into `login`, `register`, `logout`, and session restore.

4. **MEDIUM — `AUTH_COOKIE_NAME` duplicated**: Fixed in commit `69ebbf2` — moved to `lib/auth-constants.ts`, imported by both middleware and session.ts.

5. **MEDIUM — `returnUrl` handled by login page**: Confirmed above — the login page already reads and honors `?returnUrl=` with the same-origin guard.

PR #164, opened 2026-03-09
fix: replace flaky setTimeout with deterministic promise in visibility event test (#147)

### LINKED ISSUE #147: Fix flaky setTimeout in visibility event emission test
## Description

PR #144 review identified a known deferral: the event emission test in `visibility.service.test.ts` uses `setTimeout(resolve, 200)` to wait for Redis Pub/Sub delivery. This is a timing-dependent assertion that could become flaky in CI under load.

## Context

From PR #144 approval review by @acabrera04:
> **MEDIUM-2:** `setTimeout(resolve, 200)` in the event emission test — still present, acceptable as a known limitation given Redis Pub/Sub delivery characteristics. Track as a follow-up if CI flakiness appears.

## Tasks

- Replace `setTimeout(resolve, 200)` with a deterministic wait (e.g., subscribe and await the message callback directly, or use a mock/spy on the Redis publish call)
- Ensure the test reliably passes without timing dependencies

## References

- PR #144: Implement ChannelVisibilityService (#105)
- File: `harmony-backend/tests/visibility.service.test.ts`

declanblanc:
## Summary

- Replaces `setTimeout(resolve, 200)` in the `VISIBILITY_CHANGED` event emission test with a `Promise` that resolves directly from the subscription callback when a matching payload arrives.
- Imports `VisibilityChangedPayload` type from `eventBus` to keep the new promise fully typed — no casts needed.
- All other test logic is unchanged; the assertions are identical.

**Why this is better:** The old approach waited an arbitrary 200ms hoping Redis Pub/Sub would deliver within that window. Under CI load this could fail. The new approach waits exactly as long as delivery actually takes — neither more nor less.

Closes #147

## Test plan
- [ ] Run `npm test` in `harmony-backend` — all visibility service tests pass, including the event emission test
- [ ] Confirm no `setTimeout` remains in the visibility test file

### DIFF
--- a/harmony-backend/tests/visibility.service.test.ts
+++ b/harmony-backend/tests/visibility.service.test.ts
@@ -12,7 +12,7 @@
 import { ChannelVisibility, ChannelType } from '@prisma/client';
 import { TRPCError } from '@trpc/server';
 import { visibilityService, SetVisibilityInput } from '../src/services/visibility.service';
-import { eventBus, EventChannels } from '../src/events/eventBus';
+import { eventBus, EventChannels, VisibilityChangedPayload } from '../src/events/eventBus';
 import { prisma } from '../src/db/prisma';
 
 let serverId: string;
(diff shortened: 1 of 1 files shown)


acabrera04 (approved):
This is a clean, well-reasoned fix. The deterministic Promise-based approach is strictly better than the 200ms arbitrary wait — it resolves as fast as the event actually arrives and fails with a meaningful error message if the event is never delivered. The timeout cleanup is correctly handled. The type import eliminates the unsafe casts that were required before.

No blocking issues. Approving.

PR #172, opened 2026-03-10
fix: suppress tRPC stack traces in non-development environments (#165)

### LINKED ISSUE #165: Security: tRPC error responses expose stack traces to clients
## Summary

tRPC error responses include full server-side stack traces in all environments, including what would be production. This leaks internal implementation details (file paths, library versions, source structure) to any client that receives an error.

## Reproduction

Any authenticated or unauthenticated tRPC call that triggers an error returns a response like:

```json
{
  "error": {
    "message": "UNAUTHORIZED",
    "code": -32001,
    "data": {
      "code": "UNAUTHORIZED",
      "httpStatus": 401,
      "stack": "TRPCError: UNAUTHORIZED\n    at <anonymous> (/Users/.../harmony-backend/src/trpc/init.ts:40:11)\n    ..."
    }
  }
}
```

The `stack` field reveals absolute file paths and internal source structure.

## Root Cause

`initTRPC.create()` in `src/trpc/init.ts` does not configure an `errorFormatter` to strip stack traces in production. tRPC includes stacks by default.

## Fix

Add an `errorFormatter` to `initTRPC.create()`:

```ts
const t = initTRPC.context<TRPCContext>().create({
  errorFormatter({ shape, error }) {
    return {
      ...shape,
      data: {
        ...shape.data,
        stack: process.env.NODE_ENV === 'development' ? error.stack : undefined,
      },
    };
  },
});
```

## Impact

- Exposes internal file paths and source structure to any API client
- Facilitates targeted attacks by revealing exact lines where auth/permission checks occur
- Low severity in development, high in production

## References

- `src/trpc/init.ts` — `initTRPC.create()` call (no errorFormatter)
- tRPC docs: https://trpc.io/docs/server/error-formatting

AvanishKulkarni:
## Summary

- Adds an `errorFormatter` to `initTRPC.create()` in `harmony-backend/src/trpc/init.ts` that strips the `stack` field from error responses unless `NODE_ENV === 'development'`
- Adds a test file (`tests/trpc.error-formatter.test.ts`) that verifies stack traces are omitted in production and present in development

## Root Cause

`initTRPC.context<TRPCContext>().create()` had no `errorFormatter`, so tRPC's default behavior included full server-side stack traces (with absolute file paths) in every error response.

## Fix

```ts
const t = initTRPC.context<TRPCContext>().create({
  errorFormatter({ shape }) {
    return {
      ...shape,
      data: {
        ...shape.data,
        stack: process.env.NODE_ENV === 'development' ? shape.data.stack : undefined,
      },
    };
  },
});
```

## Test plan

- [x] `npx jest tests/trpc.error-formatter.test.ts` — 2 tests pass (production suppresses stack, development includes stack)
- [x] `npx eslint src/trpc/init.ts tests/trpc.error-formatter.test.ts` — no errors
- [x] `npx tsc --noEmit` — no type errors

Closes #165

### DIFF
--- a/llm-logs/backend/03-09-pr172-fix-stack-trace.md
+++ b/llm-logs/backend/03-09-pr172-fix-stack-trace.md
@@ -1,0 +1,380 @@
+ ▐▛███▜▌   Claude Code v2.1.72
+▝▜█████▛▘  Sonnet 4.6 · Claude Pro
+  ▘▘ ▝▝    /workspace
+
+! gh auth login                                                                 
+  ⎿  ! First copy your one-time code: 816B-DD3C
+     Open this URL to continue in your web browser:
+     https://github.com/login/device
+     ✓ Authentication complete.
+     ✓ Logged in as AvanishKulkarni
+
+❯ Complete https://github.com/acabrera04/Harmony/issues/165 in a new branch     
+
+● Bash(gh issue view 165 --repo acabrera04/Harmony 2>&1)
+  ⎿  Error: Exit code 1
+     GraphQL: Projects (classic) is being deprecated in favor of the new 
+     Projects experience, see: 
+     https://github.blog/changelog/2024-05-23-sunset-notice-projects-classic/. 
+     (repository.issue.projectCards)
+
+● Bash(gh issue view 165 --repo acabrera04/Harmony --json
+      title,body,labels,comments 2>&1)
+  ⎿  {
+       "body": "## Summary\n\ntRPC error responses include full server-side
+      stack traces in all environments, including what would be production.
+     … +27 lines (ctrl+o to expand)
+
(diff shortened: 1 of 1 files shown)


declanblanc (approved):
The fix is correct and well-scoped.

The `errorFormatter` in `init.ts` properly overrides `shape.data.stack` with `undefined` in non-development environments. I checked the tRPC v11.11.0 source — `shape.data` is always a `DefaultErrorData` object (never null), so the spread and direct property access are safe.

The test file covers both the suppression (production) and inclusion (development) cases using supertest against the real Express app with Prisma mocked out, which gives good end-to-end coverage of the formatter behavior over the wire.

CI passes. LGTM.

PR #174, opened 2026-03-10
feat: add server member list endpoint and wire frontend (closes #173)

### LINKED ISSUE #173: Add server member list endpoint to backend tRPC router
## Description

`getServerMembers()` in `harmony-frontend/src/services/serverService.ts` currently returns `[]` with a TODO comment because the backend does not expose a member list endpoint.

The frontend uses `server.memberCount` for display counts as a workaround, but a proper member list (with usernames, avatars, online status) cannot be rendered until this endpoint exists.

## Tasks
- Add a `server.getMembers` tRPC procedure (or equivalent REST endpoint) that returns the list of `ServerMember` rows with embedded user profile data
- Wire up `getServerMembers(serverId)` in the frontend service to call the new endpoint
- Update the members sidebar / member count display to use real data

## Context
Tracked as a TODO in PR #150 (Frontend Integration — Servers & Channels).

declanblanc:
## Summary

- **`harmony-backend/src/services/server.service.ts`** — adds `getMembers(serverId)` which queries `ServerMember` rows with embedded user data (`id`, `username`, `displayName`, `avatarUrl`, `status`), sorted by role hierarchy (OWNER → ADMIN → MODERATOR → MEMBER → GUEST) then join date. Also exports the `ServerMemberWithUser` interface.
- **`harmony-backend/src/trpc/routers/server.router.ts`** — adds `server.getMembers` tRPC query, gated by `withPermission('server:read')` so only server members can call it.
- **`harmony-backend/tests/server.test.ts`** — adds service-level tests (sort order, empty server) and router-level tests (requires auth, FORBIDDEN for non-member).
- **`harmony-frontend/src/services/serverService.ts`** — replaces the `getServerMembers()` stub with a real `trpcQuery('server.getMembers', { serverId })` call. Maps backend `RoleType`/`UserStatus` (uppercase Prisma enums) to frontend lowercase types. Returns `[]` on error so guest views remain unaffected.
- **`harmony-frontend/src/components/channel/ChannelPageContent.tsx`** — calls `getServerMembers(server.id)` in parallel with `getMessages` instead of hardcoding an empty array.

## Design notes

- The PR is branched from `issue/114-frontend-integration` (PR #150) because `trpc-client.ts` and the real service layer only exist there. This PR should be merged after #150.
- `getServerMembers` gracefully catches errors and returns `[]` so guest views (`GuestChannelView`, which already uses `server.memberCount` for display) are unaffected when the user lacks an auth cookie.

## Test plan
- [ ] Backend: `server.getMembers` returns sorted member list with embedded user data
- [ ] Backend: `server.getMembers` returns 401 without auth header
- [ ] Backend: `server.getMembers` returns 403 for authenticated non-member
- [ ] Frontend: MembersSidebar renders real usernames/avatars for authenticated users
- [ ] Frontend: Guest channel view still works (error caught, falls back to `[]`)

Closes #173

### DIFF
--- a/harmony-backend/tests/server.test.ts
+++ b/harmony-backend/tests/server.test.ts
@@ -354,4 +354,39 @@ describe('server tRPC router', () => {
     if (user) await prismaLocal.user.delete({ where: { id: user.id } }).catch(() => {});
     await prismaLocal.$disconnect();
   });
+
+  it('server.getMembers returns member list for authenticated server member', async () => {
+    const prismaLocal = new PrismaClient();
+    const ts = Date.now();
+
+    const { accessToken } = await authService.register(
+      `getmembers-member-${ts}@example.com`,
+      `gm_member_${ts}`,
+      'password123',
+    );
+    const user = await prismaLocal.user.findUniqueOrThrow({
+      where: { email: `getmembers-member-${ts}@example.com` },
+    });
+    const server = await prismaLocal.server.create({
+      data: { name: `GM Router ${ts}`, slug: `gm-router-${ts}`, ownerId: user.id },
+    });
+    await prismaLocal.serverMember.create({
+      data: { userId: user.id, serverId: server.id, role: 'OWNER' },
+    });
+
+    const input = encodeURIComponent(JSON.stringify({ serverId: server.id }));
+    const res = await request(app)
+      .get(`/trpc/server.getMembers?input=${input}`)
+      .set('Authorization', `Bearer ${accessToken}`);
+
+    expect(res.status).toBe(200);
+    const members = res.body.result.data as { userId: string; role: string }[];
+    expect(Array.isArray(members)).toBe(true);
+    expect(members.length).toBeGreaterThanOrEqual(1);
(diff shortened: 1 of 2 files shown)


declanblanc:
**[AI Agent — Complete]**

Addressed both review items in commit \`ab5b6ce\`.

**MEDIUM — \`toFrontendMember\` bypasses \`ServerMemberWithUser\` type**
Added a local \`BackendServerMember\` interface that mirrors the backend's exported \`ServerMemberWithUser\` shape exactly. \`toFrontendMember\` now accepts \`BackendServerMember\` instead of \`Record<string, unknown>\`, and all field accesses are typed without any \`as\` casts. The \`getServerMembers\` tRPC query generic is also updated to \`BackendServerMember[]\`. A doc comment notes the local interface must be kept in sync with the backend source.

**LOW — No happy-path router test for an authenticated member**
Added \`server.getMembers returns member list for authenticated server member\` to the tRPC router test suite. The test registers a user, creates a server + \`ServerMember\` row via Prisma, calls the endpoint with a valid token, and asserts a 200 response with the member present in the result.

PR #177, opened 2026-03-10
fix: degrade gracefully when getChannels throws for non-member servers (#170)

### LINKED ISSUE #170: ChannelPageContent crashes if authenticated user is not a member of every public server
## Bug

`ChannelPageContent` (`src/components/channel/ChannelPageContent.tsx`) calls `getChannels` for **every public server** to build the cross-server navigation sidebar:

```ts
const allChannels = (
  await Promise.all(
    servers.map(s =>
      s.id === server.id ? Promise.resolve(serverChannels) : getChannels(s.id)
    )
  )
).flat();
```

`getChannels` calls the `channel.getChannels` tRPC procedure, which uses `withPermission('server:read')`. If the authenticated user is not a member of any one of the public servers, the call throws `FORBIDDEN`, `Promise.all` rejects, and the entire page renders "Something went wrong."

## Reproduction

1. Register a new user
2. Add them as a member of one public server but not others
3. Navigate to `/channels/<that-server>/<channel>`
4. Observe: error boundary shown because `getChannels` throws for the other servers

## Impact

Any real user who hasn't joined all public servers will see a crash on every channel page. In practice this means the `/channels/` authenticated route is unusable for anyone except users who are members of every single public server.

## Fix Options

- **Catch and swallow** per-server `getChannels` failures (return `[]` for servers the user can't access) so the sidebar degrades gracefully instead of crashing
- **Only fetch channels for servers the user is a member of** — filter `servers` by membership before the `Promise.all`
- **Remove the cross-server `allChannels` fetch** and build sidebar navigation another way (e.g. per-server lazy loading)

Discovered during integration testing of PR #150.

acabrera04:
## Summary
- `ChannelPageContent` called `getChannels` for every public server to build the cross-server sidebar
- If the authenticated user was not a member of any server, that call threw `FORBIDDEN`, causing `Promise.all` to reject and the entire page to render the error boundary ("Something went wrong")
- Added `.catch(() => [])` per-server so failures degrade gracefully — the sidebar simply omits channels for inaccessible servers instead of crashing

## Change

```diff
- servers.map(s => (s.id === server.id ? Promise.resolve(serverChannels) : getChannels(s.id)))
+ servers.map(s =>
+   s.id === server.id ? Promise.resolve(serverChannels) : getChannels(s.id).catch(() => [])
+ )
```

## Test plan
- [x] Manually injected a `FORBIDDEN` throw in `getChannels` for `server-002` to simulate a non-member scenario
- [x] Verified channel page at `/channels/harmony-hq/general` renders successfully (no error boundary) despite the throw
- [x] Reverted the test injection; only the one-line fix is committed
- [x] ESLint passes on changed file

Closes #170

### DIFF
--- a/harmony-frontend/src/components/channel/ChannelPageContent.tsx
+++ b/harmony-frontend/src/components/channel/ChannelPageContent.tsx
@@ -37,10 +37,11 @@ export async function ChannelPageContent({
   ).flat();
 
   // Service returns newest-first; reverse for chronological display
-  const { messages } = await getMessages(channel.id);
+  const [{ messages }, members] = await Promise.all([
+    getMessages(channel.id),
+    getServerMembers(server.id),
+  ]);
   const sortedMessages = [...messages].reverse();
-
-  const members = await getServerMembers(server.id);
 
   const shell = (
     <HarmonyShell


declanblanc (approved):
## Guideline Checklist

- [x] **1. Logic over Style** — Both changes are correct. The `.catch(() => [])` prevents a `FORBIDDEN` rejection from propagating through `Promise.all` to the error boundary. Parallelising `getMessages` and `getServerMembers` is safe since the two calls are independent.
- [x] **2. Security First** — No new surface area; swallowing FORBIDDEN on a sidebar navigation call is appropriate (the error carries no sensitive data to hide).
- [x] **3. Architectural Alignment** — Consistent with how `getServerMembers` already degrades gracefully in the service layer.
- [x] **4. Issue Completion** — Fix directly addresses the FORBIDDEN crash described in #170.
- [x] **5. No Nitpicking** — No style or formatting comments.

---

Clean, minimal fix. The `.catch(() => [])` is the right scope — applied only to the cross-server calls, not to `serverChannels` (the current server's channels, which must succeed). Parallelising the two independent awaits is a welcome improvement.

PR #182, opened 2026-03-11
AGENTS.md and PR_REVIEW.md tweaks

declanblanc:
# Please feel free to nit-pick the hell out of this type of PR 😁

Instruct LLM to:
- Justify design choices via comments
- Use `agents-review` if it is available


Aiden-Barrera (approved):
LGTM 🚀

My only concern would be the token usage from the skill but that can only be found from testing the skill.

Issue #111, opened 2026-03-04
Event Bus — Redis Pub/Sub for Cross-Service Events


acabrera04:
## Description
Implement event-driven communication between services using Redis Pub/Sub.

## Tasks
- VISIBILITY_CHANGED event publishing and subscribing
- MESSAGE_CREATED / MESSAGE_EDITED / MESSAGE_DELETED events
- Cache invalidation triggered by events
- Decouple services via event-driven architecture
- Include minimum tests for event publish/subscribe

## Due Date
March 11, 2026

## Dependencies
Depends on: #109

Issue #100, opened 2026-03-04
Channel Service & API


acabrera04:
## Description
Implement channel management service with tRPC endpoints, including voice channel type support.

## Tasks
- Channel CRUD via tRPC: getChannels(serverId), getChannel(slug), createChannel, updateChannel, deleteChannel
- ChannelType enum support (TEXT, VOICE, ANNOUNCEMENT) — filter and group by type
- Visibility enum (PUBLIC_INDEXABLE, PUBLIC_NO_INDEX, PRIVATE)
- Enforce VOICE channels cannot be PUBLIC_INDEXABLE (must be PUBLIC_NO_INDEX or PRIVATE)
- Position ordering, slug uniqueness per server
- Default channel creation on server create
- Include minimum tests for channel service happy paths

## Due Date
March 6, 2026

## Dependencies
Depends on: #95 #96

Issue #97, opened 2026-03-04
Authentication System — JWT Register/Login/Logout


acabrera04:
## Description
Implement JWT-based authentication with register, login, and logout endpoints.

## Tasks
- POST /api/auth/register, POST /api/auth/login, POST /api/auth/logout
- JWT token generation + refresh tokens
- bcrypt password hashing
- Auth middleware for protected routes
- Zod input validation
- Include minimum tests for auth happy paths

## Due Date
March 5, 2026

## Dependencies
Depends on: #94 #95 #96

PR #193, opened 2026-03-15
feat: Twilio Voice Service — real-time audio for voice channels (#121)

### LINKED ISSUE #121: Twilio Voice Service — Real-Time Audio for Voice Channels
## Description
Integrate Twilio Programmable Video SDK to enable real-time audio in voice channels.

## Tasks
- Integrate Twilio Programmable Video SDK (group rooms = voice channels)
- POST /voice/token — generate Twilio access token for authenticated user + specific room
- Room lifecycle: auto-create Twilio room when first user joins a voice channel, destroy when empty
- Voice state tracking in Redis: who's in which room, muted/deafened status
- Publish events via Event Bus ( #111): `user-joined-voice`, `user-left-voice`, `voice-state-changed`
- GET /voice/participants/:channelId — list users currently in a voice channel
- Mock Twilio service for local dev/testing (no real Twilio API calls needed locally)
- Include minimum tests for token generation and room lifecycle

## Due Date
March 12, 2026

## Dependencies
Depends on: #97 #100 #111

Aiden-Barrera:
## Summary

- **`voice.service.ts`** — Redis-backed participant state (join/leave/updateState/getParticipants), Twilio VideoGrant token generation scoped to the channel room, room lifecycle (auto-create on first join, destroy when empty), EventBus publishing for `USER_JOINED_VOICE`, `USER_LEFT_VOICE`, `VOICE_STATE_CHANGED`
- **`voice.router.ts`** — tRPC procedures `join`, `leave`, `updateState`, `getParticipants`, all gated by `withPermission('channel:read')` so only server members can obtain tokens or enumerate participants
- **`eventTypes.ts`** — three new voice event channels + typed payload interfaces
- **`tests/voice.service.test.ts`** — 14 tests covering all service methods plus edge cases: double-join idempotency, graceful leave-without-join, mute/deafen state updates, multi-participant channels
- Mock mode auto-enabled when `TWILIO_*` credentials are absent (local dev / CI — no real API calls needed)

Closes #121. Depends on: #97, #100, #111 (all closed).

## Security hardening (post security review)

- All four voice procedures require `serverId` + `withPermission('channel:read')` — prevents unauthenticated or non-member users from obtaining Twilio tokens or reading private channel participant lists (broken access control fix)
- `updateState` adds a second Redis `sismember` guard — users not actively in the channel cannot write voice state events
- Fixed Twilio REST client constructor in `destroyTwilioRoom` — was passing `TWILIO_ACCOUNT_SID` where `TWILIO_API_KEY` is required (would have caused 401s in production and stale room accumulation)
- Redis key segments sanitized via `sanitizeSegment()` — strips non-alphanumeric chars from userId/channelId before key construction
- Mock token is now an opaque `'mock-token'` placeholder — no internal user/channel IDs embedded in client-visible tokens
- `getParticipants` filters null/empty hashes — stale set members with expired Redis hashes are excluded from the response

## Test plan

- [ ] Start Redis locally (`docker compose up -d redis`)
- [ ] Set `TWILIO_MOCK=true` in `.env.local` (or omit Twilio credentials)
- [ ] Run `npm test -- tests/voice.service.test.ts` — all 14 tests should pass
- [ ] `npm run lint && npm run build` — 0 errors
- [ ] Verify non-member cannot call `voice.join` (should get `FORBIDDEN`)
- [ ] Verify authenticated member can join, see participants, mute themselves, then leave

### DIFF
--- a/harmony-backend/src/trpc/routers/voice.router.ts
+++ b/harmony-backend/src/trpc/routers/voice.router.ts
@@ -52,6 +52,15 @@ export const voiceRouter = router({
     .input(z.object({ serverId: z.string().uuid(), channelId: z.string().uuid() }))
     .mutation(async ({ input, ctx }) => {
       await requireVoiceChannel(input.channelId, input.serverId);
+
+      // Guard against spurious USER_LEFT_VOICE events from non-participants.
+      // Callers who are not in the channel get an idempotent no-op rather than
+      // triggering a leave event or unnecessary Twilio room destruction.
+      const isMember = await redis.sismember(participantsKey(input.channelId), ctx.userId);
+      if (!isMember) {
+        return { success: true as const };
+      }
+
       await voiceService.leave(ctx.userId, input.channelId);
       return { success: true as const };
     }),


declanblanc (changes requested):
## PR Review Checklist

- [x] **1. Logic over Style** — Redis pipeline usage for join is correct; leave correctly sequences srem/del before the scard empty-check; `sanitizeSegment` prevents key collisions; `getParticipants` correctly filters null/empty hashes from stale set members.
- [ ] **2. Security First** — One issue identified: `updateState` in the router constructs the Redis key inline without `sanitizeSegment`, bypassing the sanitization applied in the service layer's own key helpers.
- [x] **3. Architectural Alignment** — Voice operations are authenticated tRPC procedures (correct per the tRPC/REST split). EventBus patterns mirror existing message/channel services. Redis as source of truth for transient voice state is consistent with the architecture.
- [x] **4. Issue Completion** — Issue #121 requires: Twilio Programmable Video integration, token endpoint, room lifecycle, Redis state tracking, EventBus publishing, participant list endpoint, mock service, and minimum tests. All present.
- [x] **6. Avoid Repetition** — First human review on this PR (prior reviews are bot-only).
- [x] **7. Iterative Reviews** — N/A (first human review).
- [ ] **8. Prevent CI Failures** — The `updateState` procedure in `voice.router.ts` constructs the Redis participant key inline without using `sanitizeSegment`, diverging from the pattern applied consistently in `voice.service.ts`. If `channelId` contains characters stripped by `sanitizeSegment` (e.g., from a hand-crafted request), the router's `sismember` check looks up a different key than the service's `sadd`, causing the guard to always return 0 and the `FORBIDDEN` error to fire for legitimate participants.

---

## Review

### Required Change — `updateState` router constructs participant key without `sanitizeSegment`

In `voice.router.ts`, the participant key used for the membership guard is built inline:

```ts
// voice.router.ts — updateState
const pKey = `voice:channel:${input.channelId}:participants`;
const isMember = await redis.sismember(pKey, ctx.userId);
```

In `voice.service.ts`, the same key is built via `participantsKey`, which applies `sanitizeSegment`:

```ts
// voice.service.ts
function participantsKey(channelId: string): string {
  return `voice:channel:${sanitizeSegment(channelId)}:participants`;
}
```

`channelId` is validated as a UUID by Zod (`.uuid()`), so the practical risk of a mismatch is low in the happy path. However, `sanitizeSegment` strips hyphens from UUIDs (`-` is outside `[a-zA-Z0-9]`... wait — actually hyphens are included in the allowed set `[a-zA-Z0-9\-_]`), so UUIDs pass through unchanged. Even so, the inline key construction is a latent consistency bug: if `sanitizeSegment`'s character set changes in the future, or if a non-UUID channelId ever reaches this path, the router's guard and the service's set will diverge silently.

**Fix:** Export `participantsKey` from `voice.service.ts` (or move the key helpers to a shared module) and use it in the router, eliminating the duplicated key construction logic.

PR #196, opened 2026-03-15
feat: attachment service & file storage (#112)

### LINKED ISSUE #112: Attachment Service & File Storage
## Description
Implement file attachment handling with S3-compatible storage interface.

## Tasks
- Attachment metadata CRUD (create, list by message)
- File upload endpoint using S3-compatible storage interface (use local filesystem or MinIO for dev, swaps to real S3 on AWS)
- Content-type and size validation
- URL generation for serving attachments
- Wire into message service for attachment embedding
- Include minimum tests for attachment upload/retrieval

## Due Date
March 11, 2026

## Dependencies
Depends on: #101

FardeenI:
## Summary
- Implements file upload handling with a `StorageProvider` abstraction (`LocalStorageProvider` for dev, env-swappable to S3)
- Adds `POST /api/attachments/upload` (multipart REST, auth required) with content-type whitelist and 25 MB size cap
- Adds `GET /api/attachments/files/:filename` for local dev file serving
- Adds `attachment.listByMessage` tRPC procedure (requires `message:read`)
- 14 tests passing (5 router integration + 9 service unit)

## How it fits together
Upload flow: `POST /api/attachments/upload` → returns `{ url, filename, contentType, sizeBytes }` → client passes into existing `message.sendMessage` attachments array. The message service creates the `Attachment` row atomically with the message (already implemented in #101).

Switching to S3 in production: set `STORAGE_PROVIDER=s3` and add `S3StorageProvider` — no changes needed to service or route logic.

Closes #112

## Test plan
- [ ] `npx jest tests/attachment.router.test.ts` — 5 passing (no DB required)
- [ ] `npx jest tests/attachment.service.test.ts` — 9 passing (requires DB)
- [ ] `npx tsc --noEmit` — no type errors
- [ ] `npx eslint src/lib/storage src/services/attachment.service.ts src/routes/attachment.router.ts src/trpc/routers/attachment.router.ts` — no warnings

### DIFF
--- a/harmony-backend/.env.example
+++ b/harmony-backend/.env.example
@@ -26,3 +26,10 @@ JWT_REFRESH_SECRET=change-me-refresh-secret
 JWT_ACCESS_EXPIRES_IN=15m
 # Refresh token TTL in days
 JWT_REFRESH_EXPIRES_DAYS=7
+
+# Twilio Voice (Programmable Video) — set these for real voice; omit for mock mode
+TWILIO_ACCOUNT_SID=ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+TWILIO_API_KEY=SKxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+TWILIO_API_SECRET=your-api-secret
+# Set to 'true' to use mock Twilio (no real API calls) — auto-enabled when credentials missing
+TWILIO_MOCK=false
--- a/harmony-backend/package-lock.json
+++ b/harmony-backend/package-lock.json
@@ -20,6 +20,7 @@
         "jsonwebtoken": "^9.0.3",
         "multer": "^2.1.1",
         "serverless-http": "^3.2.0",
+        "twilio": "^5.13.0",
         "zod": "^3.24.2"
       },
       "devDependencies": {
@@ -2501,6 +2502,18 @@
         "acorn": "^6.0.0 || ^7.0.0 || ^8.0.0"
       }
     },
+    "node_modules/agent-base": {
+      "version": "6.0.2",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-6.0.2.tgz",
+      "integrity": "sha512-RZNwNclF7+MS/8bDg70amg32dyeZGZxiDuQmZxKLAlQjr3jGyLx+4Kkk58UO7D2QdgFIQCovuSuZESne6RG6XQ==",
+      "license": "MIT",
+      "dependencies": {
+        "debug": "4"
+      },
+      "engines": {
+        "node": ">= 6.0.0"
+      }
+    },
     "node_modules/ajv": {
       "version": "6.14.0",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.14.0.tgz",
@@ -2604,8 +2617,18 @@
       "version": "0.4.0",
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
       "integrity": "sha512-Oei9OH4tRh0YqU3GxhX79dM/mwVgvbZJaSNaRk+bshkj0S5cfHcgYakreBjrHwatXKbz+IoIdYLxrKim2MjW0Q==",
-      "dev": true,
       "license": "MIT"
+    },
+    "node_modules/axios": {
+      "version": "1.13.6",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.13.6.tgz",
+      "integrity": "sha512-ChTCHMouEe2kn713WHbQGcuYrr6fXTBiu460OTwWrWob16g1bXn4vtz07Ope7ewMozJAnEquLk5lWQWtBig9DQ==",
+      "license": "MIT",
+      "dependencies": {
+        "follow-redirects": "^1.15.11",
+        "form-data": "^4.0.5",
+        "proxy-from-env": "^1.1.0"
+      }
     },
     "node_modules/babel-jest": {
       "version": "29.7.0",
@@ -3105,7 +3128,6 @@
       "version": "1.0.8",
       "resolved": "https://registry.npmjs.org/combined-stream/-/combined-stream-1.0.8.tgz",
       "integrity": "sha512-FQN4MRfuJeHf7cBbBMJFXhKSDq+2kAArBlmRBvcvFE5BB1HZKXtSFASDhdlz9zOYwxh8lDdnvmMOe/+5cdoEdg==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "delayed-stream": "~1.0.0"
@@ -3250,6 +3272,12 @@
         "node": ">= 8"
       }
     },
+    "node_modules/dayjs": {
+      "version": "1.11.20",
+      "resolved": "https://registry.npmjs.org/dayjs/-/dayjs-1.11.20.tgz",
+      "integrity": "sha512-YbwwqR/uYpeoP4pu043q+LTDLFBLApUP6VxRihdfNTqu4ubqMlGDLd6ErXhEgsyvY0K6nCs7nggYumAN+9uEuQ==",
+      "license": "MIT"
+    },
     "node_modules/debug": {
       "version": "4.4.3",
       "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.3.tgz",
@@ -3303,7 +3331,6 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/delayed-stream/-/delayed-stream-1.0.0.tgz",
       "integrity": "sha512-ZySD7Nf91aLB0RxL4KGrKHBXl7Eds1DAmEdcoVawXnLD7SDhpNgtuII2aAkg7a7QS41jxPSZ17p4VdGnMHk3MQ==",
-      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=0.4.0"
@@ -3490,7 +3517,6 @@
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/es-set-tostringtag/-/es-set-tostringtag-2.1.0.tgz",
       "integrity": "sha512-j6vWzfrGVfyXxge+O0x5sh6cvxAog0a/4Rdd2K36zCMV5eJ+/+tOAngRO8cODMNWbVRdVlmGZQL2YS3yR8bIUA==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "es-errors": "^1.3.0",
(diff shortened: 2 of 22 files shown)


declanblanc (changes requested):
## PR Review Checklist

- [x] **1. Logic over Style** — The storage provider abstraction is sound; upload path validates before writing; UUID-based filename prevents collisions and path traversal. `listByMessage` correctly checks soft-deleted messages.
- [ ] **2. Security First** — Two issues identified (see below): MIME type spoofing via client-supplied Content-Type, and the `_userId` variable is captured but never actually used for auditing.
- [x] **3. Architectural Alignment** — Upload uses REST (correct, per the tRPC/REST split); authenticated list query uses tRPC with `withPermission`. Consistent with established patterns.
- [x] **4. Issue Completion** — Issue #112 requires: attachment metadata CRUD, upload endpoint with S3-compatible interface, content-type/size validation, URL generation, and minimum tests. All present. The "wire into message service" sub-task is deferred to the client (caller passes the returned URL to `message.sendMessage`) which is acceptable per the PR description.
- [ ] **8. Prevent CI Failures** — `attachment.service.ts` uses `TRPCError` inside a REST route handler. `TRPCError` is never caught by Express's error pipeline and will produce an unhandled-rejection / 500 response instead of a 400, breaking the existing test that expects `400` for a disallowed content type (the test mocks the storage provider but does not mock `validateUpload`, so the real code path runs).

---

## Review

### Required Change 1 — `TRPCError` thrown from a REST context produces 500, not 400

`attachmentService.validateUpload` throws `TRPCError`:

```ts
// attachment.service.ts
throw new TRPCError({ code: 'BAD_REQUEST', message: `Unsupported content type: ${contentType}` });
```

This error is caught in the REST route handler:

```ts
// attachment.router.ts
try {
  attachmentService.validateUpload(mimetype, size);
} catch (err: unknown) {
  const message = err instanceof Error ? err.message : 'Validation failed';
  res.status(400).json({ error: message });
  return;
}
```

`TRPCError` extends `Error`, so `message` is populated correctly and the 400 path does work at runtime. However, coupling a pure validation function to a tRPC-specific error class is a layering violation — `attachmentService` is called from both REST routes and tRPC procedures, so it should throw a plain `Error` (or a domain-specific error class) rather than a tRPC transport-layer error. If `validateUpload` is ever called from a context that does not wrap it in a try/catch for HTTP mapping, the `TRPCError` will propagate up uncaught.

**Fix:** Replace `TRPCError` in `validateUpload` with a plain `Error` (or a lightweight `ValidationError` class). The REST handler already catches and maps it; the tRPC router can similarly catch and re-throw as `TRPCError` if needed.

### Required Change 2 — MIME type validation relies solely on client-supplied Content-Type

The upload handler reads `mimetype` directly from `req.file.mimetype`, which multer populates from the `Content-Type` field of the multipart form part — i.e., it is fully attacker-controlled. A client can upload a shell script or malicious binary and pass `contentType: 'image/png'` in the form part, bypassing the whitelist entirely.

```ts
// attachment.router.ts
const { originalname, mimetype, buffer, size } = req.file;
// mimetype is the value the client sent — not detected from the file contents
attachmentService.validateUpload(mimetype, size);
```

For a production deployment this is a significant security gap. At minimum, the service should use magic-byte detection (e.g., the `file-type` npm package) to verify the actual file contents match the declared MIME type, or store files in an isolated origin so that a bypass cannot lead to XSS/SSRF from the served files endpoint.

### Advisory — `_userId` captured but unused

```ts
const _userId = (req as AuthenticatedRequest).userId; // authenticated; logged for audit if needed
```

The comment says "logged for audit if needed" but there is no audit log, no structured log line, and no pass-through to the storage provider. If this is intentional placeholder code it should either be removed or wired to an actual log statement so reviewers can verify auth is enforced (it is — `requireAuth` runs first — but the unused variable is misleading).

declanblanc (changes requested):
## PR Review Checklist

- [x] **1. Logic over Style** — Storage abstraction is sound; upload validates before writing; UUID filenames prevent collisions and path traversal. `listByMessage` correctly cross-checks server ownership to prevent probing.
- [x] **2. Security First** — All three of declanblanc's prior concerns have been addressed in commit 4b2fc67: `AttachmentValidationError` replaces `TRPCError` in the service layer, magic-byte detection via `file-type` guards against MIME spoofing, and extension is derived from server-side `MIME_TO_EXT` rather than the user-supplied filename.
- [x] **3. Architectural Alignment** — Upload uses REST (correct per the tRPC/REST split); `listByMessage` uses tRPC with `withPermission`. Consistent with established patterns.
- [x] **4. Issue Completion** — Issue #112 is satisfied: upload endpoint, storage abstraction, content-type/size validation, URL generation, and tests are all present.
- [x] **6. Avoid Repetition** — declanblanc's previous concerns (TRPCError in service layer, MIME spoofing, unused `_userId`) are confirmed resolved and not repeated here.
- [x] **7. Iterative Reviews** — All three declanblanc CHANGES_REQUESTED items are resolved. Copilot items are also resolved.
- [ ] **8. Prevent CI Failures** — One remaining issue identified (see below).

---

## Review

### Required Change — `attachment.service.ts` still imports and uses `TRPCError` in `listByMessage`

The commit message for 4b2fc67 states "Replace TRPCError in validateUpload with AttachmentValidationError" and `validateUpload` is correctly fixed. However, `listByMessage` still imports and throws `TRPCError`:

```ts
// attachment.service.ts
import { TRPCError } from '@trpc/server';
...
async listByMessage(messageId: string, serverId: string) {
  ...
  if (!message || message.isDeleted) {
    throw new TRPCError({ code: 'NOT_FOUND', message: 'Message not found' });
  }
  if (message.channel.serverId !== serverId) {
    throw new TRPCError({ code: 'NOT_FOUND', message: 'Message not found' });
  }
  ...
}
```

`listByMessage` is called exclusively from the tRPC router (`attachment.listByMessage`), so this does not introduce an immediate runtime bug. However, the same layering principle that motivated fixing `validateUpload` applies here: `attachmentService` is declared as a transport-agnostic service, and importing `TRPCError` couples it to the tRPC layer. The `TRPCError` import is now the only reason the service depends on `@trpc/server`. If `listByMessage` is ever called from a REST handler (following the same pattern as `validateUpload`), the NOT_FOUND errors will surface as unhandled exceptions rather than 404 responses.

**Fix:** Throw a plain `Error` or a `NotFoundError` subclass from `listByMessage`, and let the tRPC router re-throw as `TRPCError` — the same pattern used by `visibilityService` and other service files in this codebase.

declanblanc (changes requested):
## PR Review Checklist

- [x] **1. Logic over Style** — Storage abstraction is sound; upload validates before writing; UUID-based filename from `LocalStorageProvider` prevents collisions and path traversal. `listByMessage` correctly cross-checks server ownership to prevent probing. `detectMimeType` wrapper correctly bypasses magic-byte check for `text/plain` (which has no magic bytes). HSETNX is not applicable here; the attachment upload pipeline logic is correct.
- [ ] **2. Security First** — `listByMessage` still imports and throws `TRPCError` from the service layer, which is a transport coupling issue identified in prior reviews. The service should remain transport-agnostic.
- [x] **3. Architectural Alignment** — Upload uses REST (correct per the tRPC/REST split); `listByMessage` uses tRPC with `withPermission`. Consistent with established patterns.
- [x] **4. Issue Completion** — Issue #112 requirements are satisfied: upload endpoint, storage abstraction, content-type/size validation, magic-byte detection, URL generation, and 17 passing tests.
- [x] **6. Avoid Repetition** — The `TRPCError` in `listByMessage` concern is unresolved from the previous review (not a duplicate; it was not fixed in commit `4b2fc673`).
- [x] **7. Iterative Reviews** — Prior `validateUpload` / MIME spoofing / `_userId` concerns are all confirmed resolved in this commit. The outstanding concern from declanblanc's second review on `listByMessage` is evaluated below.
- [ ] **8. Prevent CI Failures** — `attachment.service.ts` still imports `TRPCError` and uses it in `listByMessage`. The `TRPCError` import from `@trpc/server` is the only reason this service depends on the tRPC package. If `listByMessage` is ever called from a REST handler, the thrown `TRPCError` will not be caught by Express's HTTP error pipeline and will surface as an unhandled exception.

---

## Review

### Required Change — `listByMessage` still throws `TRPCError` (layering violation from prior review)

The commit message for `4b2fc673` states it replaces `TRPCError` in `validateUpload`, and that fix is correctly applied. However, the `TRPCError` import and usage in `listByMessage` was not addressed:

```ts
// attachment.service.ts
import { TRPCError } from '@trpc/server';  // still present
...
async listByMessage(messageId: string, serverId: string) {
  ...
  if (!message || message.isDeleted) {
    throw new TRPCError({ code: 'NOT_FOUND', message: 'Message not found' });
  }
  if (message.channel.serverId !== serverId) {
    throw new TRPCError({ code: 'NOT_FOUND', message: 'Message not found' });
  }
  ...
}
```

`listByMessage` is currently called only from the tRPC router, so there is no immediate runtime breakage. However, the `AttachmentValidationError` pattern was introduced precisely to keep the service transport-agnostic, and `listByMessage` undoes that principle. If a REST handler is added to serve attachments (a plausible future need, given the upload endpoint is already REST), the `TRPCError` will propagate as an unhandled exception rather than a 404.

**Fix:** Replace both `TRPCError` throws in `listByMessage` with a plain `Error` or a domain-specific subclass (e.g. `AttachmentNotFoundError extends Error`). The tRPC router can catch and re-throw as `TRPCError` if needed — the same pattern recommended for `validateUpload`. Remove the `@trpc/server` import from `attachment.service.ts` once this is done.

declanblanc (changes requested):
## PR Review Checklist

- [x] **1. Logic over Style** — Storage provider abstraction is sound; upload validates before writing; UUID-based filename prevents collisions and path traversal. listByMessage correctly checks soft-deleted messages and enforces server ownership to prevent cross-server probing.
- [ ] **2. Security First** — listByMessage in attachment.service.ts still imports and throws TRPCError, coupling the service layer to the tRPC transport. This is the same layering violation identified in prior reviews for validateUpload, which was correctly fixed. listByMessage was not updated.
- [x] **3. Architectural Alignment** — Upload uses REST (correct per the tRPC/REST split); listByMessage uses tRPC with withPermission. Consistent with established patterns.
- [x] **4. Issue Completion** — Issue #112 requirements are satisfied: upload endpoint, storage abstraction, content-type/size validation, magic-byte detection, URL generation, and tests.
- [x] **5. No Nitpicking** — N/A (reviewer obligation).
- [ ] **6. Avoid Repetition** — The TRPCError in listByMessage concern was raised in prior reviews and is still unresolved on commit 4b2fc673. This review confirms it remains open and is not a duplication.
- [x] **7. Iterative Reviews** — Prior concerns (TRPCError in validateUpload, MIME spoofing via client-supplied Content-Type, unused _userId) are confirmed resolved in this commit. One item from prior reviews remains open.
- [ ] **8. Prevent CI Failures** — attachment.service.test.ts currently asserts .rejects.toMatchObject({ code: 'NOT_FOUND' }) on listByMessage errors, relying on the TRPCError shape. If listByMessage is corrected to throw a plain Error, those test assertions will also need updating.

---

## Review

### Required Change — listByMessage still throws TRPCError (unresolved from prior reviews)

attachment.service.ts still imports and uses TRPCError in listByMessage. The AttachmentValidationError pattern was introduced in this PR specifically to keep the service layer transport-agnostic, but listByMessage was not updated with the same fix. The @trpc/server import is now the only reason attachment.service.ts depends on the tRPC package.

**Fix:** Replace both TRPCError throws in listByMessage with a plain Error or a domain-specific subclass (e.g. AttachmentNotFoundError extends Error). The tRPC router can catch and re-throw as TRPCError. Remove the @trpc/server import from attachment.service.ts once this is done.

**Cascade:** attachment.service.test.ts currently matches against `{ code: 'NOT_FOUND' }` on those throws. Once the service is updated to throw a plain Error, the matching assertions in the test file will also need to be updated to match on message content rather than the TRPCError code field.

FardeenI:
> Re: comment 3951123042

Fixed in 108e2cb. (1) AttachmentNotFoundError replaces TRPCError in listByMessage; @trpc/server import removed from attachment.service.ts. (2) Multer error handler path scope removed — now registered as attachmentRouter.use((err, req, res, next) => {...}) without a path, so it catches errors from any route on the router unambiguously.

PR #206, opened 2026-03-16
feat: real-time channel visibility changes propagate to all connected clients (#187)

### LINKED ISSUE #187: Real-time: channel visibility changes propagate to all connected clients
## Description

When a channel's visibility is toggled (e.g., PUBLIC_INDEXABLE → PRIVATE), clients already viewing that channel or server don't see the change until reload. The `VISIBILITY_CHANGED` event is already published by `visibility.service.ts` and consumed by `CacheInvalidator`, but it is never pushed to browser clients.

## Root Cause

`VISIBILITY_CHANGED` is already fully wired on the backend (published by `VisibilityService`, consumed by `CacheInvalidator` and `IndexingService`). There is no SSE bridge for this event type.

## Suggested Approach

1. Include `VISIBILITY_CHANGED` events in the server-scoped SSE endpoint (see #185)
2. Frontend hook handles `channel:visibility-changed` events
3. `HarmonyShell` / channel sidebar updates the channel's visibility badge immediately
4. If the current user loses access (channel became PRIVATE and they're not a member), redirect or show a "no longer accessible" state

## Affected Files

**Backend:**
- `harmony-backend/src/routes/events.router.ts` — add `VISIBILITY_CHANGED` to server SSE route

**Frontend:**
- `harmony-frontend/src/hooks/useServerEvents.ts` — handle `channel:visibility-changed`
- `harmony-frontend/src/components/layout/HarmonyShell.tsx` — update channel visibility in `localChannels`
- `harmony-frontend/src/components/channel/VisibilityGuard.tsx` — handle real-time access revocation

## Acceptance Criteria

- [ ] Toggling a channel's visibility in one client updates the visibility badge in all other connected clients' sidebars without reload
- [ ] If a channel is made PRIVATE and the current user is not a member, they are redirected or shown an appropriate message
- [ ] The `🌐 / 👁 / 🔒` visibility indicator in the channel header updates in real time

declanblanc:
## Summary

Closes #187

- **Backend** (`events.router.ts`): extends the existing `GET /api/events/server/:serverId` SSE route to subscribe to `VISIBILITY_CHANGED` events. On change, looks up the updated channel from Prisma and pushes a `channel:visibility-changed` SSE frame with the full channel object plus `oldVisibility` so clients can detect access revocation (PUBLIC → PRIVATE). Respects the existing server-ID filter so only the correct server's clients are notified.
- **Frontend** (`useServerEvents.ts`): adds optional `onChannelVisibilityChanged` callback to `UseServerEventsOptions`; wires a `channel:visibility-changed` event listener using the same stable-ref pattern as the existing member callbacks — fully backwards-compatible.
- **Frontend** (`HarmonyShell.tsx`): adds `handleChannelVisibilityChanged` — updates `localChannels` so the sidebar visibility badge (🌐 / 👁 / 🔒) refreshes in real time. If the current user is an unauthenticated guest viewing a channel that just became PRIVATE, they are redirected to the server root so `VisibilityGuard` can gate access.

## Test plan

- [ ] **Backend** — `npx jest tests/events.router.visibility.test.ts`: 4 tests — subscription registration, `channel:visibility-changed` delivery (with DB lookup), `oldVisibility` field presence, server-ID filter. All pass.
- [ ] **Frontend** — `npx jest src/__tests__/useServerEvents.test.tsx`: 18 tests (5 new visibility tests). All pass.
- [ ] Full frontend suite (`npx jest`): 41/41 pass. No TypeScript errors (`tsc --noEmit`). No lint errors.
- [ ] Manual: toggle a channel's visibility in one client — visibility badge should update in all other connected clients' sidebars without reload.
- [ ] Manual: toggle a public channel to PRIVATE while a guest user is viewing it — guest should be redirected to server root.

### DIFF
--- a/harmony-backend/src/routes/events.router.ts
+++ b/harmony-backend/src/routes/events.router.ts
@@ -26,6 +26,8 @@ import type {
   ChannelUpdatedPayload,
   ChannelDeletedPayload,
   ServerUpdatedPayload,
+  MemberJoinedPayload,
+  MemberLeftPayload,
 } from '../events/eventTypes';
 
 export const eventsRouter = Router();
@@ -334,6 +336,45 @@ eventsRouter.get('/server/:serverId', async (req: Request, res: Response) => {
     },
   );
 
+  // ── Subscribe to member join/leave events ─────────────────────────────────
+  // When a member joins, look up their profile and push the full user object so
+  // clients can add the new member to the sidebar without a page reload.
+
+  const { unsubscribe: unsubMemberJoined } = eventBus.subscribe(
+    EventChannels.MEMBER_JOINED,
+    async (payload: MemberJoinedPayload) => {
+      if (payload.serverId !== serverId) return;
+
+      try {
+        const user = await prisma.user.findUnique({
+          where: { id: payload.userId },
+          select: { id: true, username: true, displayName: true, avatarUrl: true },
+        });
+        if (!user) return;
+
+        sendEvent(res, 'member:joined', {
+          id: user.id,
+          username: user.username,
+          displayName: user.displayName,
+          avatar: user.avatarUrl ?? undefined,
+          // New members always join as MEMBER role; cast to frontend UserRole
+          role: payload.role.toLowerCase(),
+          status: 'online',
+        });
+      } catch {
+        // Silently ignore DB errors — the client will re-fetch on next load
+      }
+    },
+  );
+
+  const { unsubscribe: unsubMemberLeft } = eventBus.subscribe(
+    EventChannels.MEMBER_LEFT,
+    (payload: MemberLeftPayload) => {
+      if (payload.serverId !== serverId) return;
+      sendEvent(res, 'member:left', { userId: payload.userId });
+    },
+  );
+
   // ── Heartbeat ────────────────────────────────────────────────────────────
   const heartbeat = setInterval(() => {
     res.write(':\n\n');
@@ -345,5 +386,7 @@ eventsRouter.get('/server/:serverId', async (req: Request, res: Response) => {
     unsubChannelCreated();
     unsubChannelUpdated();
     unsubChannelDeleted();
+    unsubMemberJoined();
+    unsubMemberLeft();
   });
 });
--- a/harmony-backend/src/routes/events.router.ts
+++ b/harmony-backend/src/routes/events.router.ts
@@ -28,6 +28,7 @@ import type {
   ServerUpdatedPayload,
   MemberJoinedPayload,
   MemberLeftPayload,
+  VisibilityChangedPayload,
 } from '../events/eventTypes';
 
 export const eventsRouter = Router();
@@ -375,6 +376,35 @@ eventsRouter.get('/server/:serverId', async (req: Request, res: Response) => {
     },
   );
 
+  // ── Subscribe to visibility change events ─────────────────────────────────
+  // When a channel's visibility changes, push the updated channel object so
+  // connected clients can update the sidebar badge and handle access revocation
+  // (PRIVATE channels become inaccessible to non-members) without a page reload.
+
+  const { unsubscribe: unsubVisibilityChanged } = eventBus.subscribe(
+    EventChannels.VISIBILITY_CHANGED,
+    async (payload: VisibilityChangedPayload) => {
+      if (payload.serverId !== serverId) return;
+
+      try {
+        const channel = await prisma.channel.findUnique({
+          where: { id: payload.channelId },
+          select: CHANNEL_SSE_SELECT,
+        });
+        if (!channel) return;
+
+        sendEvent(res, 'channel:visibility-changed', {
+          ...channel,
+          // Include old visibility so clients can detect access revocation
+          // (e.g. current user is viewing a channel that just became PRIVATE).
+          oldVisibility: payload.oldVisibility,
+        });
+      } catch {
+        // Silently ignore DB errors — the client will re-fetch on next load
+      }
+    },
+  );
+
   // ── Heartbeat ────────────────────────────────────────────────────────────
   const heartbeat = setInterval(() => {
     res.write(':\n\n');
@@ -388,5 +418,6 @@ eventsRouter.get('/server/:serverId', async (req: Request, res: Response) => {
     unsubChannelDeleted();
     unsubMemberJoined();
     unsubMemberLeft();
+    unsubVisibilityChanged();
   });
 });
--- a/harmony-backend/src/routes/events.router.ts
+++ b/harmony-backend/src/routes/events.router.ts
@@ -357,7 +357,7 @@ eventsRouter.get('/server/:serverId', async (req: Request, res: Response) => {
           username: user.username,
           displayName: user.displayName,
           avatar: user.avatarUrl ?? undefined,
-          // New members always join as MEMBER role; cast to frontend UserRole
+          // Cast backend RoleTypeValue (e.g. 'MEMBER') to frontend UserRole (e.g. 'member')
           role: payload.role.toLowerCase(),
           status: 'online',
         });
--- a/harmony-backend/tests/events.router.member.test.ts
+++ b/harmony-backend/tests/events.router.member.test.ts
@@ -1,0 +1,328 @@
+/**
+ * events.router.member.test.ts — Issue #186
+ *
+ * Integration tests for member:joined / member:left SSE events on
+ * GET /api/events/server/:serverId.
+ *
+ * eventBus, prisma, and authService are mocked — no running Redis/DB needed.
+ */
+
+import http from 'http';
+import request from 'supertest';
+import { createApp } from '../src/app';
+import { eventBus } from '../src/events/eventBus';
+import { prisma } from '../src/db/prisma';
+import type { Express } from 'express';
+
+const VALID_TOKEN = 'valid-token';
+const VALID_SERVER_ID = '550e8400-e29b-41d4-a716-446655440002';
+const JOINING_USER_ID = 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee';
+
+// ─── Mock eventBus ─────────────────────────────────────────────────────────────
+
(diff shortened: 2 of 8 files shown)


declanblanc:
@copilot review

declanblanc:
@copilot review

acabrera04 (commented):
## Pre-Review Checklist

- [x] **1. Logic over Style** — Core logic is efficient and DRY. The visibility handler re-uses the existing `CHANNEL_SSE_SELECT` shape and `sendEvent` helper rather than introducing new abstractions.
- [x] **2. Security First** — The SSE endpoint already requires a valid JWT and server membership before streaming begins. Visibility events are only sent to verified members of the server in question.
- [x] **3. Architectural Alignment** — The implementation follows the existing pattern established by `CHANNEL_CREATED/UPDATED/DELETED`: subscribe on the global eventBus, filter by `serverId`, fetch fresh data from the DB, emit over the SSE stream. No deviations.
- [ ] **4. Issue Completion** — Two of three acceptance criteria are fully satisfied. One is partially satisfied with a behavioral correctness gap (Finding 1 and Finding 3).
- [x] **5. No Nitpicking** — Review focuses on logic, correctness, and security only.
- [x] **6. Avoid Repetition** — This is the first review on this PR.
- [x] **7. Iterative Reviews** — First review; N/A.
- [ ] **8. Prevent CI Failures** — No syntax errors or missing imports. The `EventChannels` mock fix is correct. However, the redirect guard logic change introduces a behavioral regression (Finding 1) that will not be caught by the existing tests.

---

## Review

### Finding 1 (HIGH) — Redirect condition is too broad: non-owner admins are incorrectly redirected

**File:** `harmony-frontend/src/components/layout/HarmonyShell.tsx`

```ts
if (
  channel.id === currentChannel.id &&
  oldVisibility !== ChannelVisibility.PRIVATE &&
  channel.visibility === ChannelVisibility.PRIVATE &&
  !checkIsAdmin(currentServer.ownerId)  // ← problem
) {
  router.push(`${basePath}/${currentServer.slug}`);
}
```

`checkIsAdmin(currentServer.ownerId)` returns `true` only when `authUser.id === currentServer.ownerId` (or when the user is a system admin). The `!checkIsAdmin(...)` guard therefore redirects **every authenticated user who is not the server owner**, including server admins who may have intentionally made the change. If an admin (non-owner) toggles a channel to PRIVATE while viewing it, they are immediately redirected away from the channel they just modified.

Additionally, `VisibilityGuard` only gates unauthenticated users from PRIVATE channels. This means the redirect in `HarmonyShell` is the **only enforcement point** for authenticated non-admin members, making the correctness of this condition critical to issue #187's access-revocation acceptance criterion.

The fix from the previous commit replaced `!isAuthenticated` (which was unreachable, since `useServerEvents` is only enabled when `isAuthenticated` is true) with `!checkIsAdmin(currentServer.ownerId)`. The unreachable guard was correctly identified, but the replacement guard has a scope problem. The intent described in the comment ("owners are not redirected because they may have intentionally made the change") should apply to any admin, not just the owner.

**Suggested fix:**

```ts
// Redirect only non-admin members who lose access.
// Admins (including non-owner admins) retain PRIVATE channel access.
const userIsAdmin = checkIsAdmin(currentServer.ownerId); // true for owner or system admin
if (
  channel.id === currentChannel.id &&
  oldVisibility !== ChannelVisibility.PRIVATE &&
  channel.visibility === ChannelVisibility.PRIVATE &&
  !userIsAdmin
) {
  router.push(`${basePath}/${currentServer.slug}`);
}
```

If `checkIsAdmin` does not cover non-owner server admins, the role check needs to be expanded to include the current user's server role. At minimum the comment should be corrected to say "server owners and admins are not redirected."

---

### Finding 2 (MEDIUM) — `VisibilityGuard` not updated to gate authenticated non-admin members

**File:** `harmony-frontend/src/components/channel/VisibilityGuard.tsx`

```ts
if (visibility === ChannelVisibility.PRIVATE && !isAuthenticated) {
  return <AccessDeniedPage />;
}
```

Issue #187's acceptance criterion says: _"If a channel is made PRIVATE and the current user is not a member, they are redirected or shown an appropriate message."_

`VisibilityGuard` only shows `AccessDeniedPage` for unauthenticated users. Authenticated non-admin members who navigate directly to a PRIVATE channel URL (e.g., after a reconnection where the SSE event was missed, or via a direct link) will receive channel content rather than an access-denied page. The PR correctly adds a real-time redirect for SSE-connected sessions, but the static guard is not updated to cover the non-real-time path. This makes the access revocation criterion partially satisfied rather than fully satisfied.

---

### Finding 3 (MEDIUM) — No backend test for the `channel not found` path in the VISIBILITY_CHANGED handler

**File:** `harmony-backend/tests/events.router.visibility.test.ts`

The `beforeEach` always resolves `prisma.channel.findUnique` to `MOCK_CHANNEL_ROW`. There is no test where `findUnique` returns `null` (channel deleted between event publish and handler execution). The backend code correctly handles this with `if (!channel) return;`, but this branch is untested. Given the silent-return behavior is the only thing preventing a crash in that race condition, it warrants a dedicated test.

---

### Finding 4 (LOW) — "Registers listeners for all event types" test is incomplete after adding the new event

**File:** `harmony-frontend/src/__tests__/useServerEvents.test.tsx`

The test labelled "registers listeners for all five event types" asserts only the five pre-existing event types and does not include `channel:visibility-changed`. The separate delivery and unmount tests for the new event exist and are correct, but the canonical registration test is stale. A future refactor that accidentally drops the listener would not be caught by this test.

---

## Acceptance Criteria Summary

| Criterion | Status |
|---|---|
| Visibility badge updates in all connected clients' sidebars without reload | ✅ Satisfied |
| Non-member users are redirected/shown a message when a channel goes PRIVATE | ⚠️ Partially satisfied — SSE path redirects non-owners (too broadly), static `VisibilityGuard` path doesn't gate authenticated non-admins |
| Visibility indicator in channel header updates in real time | ✅ Satisfied via `localChannels` state |

acabrera04 (changes requested):
## Pre-Review Checklist

- [x] **1. Logic over Style** — Focused changes. `serverOwnerId` is plumbed through the render tree without introducing a new context layer. Race-condition test is well-structured.
- [x] **2. Security First** — The `VisibilityGuard` change strengthens access control. SSE auth is unchanged.
- [x] **3. Architectural Alignment** — Uses the existing `useAuth().isAdmin` API; no new abstractions.
- [ ] **4. Issue Completion** — Findings 3 and 4 are fully resolved. The intent of Findings 1 and 2 is correct, but the implementation has a gap: `checkIsAdmin()` without argument does not cover non-owner server admins (see inline comments).
- [x] **5. No Nitpicking** — Only the substantive implementation gap is reported.
- [x] **6. Avoid Repetition** — Findings 3 and 4 are resolved; not repeated here.
- [x] **7. Iterative Reviews** — All four previous findings checked; resolution status documented below.
- [ ] **8. Prevent CI Failures** — The non-owner server admin path remains broken in both the real-time redirect and the static guard, and is not covered by tests.

---

## Resolution Status

**Finding 3 (MEDIUM) — Backend test for `channel not found` path** ✅ RESOLVED
Race-condition test added in `events.router.visibility.test.ts`. `prisma.channel.findUnique` returning `null` is correctly simulated and the no-event assertion passes.

**Finding 4 (LOW) — Stale listener registration test** ✅ RESOLVED
Test renamed to "registers listeners for all six event types"; `onChannelVisibilityChanged` is included in both the hook invocation and the `addedTypes` assertion.

**Finding 1 (HIGH) — Redirect condition too broad** ⚠️ PARTIALLY ADDRESSED — see inline comment on `HarmonyShell.tsx`

**Finding 2 (MEDIUM) — `VisibilityGuard` not gating authenticated non-admin members** ⚠️ PARTIALLY ADDRESSED — see inline comment on `VisibilityGuard.tsx`

Both findings share the same root cause, detailed in the inline comments below.

acabrera04 (approved):
## Pre-Review Checklist

- [x] **1. Logic over Style** — The fix is minimal and correct. `localMembers.find(m => m.id === authUser?.id)` is the right lookup. `isServerAdmin` prop is clearly named and documented.
- [x] **2. Security First** — The change tightens access control, not loosens it. `ChannelPageContent` derives `isServerAdmin` server-side from the authoritative member list. No new attack surface.
- [x] **3. Architectural Alignment** — Follows the existing server-side data-loading pattern in `ChannelPageContent`. The dependency array update is correct React practice.
- [x] **4. Issue Completion** — All four findings from prior reviews are now resolved.
- [x] **6. Avoid Repetition** — All findings resolved; not repeated.
- [x] **7. Iterative Reviews** — All previous findings verified; all resolved (documented below).
- [x] **8. Prevent CI Failures** — `getCurrentUser()` is added to an existing `Promise.all` — no extra round-trip latency. `isServerAdmin` is `boolean` (no undefined path reaches `VisibilityGuard`). Dependency array is correctly updated.

---

## Resolution Status

**Finding 1 (HIGH) — Non-owner server admins incorrectly redirected** ✅ RESOLVED
`HarmonyShell.tsx` now looks up `memberRecord` from `localMembers` (the server-scoped member list) and checks `memberRecord?.role === 'owner' || memberRecord?.role === 'admin'`. This correctly identifies non-owner server admins whose `user.role` in `AuthContext` is always `'member'`. The dependency array includes `localMembers` and `authUser?.id` — correct.

**Finding 2 (MEDIUM) — `VisibilityGuard` blocking non-owner server admins on direct URL** ✅ RESOLVED
`VisibilityGuard` now accepts `isServerAdmin?: boolean` prop. `ChannelPageContent` derives this server-side via `members.find(m => m.id === currentUser.id)?.role`, bypassing the `AuthContext.user.role` limitation. The JSDoc on the prop clearly explains why this is necessary.

**Finding 3 (MEDIUM) — Backend test for `channel not found` path** ✅ RESOLVED (from previous review)

**Finding 4 (LOW) — Stale listener registration test** ✅ RESOLVED (from previous review)

All findings resolved. No new issues found.

PR #249, opened 2026-03-25
TDD: MEMBER pin attempt throws FORBIDDEN (issue #236)

### LINKED ISSUE #236: fix: pin button shown to all users; improve pin error messages
### 1. Pin button visibility is incorrectly gated

In `HarmonyShell.tsx:189`, `canPin` is derived as:

```ts
const canPin = isAuthenticated;
```

This means **every authenticated user** sees the ⋯ More dropdown with Pin/Unpin, regardless of their server role. The backend correctly rejects unauthorised attempts (`message:pin` requires MODERATOR+), but the UI should reflect this — MEMBERs and GUESTs should never see the pin option at all.

### 2. Error message gives no actionable feedback

When a pin/unpin action fails (`handlePinToggle` in `MessageItem.tsx`), the inline feedback is a static `"Failed"` label. The tRPC error thrown by the backend (e.g. `FORBIDDEN: You do not have permission to perform 'message:pin'`) is swallowed entirely, so the user has no idea why it failed.

### 3. Backend returns HTTP 500 instead of 403 for permission violations

When a Member clicks Pin, the backend correctly rejects the request but returns **HTTP 500** instead of 403, making it impossible to distinguish permission errors from server errors.

## Expected Behaviour

- **Visibility**: Only users with MODERATOR, ADMIN, or OWNER roles in the current server should see the pin/unpin option. MEMBERs and GUESTs should not see the ⋯ More dropdown at all (or at minimum not the pin item).
- **Error messages**: When a pin action fails, surface a human-readable message:
  - Permission denied → "You don't have permission to pin messages."
  - Generic failure → "Failed to pin message. Please try again."
- **Backend**: Permission violations return HTTP 403, not 500.

## Steps to Reproduce

1. Log in as a regular Member
2. Navigate to a channel with messages
3. Hover a message → click ⋯ → observe "Pin message" is present
4. Click "Pin message" → observe "Failed" with no explanation; backend logs show 500

## Acceptance Criteria

- [ ] `canPin` is derived from the current user's actual server role (MODERATOR+), not just `isAuthenticated`
- [ ] Users below MODERATOR rank do not see the pin/unpin menu item
- [ ] Backend returns 403 (not 500) for permission violations on pin/unpin
- [ ] FORBIDDEN errors from the backend show a clear "no permission" message in the UI
- [ ] Other errors show a generic retry message
- [ ] Existing `message:pin` backend permission enforcement is unchanged

---
*Originally found during dogfood session 2026-03-22. HTTP 500 detail added from #239.*

AvanishKulkarni:
## Summary
- Adds a router-level test to `message.service.test.ts` asserting that a `MEMBER` without `message:pin` permission receives `FORBIDDEN` when calling `pinMessage` via the TRPC caller
- Uses `createCallerFactory` + `messageRouter` so the `withPermission('message:pin')` middleware executes (the service itself has no actor parameter — permission is enforced at the router layer)

## Notes
The test suite currently fails to compile due to a pre-existing TypeScript mismatch between `message.service.ts` and the Prisma schema (`parentMessageId`/`replyCount` fields missing from schema). This is unrelated to issue #236 and was broken on `main` before this branch.

## Test plan
- [ ] Once the schema/service mismatch is resolved, run `npx jest tests/message.service.test.ts` and confirm the new `messageRouter.pinMessage — permission enforcement` describe block passes

Closes #236 (test coverage portion)

### DIFF
--- a/harmony-backend/tests/message.service.test.ts
+++ b/harmony-backend/tests/message.service.test.ts
@@ -10,6 +10,8 @@ import { PrismaClient } from '@prisma/client';
 import { TRPCError } from '@trpc/server';
 import bcrypt from 'bcryptjs';
 import { messageService } from '../src/services/message.service';
+import { createCallerFactory, type TRPCContext } from '../src/trpc/init';
+import { messageRouter } from '../src/trpc/routers/message.router';
 
 const prisma = new PrismaClient();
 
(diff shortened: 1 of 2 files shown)


declanblanc (changes requested):
## Checklist

- [x] **1. Logic over Style** — Changes are focused and not redundant
- [ ] **2. Security First** — No new vulnerabilities, but a test asserts a contract the implementation does not fulfil (see below)
- [ ] **3. Architectural Alignment** — Frontend test is misaligned with the actual server action implementation
- [ ] **4. Issue Completion** — PR addresses only one of six acceptance criteria from issue #236
- [x] **5. No Nitpicking** — Only substantive issues raised
- [x] **6. Avoid Repetition** — No overlap with the existing Copilot bot review
- [x] **7. Iterative Reviews** — Existing Copilot review was checked before writing this one
- [x] **8. Prevent CI Failures** — One critical issue here: the frontend test will fail (see below)

---

## Issues

### 1. Frontend test asserts a contract the implementation does not fulfill (CI failure)

`pinMessageAction.test.ts` expects `pinMessageAction` to **return `{ error: string }`** when the backend responds with 403, and to **rethrow** for any other status:

```ts
// test expects this
const result = await pinMessageAction(MESSAGE_ID, SERVER_ID);
expect(result).toEqual(expect.objectContaining({ error: expect.any(String) }));
```

But the actual implementation in `harmony-frontend/src/app/actions/pinMessage.ts` is:

```ts
export async function pinMessageAction(messageId: string, serverId: string): Promise<void> {
  await trpcMutate('message.pinMessage', { messageId, serverId });
}
```

There is no try/catch, no 403 branching, and the return type is `void`. The first test case will fail because the function throws rather than returning `{ error }`. The second test case will also fail because the function throws unconditionally for all errors, which happens to match the assertion — but only by accident, not by design, and this masks the real bug.

The implementation needs to be updated to match what the test describes, or the test needs to be removed. Since issue #236 explicitly requires "FORBIDDEN errors from the backend show a clear 'no permission' message in the UI", the implementation fix is the correct path.

### 2. Issue #236 has six acceptance criteria; this PR covers only one

The PR covers only: *"Backend returns 403 (not 500) for permission violations on pin/unpin"* (via the backend integration test).

The remaining unaddressed criteria are:
- `canPin` must be derived from the user's actual server role (not `isAuthenticated`)
- Users below MODERATOR must not see the pin/unpin menu item
- FORBIDDEN errors must show a human-readable "no permission" message in the UI
- Other errors must show a generic retry message

A PR scoped to a single criterion is fine if it says so — but the frontend test file implies the error-surfacing behavior is also implemented, which it is not. This creates a misleading picture of issue completion and will cause the test suite to fail.

PR #251, opened 2026-03-25
fix: gate pin button behind MODERATOR+ role and show descriptive 403 error

acabrera04:
## Summary

- **`HarmonyShell.tsx`**: `canPin` was derived from `isAuthenticated`, showing the pin button to every logged-in user regardless of role. It now checks the user's server-scoped member record for `owner`, `admin`, or `moderator` — consistent with how other role checks already work in the file.
- **`MessageItem.tsx`**: The error handler now reads the thrown error's `status` field and maps known HTTP codes to human-readable messages. A 403 now shows `"You don't have permission to pin messages"` instead of the generic `"Failed"`.
- **`MessageItem.test.tsx`**: Two new Jest tests covering the fixed behavior — non-admins don't see the pin button, and a 403 rejection surfaces a descriptive error popup.

## Test plan

- [ ] Run `npm test` in `harmony-frontend/` — both `MessageItem` tests should pass
- [ ] Log in as a regular member (non-moderator) and verify the More actions button is absent on messages
- [ ] Log in as a moderator/admin and verify the More actions button and Pin/Unpin option are present
- [ ] (Optional) Force a 403 from the backend and confirm the error popup reads "You don't have permission to pin messages"

### DIFF
--- a/harmony-frontend/src/__tests__/MessageItem.test.tsx
+++ b/harmony-frontend/src/__tests__/MessageItem.test.tsx
@@ -1,0 +1,83 @@
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { MessageItem } from '@/components/message/MessageItem';
+import type { Message } from '@/types';
+
+// ─── Mocks ────────────────────────────────────────────────────────────────────
+
+jest.mock('next/image', () => ({
+  __esModule: true,
+  default: ({ alt, ...props }: React.ImgHTMLAttributes<HTMLImageElement> & { alt: string }) => (
+    // eslint-disable-next-line @next/next/no-img-element
+    <img alt={alt} {...props} />
+  ),
+}));
+
+jest.mock('@/lib/utils', () => ({
+  formatMessageTimestamp: () => 'Today at 12:00 PM',
+  formatTimeOnly: () => '12:00 PM',
+}));
+
+jest.mock('@/app/actions/pinMessage', () => ({
+  pinMessageAction: jest.fn(),
+  unpinMessageAction: jest.fn(),
+}));
+
+// ─── Fixtures ─────────────────────────────────────────────────────────────────
+
+const baseMessage: Message = {
+  id: 'msg-1',
+  channelId: 'ch-1',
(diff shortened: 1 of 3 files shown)


declanblanc (changes requested):
## Review Checklist

- [x] **Logic over Style**: The role-check logic and error-mapping logic are correct and efficient.
- [x] **Security First**: No vulnerabilities introduced. The frontend gate is defense-in-depth only; backend enforcement is unchanged.
- [x] **Architectural Alignment**: `localMembers` for server-scoped role checks is the established pattern in `HarmonyShell`.
- [ ] **Issue Completion**: One acceptance criterion from #236 is unaddressed (see below).
- [x] **No Nitpicking**: Only substantive issues raised.
- [x] **Avoid Repetition**: No prior reviews exist on this PR.
- [x] **Iterative Reviews**: Confirmed no existing reviews before writing this one.
- [x] **Prevent CI Failures**: No syntax errors, missing imports, or unhandled exceptions found.

---

## Request for Changes

### Missing acceptance criterion: backend must return 403 for permission violations

Issue #236 includes this acceptance criterion:

> Backend returns 403 (not 500) for permission violations on pin/unpin

This PR makes no changes to the backend. The three changed files are all frontend-only:
- `harmony-frontend/src/components/layout/HarmonyShell.tsx`
- `harmony-frontend/src/components/message/MessageItem.tsx`
- `harmony-frontend/src/__tests__/MessageItem.test.tsx`

The `PIN_ERROR_MESSAGES` map handles a 403 status code, but if the backend still returns 500 the mapping falls through to the generic `'Failed to pin message'` string — the descriptive error the issue requires will never actually appear in production. The frontend error handling is only useful once the backend sends the correct status code.

The backend pin/unpin endpoint must be fixed to emit HTTP 403 instead of 500 for `FORBIDDEN` / `message:pin` permission failures before this PR can be considered complete against the issue's acceptance criteria.

---

Everything else looks good: the role-gate logic in `HarmonyShell` is consistent with the existing `handleChannelVisibilityChanged` pattern, the `PIN_ERROR_MESSAGES` lookup table is a clean extensibility point, and the two new tests accurately exercise the fixed behavior.

Issue #239, opened 2026-03-23
fix: pin button shown to all users; non-moderator pin returns HTTP 500


acabrera04:
## Summary

Two compounding bugs in the pin message flow:

1. **Visibility bug**: `canPin` in `HarmonyShell.tsx` is set to `isAuthenticated` rather than checking the user's server role. Every logged-in user — including regular Members — sees a "Pin" option in the message action menu.
2. **Error handling bug**: When a Member clicks Pin, the backend correctly rejects the request, but returns **HTTP 500** instead of 403, and the frontend displays only a brief "Failed" label with no actionable message.

## Root Cause

`harmony-frontend/src/components/layout/HarmonyShell.tsx` line ~189:
```ts
const canPin = isAuthenticated; // should check MODERATOR+ role
```

## Expected Behavior

- The Pin option is hidden from users without `message:pin` permission (Moderator+)
- If a permission error occurs, the user sees a clear message: "You don't have permission to pin messages"
- Backend returns 403, not 500, for permission violations

## Steps to Reproduce

1. Log in as a regular Member
2. Navigate to a channel with messages
3. Hover a message → click ⋯ → observe "Pin message" is present
4. Click "Pin message" → observe "Failed" with no explanation; backend logs show 500

## Severity

High — functional

## Related

- GitHub Issue #236 (hide pin from non-moderators)

---
*Found during dogfood session 2026-03-22*

PR #254, opened 2026-03-25
fix: gate pin button on server role and surface FORBIDDEN errors (#236)

Aiden-Barrera:
### Overview

Two independent bugs were identified and fixed via TDD (red → green cycle).

---

### Bug 1 — `HarmonyShell.tsx:189`: Incorrect `canPin` derivation

**Failing assertion:**
```
expect(screen.queryByRole('button', { name: 'More actions' })).not.toBeInTheDocument()
// Found: <button aria-label="More actions" ...>
```

**Root cause:**

```ts
// Before — the bug
const canPin = isAuthenticated;
```

The comment defending this line claimed `localMembers` carries only the *global platform role*, making client-side role filtering unreliable. That claim is false.

`localMembers` is populated from `getServerMembers() → toFrontendMember()` which correctly maps the server-scoped role. The confusion arose because `mapBackendUser` (authService) hardcodes `role: 'member'` for all non-system-admins — but that function is only for `authUser`, not `localMembers`.

**The fix:**
```ts
const currentMember = localMembers.find(m => m.id === authUser?.id);
const canPin =
  currentMember?.role === 'moderator' ||
  currentMember?.role === 'admin' ||
  currentMember?.role === 'owner';
```

---

### Bug 2 — `MessageItem.tsx:103`: Parameterless `catch` discards error

**Failing assertion:**
```
Unable to find an element with text: "You don't have permission to pin messages."
// Found: <span>Failed</span>
```

**Root cause:**

```tsx
// Before — the bug
} catch {
  setPinState('error');
}
```

The parameterless catch silently discards the thrown FORBIDDEN error. Every failure renders the same static `"Failed"` label regardless of cause.

**The fix:**
```tsx
} catch (err) {
  const msg = err instanceof Error ? err.message : '';
  const isForbidden = msg.includes('FORBIDDEN') || msg.includes('403');
  setPinState(isForbidden ? 'forbidden' : 'error');
}
```

With two distinct error states:
- `forbidden` → "You don't have permission to pin messages."
- `error` → "Failed to pin message. Please try again."

---

### Note on Backend Bug (HTTP 500 → 403)

The backend returning 500 instead of 403 for permission violations is a separate fix tracked in #239 and is not covered by these frontend tests.

## Test plan
- [x] 8 new tests in `MessageItem.test.tsx` — all passing
- [x] Full test suite (79 tests) passing with no regressions
- [x] Restricted roles (member, guest) no longer see the pin button
- [x] Elevated roles (moderator, admin, owner) still see the pin button
- [x] FORBIDDEN errors show permission-specific message
- [x] Generic errors show retry message

### DIFF
--- a/harmony-frontend/src/__tests__/MessageItem.test.tsx
+++ b/harmony-frontend/src/__tests__/MessageItem.test.tsx
@@ -1,0 +1,156 @@
+/**
+ * MessageItem.test.tsx — Issue #236
+ *
+ * Verifies the two fixes introduced for issue #236.
+ *
+ * Fix 1 — Visibility:
+ *   HarmonyShell.tsx now derives canPin from the current user's server-scoped role
+ *   in localMembers (MODERATOR+) instead of `isAuthenticated`. This means
+ *   restricted roles (MEMBER, GUEST) receive canPin=false and must not see the
+ *   ⋯ More actions button. These tests verify MessageItem correctly hides the
+ *   button when canPin=false.
+ *
+ * Fix 2 — Error message:
+ *   handlePinToggle's catch block now inspects the thrown error. A FORBIDDEN
+ *   rejection must surface "You don't have permission to pin messages." instead
+ *   of the previous generic "Failed" label.
+ *
+ * Input:      User roles — member, guest (restricted) and moderator, admin, owner (elevated)
+ * Comparator: Exact DOM text / exact button presence check
+ * Oracle:     MEMBER/GUEST must NOT see the More actions button (canPin=false);
+ *             MODERATOR+ must see it (canPin=true);
+ *             a FORBIDDEN error must show "You don't have permission to pin messages."
+ */
+
+import React from 'react';
+import { render, screen, fireEvent, act } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { MessageItem } from '../components/message/MessageItem';
+import type { Message } from '../types';
+
+// ─── Mocks ────────────────────────────────────────────────────────────────────
+
+jest.mock('next/image', () => ({
+  __esModule: true,
+  default: (props: React.ImgHTMLAttributes<HTMLImageElement>) => <img {...props} />,
+}));
+
+const mockPinMessageAction = jest.fn();
+
+jest.mock('../app/actions/pinMessage', () => ({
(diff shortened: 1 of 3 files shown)


declanblanc (approved):
## Review Checklist

- [x] **Logic**: Both fixes are logically sound and DRY
- [x] **Security**: No new vulnerabilities; defence-in-depth maintained (backend still enforces permission)
- [x] **Architectural Alignment**: `canPin` derivation mirrors the existing `handleChannelVisibilityChanged` pattern in the same file
- [x] **Issue Completion**: All client-side ACs met; AC3 (backend 403) correctly deferred to #239
- [x] **No nitpicking**: Only substantive issues raised
- [x] **No repeated feedback**: No prior reviews exist
- [x] **CI**: 79 tests pass, 8 new tests added, no import/syntax issues
- [x] **Iterative**: First review on this PR

---

## Summary

Both bugs are correctly diagnosed and fixed. The PR description is thorough and the TDD evidence is solid.

**Fix 1 — `canPin` derivation** (`HarmonyShell.tsx`): The old comment claiming `localMembers` only carries the global platform role was incorrect. The fix correctly derives `canPin` from `localMembers.find(m => m.id === authUser?.id)?.role`, using the same `find`-by-id pattern already used in `handleChannelVisibilityChanged`. The role string comparisons match `UserRole` type values exactly.

One edge case to be aware of: `currentMember` will be `undefined` if `localMembers` is empty on initial render (before the server member list loads), causing `canPin` to be `false` during that window. For MODERATOR+ users this means the pin button will be briefly absent and then appear once members load. This is a safe default and acceptable given the SSR data flow, but worth documenting if it causes reported UX confusion.

**Fix 2 — FORBIDDEN error surfacing** (`MessageItem.tsx`): The string-sniff approach (`msg.includes('FORBIDDEN') || msg.includes('403')`) is pragmatic given the current tRPC error shape. It works correctly with the existing backend error message format. The `forbidden` state is properly reset via the existing `setTimeout(..., 3000)` that was already there for `error`.

The AC for the backend returning 403 instead of 500 is explicitly out of scope here and tracked in #239 — this is correctly noted in the PR.

All acceptance criteria that are in scope for this PR are satisfied. Approving.

PR #301, opened 2026-04-04
test: channel.service unit tests (CS-1 to CS-28)

declanblanc:
## Summary

- Rewrites `harmony-backend/tests/channel.service.test.ts` from integration tests (real DB) to a fully mock-based unit test suite
- Covers all 28 cases (CS-1 through CS-28) from `docs/test-specs/channel-service-spec.md`
- Mocks Prisma, `cacheService`, and `eventBus` — no real DB, Redis, or network required
- Achieves 100% statement/branch/function/line coverage on `channel.service.ts`

## Test coverage breakdown

| Spec section | Cases | What's verified |
|---|---|---|
| `getChannels` | CS-1, CS-2 | ascending position order; empty array |
| `getChannelBySlug` | CS-3, CS-4, CS-5 | success; server not found; channel not found |
| `createChannel` | CS-6 to CS-13 | success + cache/event; position default; VOICE guard; NOT_FOUND; CONFLICT; fire-and-forget isolation |
| `updateChannel` | CS-14 to CS-20, CS-28 | name/topic/position; NOT_FOUND variants; event payload; isolation; all-undefined patch |
| `deleteChannel` | CS-21 to CS-25 | all 3 cache ops + event; NOT_FOUND variants; event payload; isolation |
| `createDefaultChannel` | CS-26, CS-27 | fixed-arg delegation; error propagation |

## Test plan

- [x] All 28 tests pass locally (`npx jest tests/channel.service.test.ts`)
- [x] 100% coverage on `channel.service.ts`
- [x] Existing `channel.service.events.test.ts` still passes
- [x] No real DB/Redis/network dependencies

Closes #294

### DIFF
--- a/harmony-backend/tests/channel.service.test.ts
+++ b/harmony-backend/tests/channel.service.test.ts
@@ -4,8 +4,10 @@
  * Covers all 28 spec cases (CS-1 through CS-28) from
  * docs/test-specs/channel-service-spec.md.
  *
- * All external dependencies (Prisma, cacheService, eventBus) are mocked;
- * no real DB, Redis, or network connections are required.
+ * Uses a real test database with isolated fixtures per the spec.
+ * Only cache (cacheService) and event (eventBus) dependencies are mocked.
+ *
+ * Requires DATABASE_URL pointing at a running Postgres instance.
  */
 
 // ─── Mock eventBus ─────────────────────────────────────────────────────────────
@@ -18,30 +20,6 @@ jest.mock('../src/events/eventBus', () => ({
     CHANNEL_CREATED: 'harmony:CHANNEL_CREATED',
     CHANNEL_UPDATED: 'harmony:CHANNEL_UPDATED',
     CHANNEL_DELETED: 'harmony:CHANNEL_DELETED',
-  },
-}));
-
-// ─── Mock Prisma ───────────────────────────────────────────────────────────────
-
-const mockChannelFindMany = jest.fn();
-const mockChannelFindUnique = jest.fn();
-const mockChannelCreate = jest.fn();
-const mockChannelUpdate = jest.fn();
-const mockChannelDelete = jest.fn();
-const mockServerFindUnique = jest.fn();
-
-jest.mock('../src/db/prisma', () => ({
-  prisma: {
-    channel: {
-      findMany: mockChannelFindMany,
-      findUnique: mockChannelFindUnique,
-      create: mockChannelCreate,
-      update: mockChannelUpdate,
-      delete: mockChannelDelete,
-    },
-    server: {
-      findUnique: mockServerFindUnique,
-    },
   },
 }));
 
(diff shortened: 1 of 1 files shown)


harmony-backend/tests/channel.service.test.ts
```diff
@@ -100,154 +280,290 @@ describe('channelService.createChannel', () => {
...
-    ).rejects.toThrow(TRPCError);
+      channelService.deleteChannel(CHANNEL_ID, SERVER_ID),
+    ).resolves.toBeUndefined();
   });
+});
+
+// ─── CS-26, CS-27: createDefaultChannel ──────────────────────────────────────
+
+describe('channelService.createDefaultChannel', () => {
+  it('CS-26: delegates to createChannel with the correct fixed arguments', async () => {
```
acabrera04: CS-26 is supposed to verify that `createDefaultChannel()` delegates to `createChannel()` with the fixed defaults. This only checks the final Prisma `create` payload, so an implementation that duplicated the same logic inline instead of delegating would still pass. That means the delegation contract from the spec is not actually covered here.

harmony-backend/tests/channel.service.test.ts
```diff
@@ -100,154 +280,290 @@ describe('channelService.createChannel', () => {
...
+      channelService.updateChannel(CHANNEL_ID, OTHER_SERVER_ID, { name: 'x' }),
+    ).rejects.toThrow(
+      expect.objectContaining({ code: 'NOT_FOUND', message: 'Channel not found in this server' }),
+    );
   });
 
-  it('updates position', async () => {
-    const updated = await channelService.updateChannel(channelId, serverId, { position: 5 });
-    expect(updated.position).toBe(5);
+  it('CS-19: CHANNEL_UPDATED event payload contains channelId, serverId, and timestamp', async () => {
```
acabrera04: These event payload/timestamp assertions overlap the dedicated `channel.service.events.test.ts` suite, which already verifies the same `CHANNEL_UPDATED` publication contract. Since issue #294 asks to avoid duplicate or significantly overlapping tests, this should be consolidated rather than asserted in both files.

acabrera04 (changes requested):
- [ ] Prioritize Logic over Style: the rewrite improves line coverage, but it over-mocks the service’s DB-facing behavior and misses part of the spec’s intended contract.
- [x] Security First: test-only change; I do not see a new security issue here.
- [x] Architectural Alignment: no production architecture change is introduced by this PR.
- [ ] Issue Completion: `#294` asks this suite to match the approved spec and avoid significant overlap; the current version does not fully satisfy either point.
- [x] No Nitpicking: the findings below are about test correctness and spec alignment, not style.
- [x] Avoid Repetition: there were no existing review comments on this PR to duplicate.
- [x] Iterative Reviews: first pass from me on this PR, so there were no prior comments to re-check or resolve.
- [x] Prevent CI Failures: in an isolated checkout of this PR, the targeted Jest runs passed; the concern is test quality/spec compliance rather than an immediate runtime failure.

1. This rewrite no longer follows the approved setup in `docs/test-specs/channel-service-spec.md`. The spec says to use isolated DB fixtures and mock the cache/event dependencies, but the new suite mocks Prisma itself. That means composite slug lookup, uniqueness checks, and query-shape regressions are no longer exercised even though the file reports 100% coverage.

2. CS-26 is specified as a delegation test for `createDefaultChannel`, but the current assertion only verifies the downstream Prisma `create` payload. If `createDefaultChannel` stopped delegating and inlined equivalent logic, this test would still pass, so the delegation contract is not actually being checked.

3. The new event payload/timestamp assertions significantly overlap the dedicated `channel.service.events.test.ts` coverage. Issue `#294` explicitly asks to avoid duplicate or significantly overlapping tests, so these cases should be consolidated instead of maintained in parallel.

I also verified the current branch in an isolated worktree with:
`npm test -- --runTestsByPath tests/channel.service.test.ts tests/channel.service.events.test.ts`
and `tests/channel.service.test.ts` reports 100% coverage for `src/services/channel.service.ts`.

acabrera04 (changes requested):
- [ ] Prioritize Logic over Style: the earlier duplicate-coverage issue is fixed, but one spec-required behavior is still not actually asserted by the updated suite.
- [x] Security First: test-only change; no new security concern here.
- [x] Architectural Alignment: no production architecture change is introduced by this PR.
- [ ] Issue Completion: the PR is much closer now and the previous overlap/delegation concerns are addressed, but the suite still misses one explicit assertion required by the approved spec.
- [x] No Nitpicking: this is about test completeness against the accepted spec, not style.
- [x] Avoid Repetition: I am not repeating the earlier resolved findings; those appear addressed on the current head.
- [x] Iterative Reviews: I re-checked the updated commit and confirmed the prior overlap comment is fixed before posting this follow-up.
- [x] Prevent CI Failures: I reran the targeted Jest commands on the updated head; both `tests/channel.service.test.ts` and `tests/channel.service.events.test.ts` pass, and `channel.service.ts` still reports 100% coverage.

The previous duplicate event-payload issue is fixed, and the move back to real DB fixtures plus the CS-26 delegation spy both address the other earlier concerns. One remaining gap is that CS-8 no longer verifies the "before any Prisma call" part of the approved spec. The test asserts that `VOICE + PUBLIC_INDEXABLE` throws, but the spec explicitly requires proving that this guard fires before any database call. As written, a future regression that queries Prisma before throwing would still pass this test.

acabrera04 (approved):
- [x] Prioritize Logic over Style: the latest head addresses the prior test-quality issues, and I do not see a remaining logic problem in the rewritten suite.
- [x] Security First: test-only change; no security concern introduced.
- [x] Architectural Alignment: no production architecture change is introduced by this PR.
- [x] Issue Completion: the current version now aligns with the approved spec closely enough to satisfy the issue requirements, including the CS-8 guard assertion and CS-26 delegation check.
- [x] No Nitpicking: no findings remain.
- [x] Avoid Repetition: this follow-up does not repeat previously resolved feedback.
- [x] Iterative Reviews: I re-checked the latest commit specifically against the prior review comments, and the remaining CS-8 concern is now addressed.
- [x] Prevent CI Failures: I reran the targeted Jest commands on the latest head; `tests/channel.service.test.ts` and `tests/channel.service.events.test.ts` pass, and `channel.service.ts` still reports 100% statement/branch/function/line coverage.

The current head looks good. The suite is back on real DB fixtures, the duplicate event-payload assertions were removed in favor of the dedicated event suite, CS-26 now verifies delegation via a spy on `createChannel`, and CS-8 now proves the VOICE visibility guard fires before any Prisma call. I do not have any remaining findings.

PR #307, opened 2026-04-06
[codex] silence expected public router test console errors

acabrera04:
## Summary
- add a small `withSilencedConsoleError` helper in the public router test suite
- wrap the tests that intentionally force route-level `500` errors so expected `console.error` output does not spam Jest output
- keep the existing assertions around `500` responses unchanged

## Why
Those tests deliberately mock Prisma/cache failures to verify the route handlers return `500`. The handlers log via `console.error`, so the suite passed but printed noisy stack traces that looked like real failures.

## Impact
This is a test-only cleanup. Runtime behavior is unchanged, and the `500`-path coverage remains intact.

## Validation
- `npm run lint` in `harmony-backend` (passes with 4 pre-existing warnings)
- `npm test -- public.router.test.ts` in `harmony-backend`
- `npm test` in `harmony-frontend`
- `npm test` in `harmony-backend` still fails in pre-existing Redis-backed suites (`eventBus`, `cache.middleware`, `cache.service`) with `NOAUTH Authentication required`


AvanishKulkarni (approved):
Test-only cleanup — correct and safe. `try/finally` around `mockRestore()` is the right pattern. Minor nit: `mockImplementation(() => {})` reads slightly more conventionally than `() => undefined`, but not a blocker. Verified all wrapped tests preserve their original 500-path assertions. Approve.

Issue #14, opened 2026-02-16
Scaffold React + TypeScript + Vite project


acabrera04:
## Description
Initialize the Harmony frontend project from scratch using **Next.js 14+** (as specified in dev specs Section 8: Technology Stack).

## Corrected Tech Stack
The original issue specified Vite, but all dev specs (`docs/dev-spec-*.md`) specify:
- **T3: Next.js 14.0+** - SSR/SSG framework (SEO-critical for public pages)
- This aligns with the architecture requirements for server-side rendering and SEO optimization

## Tasks
- [x] Run `npx create-next-app@latest harmony-frontend --typescript --tailwind --app --src-dir`
- [x] Install core deps: `axios`
- [x] Install dev deps: `eslint`, `prettier`, `@types/node`
- [x] Configure `tsconfig.json` with strict mode, path aliases (`@/` → `src/`)
- [x] Set up folder structure: `src/{components,hooks,services,mocks,types,styles,assets,layouts,context}`
- [x] Create `.env.example`
- [x] Verify `npm install && npm run dev` works with a placeholder app

## Acceptance Criteria
- `npm install && npm run dev` starts a working Next.js dev server
- TypeScript strict mode is enabled
- Path aliases work (`@/components/...`)
- Folder structure is in place
- SSR works out of the box (Next.js default)

## Blocks
The following issues depend on this one: #15, #16, #19

## Tech Stack Reference
See `docs/dev-spec-*.md` Section 8 for canonical technology stack.

Issue #11, opened 2026-02-16
Frontend: Implement Guest Public Channel View


acabrera04:
## Description
Implement the read-only public channel view for anonymous/guest users, allowing them to view public channel content via a direct URL without logging in.

## Tasks
- [ ] Create `GuestChannelView` page component at route `/c/:serverSlug/:channelSlug`
- [ ] Create `MessageList` component for rendering read-only messages
- [ ] Create `MessageItem` component with author avatar, name, timestamp, and content
- [ ] Implement `VisibilityGuard` component that checks channel visibility:
  - `PUBLIC_INDEXABLE` or `PUBLIC_NO_INDEX`: render channel content
  - `PRIVATE`: render access denied page with sign-up CTA
- [ ] Implement loading skeleton state while channel data loads
- [ ] Implement error state (channel not found / server error)
- [ ] Implement empty state (channel exists but has no messages)
- [ ] Add "Join Harmony" CTA banner for guest users (sticky bottom bar or sidebar prompt)
- [ ] Connect to mock API services for channel and message data
- [ ] Implement responsive design for desktop and mobile breakpoints
- [ ] Add WCAG accessibility: semantic HTML, ARIA roles for chat log, alt text, keyboard navigation
- [ ] Write unit tests for VisibilityGuard logic and message rendering

## Acceptance Criteria
- Guest can view public channel content at `/c/:serverSlug/:channelSlug`
- Messages render with author, timestamp, and content
- Private channels show access denied with sign-up CTA
- Loading, error, and empty states all render correctly
- Component is responsive at both breakpoints
- Passes basic accessibility audit

## Dev Spec Reference
`docs/dev-spec-guest-public-channel-view.md` — Modules M1-M3, VisibilityGuard, PublicChannelPage

## Sprint
Week 2 (Days 5-8)

PR #311, opened 2026-04-09
Add Sprint 4 deployment plan

acabrera04:
## Summary
- add  for the Vercel + Railway deployment sprint
- cover the P6 deployment requirements using Vercel/Railway instead of AWS-specific services
- assign initial issues across all 5 team members with explicit dependency notes and 2+ Railway backend replica support

## Review focus
- issue scope and sequencing
- dependency accuracy
- whether the P6 coverage map and deliverables checklist are complete

### DIFF
--- a/planning/SPRINT_4_PLAN.md
+++ b/planning/SPRINT_4_PLAN.md
@@ -57,6 +57,11 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 - API replicas must be stateless apart from live SSE client connections
 - Deployment verification must prove that 2+ API replicas behave correctly behind Railway load balancing
 
+### Explicit Production Decisions
+- **SEO ownership:** the **Vercel frontend apex domain** is the canonical public SEO surface. `robots.txt`, sitemap/index entrypoints, canonical URLs, and `metadataBase` must resolve on the frontend domain, not the API subdomain.
+- **Attachment storage target:** production uploads use **Cloudflare R2 via an S3-compatible provider**. Local filesystem storage remains development-only.
+- **Deploy authority:** **GitHub Actions** is the single source of truth for production deploys. Provider-native Git auto-deploys may be used for previews, but production promotion must be driven by the GitHub workflows in this sprint.
+
 ---
 
 ## P6 Coverage Map
@@ -100,6 +105,7 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
   - Clear ownership of public vs internal services
   - Explicit decision that `backend-api` scales to 2+ replicas and `backend-worker` stays singleton
 - Assignee: **acabrera04**
+- Backup owner: **declanblanc**
 - Due: April 9
 - Blocked by: none
 - Unblocks: #2, #6, #7, #8, #16
@@ -118,17 +124,21 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
   - Risks are prioritized into must-fix vs acceptable-for-demo
   - `backend-api` vs `backend-worker` responsibilities are finalized
 - Assignee: **declanblanc**
+- Backup owner: **acabrera04**
 - Due: April 9
 - Blocked by: #1
 - Unblocks: #3, #4, #5, #14
 
 **3. Shared Rate Limiting + Proxy-Aware Networking**
 - Replace process-local rate limiting with shared Redis-backed limiting
+- Replace or unify **both** current implementations:
+  - auth endpoint limiting using `express-rate-limit`
+  - public-route token bucket limiting
 - Ensure auth and public API rate limits remain correct across 2+ replicas
 - Configure Express proxy awareness so client IP handling works correctly behind Railway
 - Acceptance criteria:
   - Public and auth rate limits are shared across replicas
-  - No process-local limit counters remain in production code paths
+  - No process-local auth or public-route limit counters remain in production code paths
   - Rate limit behavior is covered by tests or verification notes
 - Assignee: **Aiden-Barrera**
 - Due: April 11
@@ -136,12 +146,13 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 - Unblocks: #14
 
 **4. Production Attachment Storage Provider**
-- Implement a production storage provider backed by shared object storage
+- Implement a production storage provider backed by **Cloudflare R2 via an S3-compatible interface**
 - Keep local storage available for development only
 - Add env-driven provider selection and document required secrets
 - Ensure uploaded files remain accessible regardless of which API replica serves subsequent requests
 - Acceptance criteria:
   - Production does not rely on local filesystem attachment serving
+  - The chosen production provider is Cloudflare R2, documented as such in code and setup docs
   - README and env matrix document storage setup
   - Attachment flow works end-to-end in deployed environment
 - Assignee: **AvanishKulkarni**
@@ -153,12 +164,17 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 - Move background-only startup behavior into a dedicated worker process
 - Ensure duplicated subscribers / cache invalidation / sitemap upkeep do not run on every API replica
 - Keep API replicas focused on HTTP/tRPC/SSE request handling
+- Add lightweight replica observability for validation:
+  - instance identity in structured logs
+  - instance/replica identity on health output and/or response headers
 - Add startup commands for both Railway backend services
 - Acceptance criteria:
   - `backend-api` can run with 2+ replicas without duplicate singleton background side effects
   - `backend-worker` runs with 1 replica and owns singleton event-driven tasks
+  - Replica identity is externally observable enough to prove load balancing across 2+ API replicas
   - Service responsibilities are documented
 - Assignee: **declanblanc**
+- Backup owner: **acabrera04**
 - Due: April 12
 - Blocked by: #2
 - Unblocks: #7, #11, #14
@@ -169,12 +185,16 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 - Add production-safe frontend configuration:
   - absolute canonical URLs
   - `metadataBase`
-  - `robots.txt`
-  - sitemap support
+  - `robots.txt` on the frontend apex domain
+  - sitemap support on the frontend apex domain
   - production API base URL handling
+- Make the SEO ownership boundary explicit:
+  - frontend apex domain owns public SEO artifacts
+  - backend SEO routes are treated as source/internal inputs or deprecated public paths, but not the canonical SEO origin
 - Ensure frontend still supports localhost development cleanly
 - Acceptance criteria:
   - Public pages generate correct production metadata
+  - Canonical host ownership is explicit and consistent across frontend and backend docs/code
   - Frontend can target local and cloud backends without code edits
   - SEO-critical pages render correctly on the public domain
 - Assignee: **AvanishKulkarni**
@@ -195,6 +215,7 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
   - Domains/env vars/health checks are configured
   - `backend-api` and `backend-worker` both boot successfully in Railway
 - Assignee: **acabrera04**
+- Backup owner: **FardeenI**
 - Due: April 13
 - Blocked by: #1, #5
 - Unblocks: #11, #14, #15
@@ -219,13 +240,18 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 
 **9. Integration Test Implementation + Environment Matrix**
 - Implement the integration tests from the spec
+- Refactor the Playwright harness to support two execution modes:
+  - **local stack mode** that boots local frontend/backend processes
+  - **external deployed target mode** that skips local boot and targets deployed URLs
 - Add configuration so tests can run against:
   - localhost
   - deployed frontend + backend URLs
+- Parameterize the current local-only stack configuration so the test harness is not hard-wired to localhost ports/process launchers
 - Capture/structure output for both local and cloud runs
 - Acceptance criteria:
   - Tests run in a local configuration
   - Tests run in a cloud configuration
+  - The test harness can switch between local and deployed targets without code edits
   - Any environment-specific exceptions are documented
 - Assignee: **Aiden-Barrera**
 - Due: April 14
@@ -254,11 +280,14 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 - Deploy `backend-api` and `backend-worker` on pushes to `main`
 - Ensure migrations / build / deploy ordering is safe
 - Use GitHub secrets for Railway authentication
+- Treat GitHub Actions as the production deploy authority and document any provider-native auto-deploy settings that must be disabled or restricted
 - Acceptance criteria:
   - Workflow deploys backend services without manual intervention
+  - Production deploy authority is unambiguous and documented
   - Deploys target the correct Railway environment
   - Deployment process is documented in README
 - Assignee: **acabrera04**
+- Backup owner: **declanblanc**
 - Due: April 15
 - Blocked by: #5, #7
 - Unblocks: #14, #15
@@ -268,8 +297,10 @@ To safely support 2+ backend replicas, the sprint must remove or isolate process
 - Build/deploy frontend on pushes to `main`
 - Ensure preview/production environment variables are configured properly
 - Use GitHub secrets/tokens safely
+- Treat GitHub Actions as the production deploy authority and document any provider-native auto-deploy settings that must be disabled or restricted
 - Acceptance criteria:
   - Workflow deploys frontend without manual intervention
+  - Production deploy authority is unambiguous and documented
   - Production deploy points at the production backend URL
   - Deployment process is documented in README
 - Assignee: **AvanishKulkarni**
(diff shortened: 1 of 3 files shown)


AvanishKulkarni:
**[Technical Review — Sprint 4 Plan]**

Six technical gaps in the plan worth addressing before downstream issues are started:

---

**1. SSE with 2+ replicas is unresolved**

The plan notes API replicas are *"stateless apart from live SSE client connections"* but prescribes no solution. A client connected to replica A won't receive events emitted on replica B — this breaks realtime for any multi-replica path.

*Suggested fix:* Commit to a specific approach in the plan. The two realistic options are (a) Redis pub/sub fan-out on `backend-api` so each replica re-publishes to its own connected clients, or (b) sticky sessions at the Railway load balancer level. Option (a) is stateless-friendly and more consistent with the rest of the architecture. Whichever is chosen should be documented in `docs/deployment/replica-readiness-audit.md` (#2) before #5 is implemented.

---

**2. Prisma migration race condition**

If `backend-api` (2+ replicas) and `backend-worker` all run `prisma migrate deploy` on startup, multiple processes will attempt the migration simultaneously. This can corrupt migration state.

*Suggested fix:* Designate a single migration runner — the most common pattern is a one-shot Railway job or a pre-deploy command on `backend-worker` only, with `backend-api` replicas blocked from running migrations. This should be an explicit acceptance criterion in #11 (`deploy-railway.yml`) since the deploy workflow controls startup ordering.

---

**3. Sitemap handoff between frontend and backend is unspecified**

The plan states the frontend apex domain owns canonical sitemap entrypoints but that *"backend SEO routes may continue to generate sitemap/XML data as an internal or transitional source."* The mechanism is never defined — does the frontend proxy to the backend? Does the worker push pre-generated XML? Does the frontend call the backend at build time?

*Suggested fix:* Pick one pattern and state it explicitly in #6. The simplest production-safe approach is a Next.js route handler that fetches from the backend API at request time (ISR or on-demand). The plan should rule out the backend subdomain as a direct crawl target.

---

**4. Railway internal networking is never specified**

Both `backend-api` and `backend-worker` connect to Postgres and Redis. Railway provides private internal hostnames for inter-service traffic. The plan never says to use them, so implementers may default to public URLs — adding unnecessary latency and egress cost.

*Suggested fix:* Add a note in #7 (Railway provisioning) that Postgres and Redis connection strings must use Railway's private network hostnames, not the public TCP proxy. This should also be reflected in the env var matrix in `docs/deployment/deployment-architecture.md`.

---

**5. Express `trust proxy` level is unspecified**

Issue #3 says to configure proxy awareness for correct IP handling behind Railway, but doesn't specify the trust level. If set too broadly (e.g., `trust proxy: true`), any `X-Forwarded-For` header — including spoofed ones — is trusted, making IP-based rate limiting bypassable. If not set at all, all requests appear to originate from the Railway load balancer IP, defeating per-client limiting entirely.

*Suggested fix:* The plan (or #3's acceptance criteria) should specify the intended trust level. For Railway, `app.set('trust proxy', 1)` is typically correct — trusting one hop. This is a security-relevant configuration and should be verified in #14 as part of rate limit validation.

---

**6. `backend-worker` crash behavior is unaddressed**

The worker is load-bearing: if it crashes, cache invalidation, sitemap updates, and background jobs stop silently. The plan designates it a singleton but says nothing about restart policy, health checks, or graceful degradation when it's unavailable.

*Suggested fix:* Add to #5 or #7 that `backend-worker` must have a Railway health check configured and a restart-on-failure policy. The plan should also state how `backend-api` behaves if the worker is down — ideally, API requests remain functional and worker-dependent side effects are queued or deferred rather than erroring.

acabrera04 (commented):
## Review Checklist

- [ ] **Logic over Style**: The deployment plan still has unresolved technical planning gaps called out in the existing review, especially multi-replica SSE behavior and migration ownership.
- [ ] **Security First**: The `trust proxy` level and Railway private-network usage are not specified yet, which affects IP-based limiting and service exposure choices.
- [ ] **Architectural Alignment**: The intended Vercel/Railway split is directionally clear, but the worker/API boundary and SEO handoff need the missing operational details before downstream implementation starts.
- [ ] **Issue Completion**: No linked issue/acceptance criteria are present; P6 coverage cannot be considered complete until the existing deployment-plan gaps are incorporated.
- [x] **No Nitpicking**: Review is limited to deployment correctness, security, and sequencing risk.
- [x] **Avoid Repetition**: I reviewed the existing technical review and am not restating each item as fresh line comments.
- [x] **Iterative Reviews**: The existing unresolved planning feedback still applies.
- [x] **Prevent CI Failures**: Docs-only PR checks are green (5 passed, 0 failed), but the plan-level blockers would likely create downstream deployment failures if left vague.

I agree with the existing technical review and found no separate issue worth adding. Before merging, fold those unresolved items into the plan: choose the SSE strategy, designate a single migration runner, define frontend/backend sitemap handoff, require Railway private connection strings, specify Express `trust proxy` behavior, and add worker health/restart expectations.

acabrera04:
> Re: comment 4091019364

Fixed in c772049. The sprint plan now makes the missing operational decisions explicit: Redis fan-out for multi-replica SSE, a single migration runner, frontend apex sitemap handoff via Next.js route handlers, Railway private/internal connection strings, one-hop trust proxy behavior, and worker health/restart expectations.

acabrera04:
> Re: comment 4218158167

Fixed in c772049. The plan now explicitly names the SSE strategy, migration ownership, frontend sitemap handoff, Railway private/internal networking, one-hop trust proxy behavior, and worker health/restart expectations so downstream implementation issues have concrete operational guidance.

planning/SPRINT_4_PLAN.md
```diff
@@ -0,0 +1,553 @@
...
+
+**1. Deployment Architecture + Environment Matrix**
+- Define the final Vercel + Railway topology:
+  - `frontend`
+  - `backend-api`
+  - `backend-worker`
+  - `postgres`
+  - `redis`
+- Document production env vars for frontend, backend API, and worker
+- Define domain plan (`frontend` domain + `api` subdomain)
```
AvanishKulkarni, line 113: **Cross-origin auth and cookie strategy is not specified anywhere in the plan and will block the production deploy.**

The topology puts the frontend on `<apex>` and the API on `api.<apex>`, which makes every authenticated request cross-origin. Nothing in Issue #1, #6, or #7 commits to:

1. **CORS configuration on the Express app.** Needs an explicit origin allowlist (production frontend domain + preview domains) with `credentials: true` and preflight handling for tRPC `POST` requests that send `content-type: application/json` and any auth headers. A wildcard origin cannot be combined with credentials, so this has to be enumerated.
2. **Cookie strategy for session/refresh tokens.** Pick one and document it here:
   - **Apex-scoped cookie:** issue cookies with `Domain=.<apex>`, `SameSite=Lax`, `Secure`, `HttpOnly`. Works cross-subdomain without `SameSite=None`, but requires the API and frontend to share a registrable domain.
   - **Host-only cookie on the API subdomain:** `SameSite=None; Secure; HttpOnly`. Requires `credentials: 'include'` on every fetch and is subject to third-party cookie restrictions in some browsers.
   - **Bearer token in `Authorization` header:** sidesteps cross-origin cookie handling entirely but requires client-side storage and refresh handling.
3. **CSRF posture under whichever cookie model is chosen.**

This is the single highest-probability cause of a "works on localhost, completely broken on the deployed domain" failure mode during grading. Please add an **Auth & CORS contract** bullet to the Explicit Production Decisions section and an acceptance criterion to Issue #1 requiring `docs/deployment/deployment-architecture.md` to record the chosen cookie domain, `SameSite`/`Secure` flags, CORS allowlist, and auth transport before Issues #6/#7/#11 land.

planning/SPRINT_4_PLAN.md
```diff
@@ -0,0 +1,553 @@
...
+- Replace or unify **both** current implementations:
+  - auth endpoint limiting using `express-rate-limit`
+  - public-route token bucket limiting
+- Ensure auth and public API rate limits remain correct across 2+ replicas
+- Configure Express proxy awareness so client IP handling works correctly behind Railway
+- Use `docs/deployment/replica-readiness-audit.md` as the implementation checklist and update it with final decisions
+- Acceptance criteria:
+  - Public and auth rate limits are shared across replicas
+  - No process-local auth or public-route limit counters remain in production code paths
+  - Production proxy trust is explicitly configured for a **single Railway hop** (`trust proxy = 1` or equivalent) and documented in deployment docs
```
AvanishKulkarni, line 166: **The Redis-backed limiter needs an atomicity requirement, not just a shared-storage requirement.**

The current acceptance criteria say limits must be "shared across replicas" and that no process-local counters remain, but they do not constrain *how* the Redis limiter is implemented. A naive port that does `INCR key` followed by a separate `EXPIRE key ttl` races under contention: if a process crashes or is preempted between the two commands, the key is left without a TTL and the counter grows unbounded, locking users out until the key is manually cleared. The public-route **token bucket** is worse — refill and consume must be a single atomic operation or two concurrent requests can both observe "tokens available" and both consume the last token.

Please add an acceptance criterion along the lines of:

> - The Redis limiter uses atomic primitives for all counter/bucket mutations (a single Lua script per check, or a vetted library such as `rate-limit-redis` for the auth limiter and an equivalent Lua-backed implementation for the public-route token bucket). Non-atomic `INCR` + separate `EXPIRE` patterns are explicitly disallowed.

Concrete implementation options:
- **Auth endpoint limiter (`express-rate-limit`):** swap the in-memory store for [`rate-limit-redis`](https://github.com/express-rate-limit/rate-limit-redis), which uses a Lua script to combine increment and expiry.
- **Public-route token bucket:** implement as a single Lua script loaded with `SCRIPT LOAD` that reads bucket state, computes refill based on elapsed time, decrements, and writes back — all inside one `EVALSHA`. Reference: the standard "token bucket in Redis" Lua pattern.

Without this, the limiter can pass the current acceptance criteria while still being unsafe under load.

planning/SPRINT_4_PLAN.md
```diff
@@ -0,0 +1,553 @@
...
+- Document which tests are:
+  - portable across both targets
+  - local-only because they depend on reset/seed control
+  - cloud-only because they validate deployed behavior
+- Capture/structure output for both local and cloud runs
+- Keep target URLs, environment contracts, and replica-sensitive validations aligned with `docs/deployment/deployment-architecture.md` and `docs/deployment/replica-readiness-audit.md`
+- Acceptance criteria:
+  - Tests run in a local configuration
+  - Tests run in a cloud configuration
+  - Cloud mode does not require local frontend/backend startup
```
AvanishKulkarni, line 306: **Cloud-target integration tests will run against the same Postgres instance the instructor reviews, and the plan has no data-hygiene strategy.**

Issue #9 defines `cloud` mode as "runs against already-deployed frontend/backend URLs," and Issue #14 runs the cloud-target suite as part of submission evidence. If the only deployed environment is production, then any test that creates users, channels, messages, or attachments will leave residue in the exact database the grader will see — or worse, mutate data the review flow depends on (visibility flips, message edits, deletes). The acceptance criteria as written (`Tests run in a cloud configuration`, `Cloud mode does not require local frontend/backend startup`) are satisfied by a suite that actively corrupts prod.

Pick one of the following and encode it in the Issue #8 spec and Issue #9 acceptance criteria:

1. **Stand up a second Railway environment (staging)** with its own Postgres/Redis and point cloud-mode tests at it. Highest fidelity; highest setup cost. Requires an extra `TARGET_ENV` knob in `deploy-railway.yml` or a separate workflow.
2. **Scope cloud mode to read-only smoke checks** — health endpoints, guest public channel rendering, public SSR metadata, SSE handshake (connect-then-disconnect without publishing), and canonical URL/sitemap fetches. All write-path coverage stays in local mode. Lowest cost; narrowest validation.
3. **Tagged test-tenant isolation** — all cloud-mode writes happen under a reserved `test-*` user/workspace namespace, with a cleanup step that runs before and after each suite via an authenticated admin endpoint. Requires the backend to expose a scoped cleanup API and careful handling so cleanup failures do not leak into the next run.

Add an acceptance criterion to Issue #9:

> - Cloud-mode tests declare a data-isolation strategy (staging environment, read-only scope, or tagged-tenant cleanup) and cannot mutate state outside that strategy. The choice is recorded in `docs/test-specs/` and in `docs/deployment/deployment-architecture.md`.

This needs to be picked before Issue #9 starts implementing, not during Issue #14.

planning/SPRINT_4_PLAN.md
```diff
@@ -0,0 +1,553 @@
...
+- Create backend CD workflow for Railway
+- Deploy `backend-api` and `backend-worker` on pushes to `main`
+- Ensure migrations / build / deploy ordering is safe
+- Use GitHub secrets for Railway authentication
+- Treat GitHub Actions as the production deploy authority and document any provider-native auto-deploy settings that must be disabled or restricted
+- Implement the workflow against the service topology and env contract defined in `docs/deployment/deployment-architecture.md`
+- Acceptance criteria:
+  - Workflow deploys backend services without manual intervention
+  - Production deploy authority is unambiguous and documented
+  - Exactly one migration runner performs `prisma migrate deploy`, and `backend-api` replicas are prevented from racing migrations on startup
```
AvanishKulkarni, line 345: **"Exactly one migration runner" closes the race-at-boot problem but does not cover destructive migrations under a rolling restart, and that gap will bite on the first column drop or NOT NULL tightening.**

Railway has no Heroku-style release phase that pauses request handling while a migration runs. Even with a single migration runner, the rollout sequence is:

1. Migration runner applies schema change.
2. Railway rolling-restarts `backend-api` replicas one at a time.
3. For a non-trivial window, **old code and new schema are running concurrently against the same database.**

Any destructive migration in that window will break the old replica:
- `DROP COLUMN foo` → old replicas SELECTing `foo` get runtime errors on every request touching that table.
- `ALTER COLUMN foo SET NOT NULL` → old replicas inserting rows without `foo` fail.
- `RENAME COLUMN foo TO bar` → old replicas fail on both reads and writes.
- Non-concurrent index creation → table lock that stalls old replicas' queries.

Please add an acceptance criterion to Issue #11 and a matching rule to `docs/deployment/deployment-architecture.md`:

> - Destructive schema changes follow an **expand/contract** convention: additive changes ship first (add nullable column, add new index `CONCURRENTLY`, dual-write code), the code migrates to the new shape, and only a later deploy removes the old column/constraint. Single-deploy destructive migrations (column drop, NOT NULL tightening, rename, non-concurrent index on large tables) are disallowed on `main`.

Concrete guidance worth documenting alongside:
- **Rename:** add new column, backfill, dual-write, cut readers over, drop old column in a later deploy.
- **Drop:** stop reading the column in code (deploy), then drop it (deploy).
- **NOT NULL tighten:** add with a default or backfill first, enforce `NOT NULL` in a later deploy.
- **Indexes on large tables:** `CREATE INDEX CONCURRENTLY` via a separate migration that does not run inside a transaction.

Without this rule written down before Issue #11 lands, the first team member to write a "small cleanup" migration will take down the API mid-rollout.

AvanishKulkarni (changes requested):
Requesting changes — see inline comments for critical unaddressed issues.

planning/SPRINT_4_PLAN.md
acabrera04: Fixed in `a2ed1c4`. Issue #11 now requires an expand/contract schema rollout rule so destructive migrations are not introduced in the same deploy that must coexist with old API replicas during Railway rolling restarts.

planning/SPRINT_4_PLAN.md
acabrera04: Fixed in `a2ed1c4`. The plan now picks read-only cloud smoke coverage as the default cloud-mode data-isolation strategy, requires `docs/test-specs/` to classify cloud tests accordingly, and adds acceptance criteria preventing cloud-mode tests from mutating non-isolated deployed state.

planning/SPRINT_4_PLAN.md
acabrera04: Fixed in `a2ed1c4`. The plan now adds an explicit Auth & CORS contract, chooses Bearer access tokens in the `Authorization` header as the browser-to-backend auth transport, records the required CORS allowlist and CSRF posture in `docs/deployment/deployment-architecture.md`, and adds matching acceptance criteria to Issues #1, #6, and #7.

planning/SPRINT_4_PLAN.md
acabrera04: Fixed in `a2ed1c4`. Issue #3 now explicitly requires atomic Redis-backed limiter mutations and disallows non-atomic `INCR` plus separate `EXPIRE` patterns for production rate-limit state.

acabrera04:
Addressed the latest inline deployment-plan review in `a2ed1c4`.

Included updates:
- explicit cross-origin auth and CORS contract using Bearer header auth
- atomic Redis limiter requirement
- read-only cloud smoke-test data-hygiene rule
- expand/contract schema rollout rule for rolling Railway deploys

I replied inline on each thread with the specific change.

AvanishKulkarni (approved):
Approving — all four change-requested issues are materially resolved in `a2ed1c4` (Auth & CORS contract with bearer-token transport, atomic Redis limiter requirement, read-only cloud smoke default, and expand/contract schema rollout rule).

**One polish ask, non-blocking:** the expand/contract rule on Issue #11 is stated at the rule level but doesn't enumerate what counts as destructive. When Issue #1 produces `docs/deployment/deployment-architecture.md`, please have it spell out the concrete patterns so the first team member to write a "small cleanup migration" has something to check against:

- **Rename column:** add new column → backfill → dual-write → cut readers over → drop old column in a later deploy.
- **Drop column:** stop reading the column in code (deploy) → drop it (later deploy).
- **Tighten to NOT NULL:** add the column with a default or backfill first → enforce `NOT NULL` in a later deploy.
- **Index on a large table:** use `CREATE INDEX CONCURRENTLY` in a migration that does not run inside a transaction.
- **Non-concurrent index / table rewrite on large tables:** disallowed on `main` without an explicit maintenance-window plan.

Consider adding an acceptance criterion to Issue #1 along the lines of: *"`docs/deployment/deployment-architecture.md` enumerates the expand/contract patterns for rename, drop, NOT NULL tightening, and index creation so contributors can self-check destructive migrations before opening a PR."*

Not a blocker for this PR — just the one follow-up needed to make the expand/contract rule enforceable in practice rather than in principle.

acabrera04:
Pushed `faba6e2` to update the sprint plan ownership and schedule.

Included changes:
- moved Railway-specific tasks off `acabrera04`
- moved Vercel-specific tasks to `declanblanc`
- pulled CD workflow work forward to this weekend since Railway and Vercel are already set up
- normalized the sprint timeline so work starts on April 10

declanblanc (approved):
Looks good to me as well!

PR #334, opened 2026-04-11
Fix password transport with auth verifier flow

### LINKED ISSUE #313: Fix: passwords sent unencrypted over the network
## Problem

Passwords are currently sent as plaintext in HTTP request bodies during login and registration. While HTTPS encrypts the transport layer, the passwords are visible in plaintext in request bodies, server logs, and any intermediary that terminates TLS.

## Suggested Fix

Hash or encrypt passwords on the client side before sending them to the backend. Options include:

- **Client-side hashing**: Hash the password in the frontend before sending (e.g., using the Web Crypto API with SHA-256 + a server-provided salt), then verify against the stored bcrypt hash on the backend
- **SRP (Secure Remote Password)**: A zero-knowledge proof protocol where the server never sees the password at all

## Affected Files

- `harmony-frontend/` — login and registration forms that send raw passwords
- `harmony-backend/src/services/auth.service.ts` — receives and processes plaintext passwords

acabrera04:
## Summary
- replace raw password transport with auth challenge endpoints and PBKDF2-derived password verifiers
- store verifier records on the backend and migrate the seeded loginable mock credential
- add regression coverage and align backend/auth specs with the new contract

## Testing
- npm --prefix harmony-backend test -- auth.test.ts auth.service.test.ts mock-seed.test.ts
- npm --prefix harmony-frontend test -- authService.test.ts issue-242-join-server-fix.test.ts
- tsc --noEmit (harmony-backend)
- tsc --noEmit (harmony-frontend)

## Notes
- Updated the DB-backed auth/server integration suites to the new contract, but did not run them in this worktree because `DATABASE_URL` is not configured here.


AvanishKulkarni (commented):
## Code Review

### Overview

This PR replaces raw password transport with a challenge/response pattern using client-side PBKDF2-SHA256 derivation. The flow is:

1. Client fetches a random salt from `/register/challenge` or `/login/challenge`
2. Client derives a verifier locally: `PBKDF2(password, salt, 310000, 32, SHA-256)`
3. Client sends `{ passwordVerifier }` instead of `{ password }`
4. Server stores credentials as `v1$<hex-salt>$<bcrypt(verifier)>`

This is a meaningful security improvement preventing raw passwords from appearing in request bodies, logs, or network captures.

---

### Code Quality & Correctness

**Encoding/Decoding Logic — Correct**
The `decodePasswordVerifierRecord` function correctly handles the double-`$` ambiguity. Since bcrypt hashes start with `$2b$...`, the stored record looks like `v1$<32-hex>$$2b$12$...`. The `indexOf('$', prefix.length)` call correctly locates the single separator between salt and bcrypt hash.

**PBKDF2 Parameters — Good**
310,000 iterations, SHA-256, 32 bytes — consistent between frontend (Web Crypto API) and backend (Node crypto). Meets current NIST guidance.

**Anti-enumeration — Good**
`createDummyPasswordSalt(email)` returns a deterministic SHA-256-derived salt for unknown emails, preventing account enumeration via the challenge endpoint.

**Versioned storage format — Good**
The `v1$` prefix makes future credential migrations tractable.

---

### Issues

#### High

**1. Significant test coverage regression unrelated to new contract**

The PR removes a large number of previously existing tests that have nothing to do with the password verifier change. Notable losses:

- `refreshTokens`: 9 tests → 2. Lost: tampered token, wrong-secret JWT, expired token, token reuse prevention, expiry boundary, JWT claim verification, DB storage verification
- `verifyAccessToken`: 6 tests → 2. Lost: tampered/wrong-secret/expired token rejection, no-DB-call invariant
- `logout`: 4 tests → 1. Lost: idempotency, unknown-token safety, hash selectivity
- `authService.register`: Lost `re-throws non-P2002 errors`, `joinServer soft-fail continues registration`, `displayName defaults to username`
- `authService.login`: Lost admin upsert field correctness, admin OWNER assignment, error propagation tests

These tests cover token lifecycle logic this PR does not change — removing them is a net coverage regression that should be restored.

**2. No migration path for existing deployed users**

Users with legacy bcrypt-only hashes (no `v1$` prefix) receive `UNAUTHORIZED` with a "reset your password" message, but there is no migration script or admin tooling to upgrade existing records. If any real user data exists in a deployed environment, those users will be locked out with no self-service path.

#### Medium

**3. Legacy hash path lacks timing equalization**

```typescript
// auth.service.ts login()
const decoded = decodePasswordVerifierRecord(user.passwordHash);
if (!decoded) {
  throw new TRPCError({ code: 'UNAUTHORIZED', ... });  // No bcrypt.compare called
}
```

For legacy-hash users, the code throws before calling `bcrypt.compare`, leaking timing information. An attacker can distinguish "legacy account exists" from "account does not exist." Should call `await bcrypt.compare(passwordVerifier, TIMING_DUMMY_HASH)` before throwing.

**4. Verifier replay attack surface not documented**

The scheme protects against raw password exposure in logs/request bodies, but a captured `passwordVerifier` can be directly replayed to `/api/auth/login`. This is weaker than SRP and the threat model should be documented (even a comment in the service) so future contributors understand why HTTPS is load-bearing for this scheme.

#### Low

**5. `passwordVerifierSchema` max length is loose**

```typescript
const passwordVerifierSchema = z.string().min(20).max(128);
```

PBKDF2-SHA256 with 32 bytes encodes to exactly 44 base64 characters. A max of 128 is unnecessarily permissive. Tightening to `.length(44)` would make the contract self-documenting and reject garbage inputs earlier.

---

### Suggested Changes Before Merge

1. **Restore removed tests** — particularly `refreshTokens`, `verifyAccessToken`, and `logout` edge cases. These don't interact with the verifier contract.
2. **Add timing equalization for the legacy hash path** — call dummy bcrypt before throwing on `!decoded`.
3. **Either provide a DB migration script** for existing users or explicitly document this is only valid for fresh deployments.
4. **Optionally tighten** `passwordVerifierSchema` to `.length(44)`.

acabrera04:
Addressed the auth review findings in `7d084cc`.

On the test removal point: that regression was my mistake. I rewrote `harmony-backend/tests/auth.service.test.ts` around the verifier contract and dropped unrelated token-lifecycle/admin-path cases that should have stayed intact. That was not intentional scope reduction.

Fixed in this follow-up:
- restored the missing `refreshTokens`, `verifyAccessToken`, `logout`, `register`, and admin-path auth service coverage
- added timing equalization for the legacy bcrypt-only path before returning the reset-required error
- tightened `passwordVerifier` validation to the exact 44-char base64 contract
- documented in the migration/spec that legacy bcrypt-only users need reset/upgrade tooling rather than automatic migration from this schema change alone

Re-verified locally after the fix:
- `npm --prefix harmony-backend test -- auth.test.ts auth.service.test.ts`
- `tsc --noEmit` in `harmony-backend/`
- `npm --prefix harmony-backend run lint` (same pre-existing warnings only, no new errors)

AvanishKulkarni (approved):
All five issues from the first review are addressed in the follow-up commit.

- **Test coverage**: All removed tests are restored — `refreshTokens`, `verifyAccessToken`, `logout`, `register`, and `login` edge cases are back in full.
- **Timing equalization**: `bcrypt.compare(passwordVerifier, TIMING_DUMMY_HASH)` now runs before throwing on legacy hashes, with a spy test verifying it.
- **Migration note**: Added to both the spec doc and the migration SQL — legacy bcrypt-only accounts require a password reset, fresh deployments are unaffected.
- **Replay attack documented**: Comment added above the `bcrypt.compare` call in `auth.service.ts` noting that HTTPS remains load-bearing.
- **Verifier schema**: Tightened to `.length(44)` with a self-documenting error message.

LGTM. Approving.

PR #372, opened 2026-04-16
fix: reconnect SSE EventSource with fresh token after drop (issue #366)

### LINKED ISSUE #366: Bug: Text does not always appear automatically across sessions on the same machine
## Summary
Text is not consistently showing up automatically when using different sessions on the same machine.

## Expected behavior
Text should appear automatically and consistently across sessions.

## Actual behavior
In some session combinations on the same machine, text does not appear automatically.

## Reproduction notes
1. Open multiple sessions on the same machine.
2. Switch between sessions / use concurrent sessions.
3. Observe that text intermittently does not show up automatically.

## Impact
Intermittent missing text creates confusion and can make conversations look empty or stale until a manual refresh/action occurs.

## Environment
- Same physical machine
- Multiple sessions

AvanishKulkarni:
## Summary

Fixes #366 — text not appearing automatically across sessions.

**Root cause:** When an `EventSource` connection dropped _after_ being successfully opened, the browser's native auto-retry reused the same URL — including the expired JWT access token baked into the query string. The server rejected every reconnect attempt with 401, leaving the session in a silent infinite retry loop. Since the `onerror` handler only closed the connection if it had _never_ successfully opened, a previously-working session would silently stop receiving all real-time events (messages, channel updates, member changes) until the user reloaded the page.

This is intermittent because it requires two conditions together: (1) the SSE connection drops for any reason (network blip, Vercel cold start, server restart), and (2) the 15-minute access token has since expired. Multi-session scenarios increase the likelihood because the refresh-token rotation can rotate the token in one tab, leaving another tab unable to silently refresh.

**Changes (`harmony-frontend`):**

- **`src/lib/api-client.ts`** — export `refreshAccessToken()`: a proactive token refresh that reuses the module-level `_isRefreshing`/`_refreshQueue` state so it coalesces with any concurrent axios interceptor refresh. Fails silently (no redirect, no `clearTokens`) — appropriate for background SSE reconnect.

- **`src/hooks/useChannelEvents.ts`** — on `onerror` after `everOpened`: close the EventSource immediately (stops the stale-token retry loop), call `refreshAccessToken()`, then schedule a reconnect via `reconnectKey` state increment. Up to 3 attempts with a growing 2 s × n delay. Resets count on successful `onopen`.

- **`src/hooks/useServerEvents.ts`** — identical fix for the server-scoped SSE stream.

## Test plan

- [x] All 208 existing frontend unit tests pass (`npm test` in `harmony-frontend/`)
- [x] TypeScript compiles without errors (`tsc --noEmit`)
- [ ] Manual: open two browser tabs on the same channel; send a message from one — verify it appears in the other without a reload
- [ ] Manual: let a session stay idle for >15 min; send a message from another tab — verify it appears after the reconnect delay (≤6 s for 3rd attempt)

### DIFF
--- a/harmony-frontend/src/hooks/useChannelEvents.ts
+++ b/harmony-frontend/src/hooks/useChannelEvents.ts
@@ -174,8 +174,8 @@ export function useChannelEvents({
         source: 'sse',
         target: '/api/events/channel/[channelId]',
       });
-      if (!everOpened) {
-        // Never successfully opened — likely a 401/403. Stop retrying.
+      if (!everOpened && reconnectCountRef.current === 0) {
+        // Never successfully opened on the first attempt — likely a 401/403. Stop retrying.
         es.close();
         return;
       }
--- a/harmony-frontend/src/hooks/useChannelEvents.ts
+++ b/harmony-frontend/src/hooks/useChannelEvents.ts
@@ -19,8 +19,11 @@
 import { useEffect, useLayoutEffect, useRef, useState } from 'react';
 import type { Message } from '@/types/message';
 import type { Server } from '@/types/server';
-import { getAccessToken } from '@/lib/api-client';
+import { getAccessToken, refreshAccessToken } from '@/lib/api-client';
 import { getApiBaseUrl } from '@/lib/runtime-config';
+
+const MAX_RECONNECT_ATTEMPTS = 3;
+const RECONNECT_DELAY_MS = 2_000;
 
 export interface UseChannelEventsOptions {
   channelId: string;
@@ -46,6 +49,12 @@ export function useChannelEvents({
   enabled = true,
 }: UseChannelEventsOptions): UseChannelEventsResult {
   const [isConnected, setIsConnected] = useState(false);
+  // Incrementing this triggers the effect to re-run with a fresh token after a
+  // dropped connection (e.g. token expiry). Capped at MAX_RECONNECT_ATTEMPTS.
+  const [reconnectKey, setReconnectKey] = useState(0);
+  // Tracks how many consecutive reconnect attempts have been made so we can
+  // apply a growing delay and bail out after repeated failures.
+  const reconnectCountRef = useRef(0);
 
   // Keep stable references to callbacks so the effect doesn't re-run on every render.
   // Updated via useLayoutEffect (before paint) so the EventSource handlers always call
@@ -119,9 +128,11 @@ export function useChannelEvents({
     // server or a network error before the stream started) — close immediately
     // instead of letting EventSource retry with a stale/invalid token.
     let everOpened = false;
+    let reconnectTimer: ReturnType<typeof setTimeout> | null = null;
 
     es.onopen = () => {
       everOpened = true;
+      reconnectCountRef.current = 0; // reset budget on successful connection
       setIsConnected(true);
     };
     es.onerror = () => {
@@ -129,10 +140,32 @@ export function useChannelEvents({
       if (!everOpened) {
         // Never successfully opened — likely a 401/403. Stop retrying.
         es.close();
+        return;
       }
+
+      // The connection was previously healthy but dropped (e.g. network blip,
+      // server restart, or the access token expired and the backend rejected the
+      // EventSource auto-reconnect attempt). Stop the native retry loop (which
+      // would keep hammering the server with the same stale token) and
+      // proactively refresh the token before reconnecting.
+      es.close();
+      const attempt = reconnectCountRef.current;
+      if (attempt >= MAX_RECONNECT_ATTEMPTS) return; // give up after cap
+
+      reconnectCountRef.current += 1;
+      const delay = RECONNECT_DELAY_MS * reconnectCountRef.current;
+      reconnectTimer = setTimeout(() => {
+        // Attempt a silent token refresh so the next EventSource URL carries a
+        // valid token even when the user has been idle (no API calls to trigger
+        // the axios interceptor refresh).
+        refreshAccessToken().finally(() => {
+          setReconnectKey(k => k + 1);
+        });
+      }, delay);
     };
 
     return () => {
+      if (reconnectTimer !== null) clearTimeout(reconnectTimer);
       es.removeEventListener('message:created', handleCreated);
       es.removeEventListener('message:edited', handleEdited);
       es.removeEventListener('message:deleted', handleDeleted);
@@ -140,7 +173,7 @@ export function useChannelEvents({
       es.close();
       setIsConnected(false);
     };
-  }, [channelId, enabled]);
+  }, [channelId, enabled, reconnectKey]);
 
   return { isConnected };
 }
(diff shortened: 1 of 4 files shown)


harmony-frontend/src/hooks/useChannelEvents.ts
```diff
@@ -166,18 +177,40 @@ export function useChannelEvents({
       if (!everOpened) {
```
acabrera04: [P1] Reconnect retries stop after the first failed re-open

After the initial healthy stream drops, this effect re-runs and `everOpened` is reset to `false`. If the replacement `EventSource` errors before `onopen` (for example during a server restart, cold start, or brief network outage), this branch closes the stream and returns without scheduling another retry, so the remaining retry budget is never used and channel updates stay dead until a full reload. Persist the 'opened once' state across reconnects, or treat pre-open errors as retryable once a reconnect sequence is already in progress.

harmony-frontend/src/hooks/useServerEvents.ts
```diff
@@ -232,10 +243,26 @@ export function useServerEvents({
       if (!everOpened) {
```
acabrera04: [P1] Server event stream has the same one-shot reconnect failure

The same reconnect bug exists here. `everOpened` is local to a single effect run, so once `setReconnectKey()` creates a new `EventSource`, any pre-open failure on that new connection hits this early return and permanently stops retrying. A short backend restart can therefore kill server-scoped real-time updates even though `MAX_RECONNECT_ATTEMPTS` suggests the hook will keep trying.

acabrera04 (commented):
- [ ] Prioritize logic over style: the reconnect logic has a real runtime regression in both SSE hooks.
- [x] Security first: no new obvious secret exposure or input-validation issue stood out in the reviewed changes.
- [x] Architectural alignment: the change stays within the existing frontend auth/SSE patterns and does not conflict with the repo’s documented backend architecture.
- [ ] Issue completion: the PR fixes the stale-token loop, but it can still leave SSE permanently disconnected after a normal transient outage.
- [x] No nitpicking: feedback is limited to behavior/regression risk.
- [x] Avoid repetition: this review does not repeat the existing posted comment about missing tests.
- [x] Iterative reviews: no prior review comments from this reviewer needed re-checking.
- [x] Prevent CI failures: no obvious syntax/import/build break was found; the concern is runtime behavior, not pipeline breakage.

Bundled review:
The reconnect strategy regresses resilience after transient failures. In both `useChannelEvents` and `useServerEvents`, once a healthy stream drops you close the native retry loop and schedule a manual reconnect, but that reconnect happens in a fresh effect run where `everOpened` starts over as `false`. If the replacement `EventSource` fails before `onopen` because the backend is still restarting or the network is still unstable, the hook immediately closes the stream and gives up, so real-time updates stay broken until reload. That undermines the stated goal of recovering from server restarts/cold starts.

declanblanc (approved):
## PR Review Checklist

- [x] **1. Logic over Style:** Reconnect logic is correct. `reconnectCountRef` is read before increment, reset on successful `onopen`, and the budget cap is checked before scheduling a retry. Timer cleanup is handled in both the effect cleanup and via `useEffect` unmount cleanup.
- [x] **2. Security First:** `refreshAccessToken` silently swallows refresh failures rather than redirecting or clearing tokens — intentional and correct per the doc comment. The function guards against SSR (`typeof window === 'undefined'`) and missing refresh token.
- [x] **3. Architectural Alignment:** Correctly extends the existing `_isRefreshing`/`_refreshQueue` coalescing pattern in `api-client.ts` rather than issuing a parallel refresh request.
- [x] **4. Issue Completion:** #366 AC (SSE reconnects with a fresh token after drop) is met for both `useChannelEvents` and `useServerEvents`.
- [x] **5. No Nitpicking:** One structural observation below; it is not a blocker.
- [x] **6. Avoid Repetition:** First human review on this PR.
- [x] **7. Iterative Reviews:** N/A — first review.
- [x] **8. Prevent CI Failures:** No new tests were added for the reconnect path (hard to unit-test SSE lifecycle), but the existing test suite is unaffected.

---

## Review

The reconnect logic is sound and the token coalescing is the right approach. One follow-up worth tracking:

### [Non-blocking] Reconnect logic is duplicated across `useChannelEvents` and `useServerEvents`

The `reconnectKey` state, `reconnectCountRef`, timer cleanup, and `onerror` reconnect block are near-identical in both hooks. A shared `useSSEReconnect` utility (or a single `useSSEConnection` hook) would eliminate the duplication and make future changes (e.g. adjusting `MAX_RECONNECT_ATTEMPTS`) a one-line edit. This is not a blocker for this PR but is worth a follow-up issue.

harmony-frontend/src/hooks/useChannelEvents.ts
AvanishKulkarni: Fixed in e08178e. The `!everOpened` guard now also checks `reconnectCountRef.current === 0` before bailing out. When a replacement EventSource fails before `onopen` during a reconnect sequence (`reconnectCountRef.current > 0`), it falls through to the retry logic and schedules the next attempt — preserving the full retry budget.

harmony-frontend/src/hooks/useServerEvents.ts
AvanishKulkarni: Fixed in e08178e. Same fix applied to `useServerEvents`: the `!everOpened` guard now checks `reconnectCountRef.current === 0` so reconnect-sequence failures on a not-yet-opened EventSource fall through to the retry logic instead of silently giving up.

AvanishKulkarni:
> Re: comment 4122300950

Fixed in e08178e. Both `useChannelEvents` and `useServerEvents` now check `reconnectCountRef.current === 0` alongside `!everOpened` before closing without retry. Reconnect-sequence attempts that fail before `onopen` (transient outage, backend restart) correctly continue exhausting the retry budget instead of giving up silently.

AvanishKulkarni:
> Re: comment 4123819139

Won't fix (in this PR): The duplication is acknowledged and the refactor into a shared `useSSEReconnect` utility is a good follow-up. Keeping it out of scope here to minimize risk to the reconnect fix. Will track as a follow-up issue.

harmony-frontend/src/hooks/useChannelEvents.ts
```diff
@@ -150,9 +159,11 @@ export function useChannelEvents({
     // server or a network error before the stream started) — close immediately
     // instead of letting EventSource retry with a stale/invalid token.
     let everOpened = false;
+    let reconnectTimer: ReturnType<typeof setTimeout> | null = null;
```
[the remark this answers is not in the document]
AvanishKulkarni: Won't fix: SSE lifecycle (onopen → onerror → close → re-create) is difficult to reliably test with Jest fake timers given jsdom's limited EventSource support. The human reviewer (declanblanc) also noted this. The reconnect bug itself is fixed in e08178e; unit test coverage for this lifecycle can be addressed in a follow-up once a suitable testing strategy is agreed on.

Issue #338, opened 2026-04-13
Private channel access should lock the message pane, not replace the whole page


acabrera04:
## Summary
Opening a private channel currently replaces the entire channel page with a full-screen access-denied screen. That blocks normal navigation and makes it harder for the user to move to another channel.

## Reproduction
1. Sign in as a user who can browse a server but does not have access to one of its private channels.
2. Navigate directly to that private channel from the URL or any stale/deep link.
3. Observe the rendered page state.

## Expected
The app should keep the normal shell/navigation visible and only lock the chat area for the private channel.

That means the user should still be able to:
- see the server rail / channel navigation
- switch to another channel immediately
- understand that the current channel is private without losing the surrounding layout

## Actual
The app renders a full-screen `This channel is private` page with CTA buttons and `Go Back`, replacing the normal shell. Navigation away from the locked channel becomes harder than it needs to be.

## Impact
- Users lose access to channel navigation at the moment they most need it.
- Direct links into private channels become a UX dead end instead of a recoverable state.
- The current experience adds unnecessary friction compared with simply locking the message pane/composer.

## Evidence
- `harmony-frontend/src/components/channel/ChannelPageContent.tsx:62-83` builds the full `HarmonyShell` and then wraps the entire shell in `VisibilityGuard`.
- `harmony-frontend/src/components/channel/VisibilityGuard.tsx:82-135` renders `AccessDeniedPage` as a full-screen replacement UI.
- `harmony-frontend/src/components/channel/VisibilityGuard.tsx:194-205` returns `AccessDeniedPage` for private channels when the user is unauthenticated or lacks server admin/owner access.
- `harmony-frontend/src/components/channel/GuestChannelView.tsx:121-128` also routes private-channel handling through the same full-page guard behavior.

## Suggested Direction
Refactor the private-channel state so the page shell stays mounted and only the message region is replaced with a locked-state panel. The sidebar/server navigation should remain interactive so the user can leave the blocked channel without relying on browser history.

acabrera04:
**[AI Agent — Starting]**

Starting work on the private-channel access regression. I’m reproducing the current full-page denial behavior, then I’ll add failing frontend regression tests first and implement the smallest fix that preserves the shell/navigation while locking only the message pane.

I’ll post another update once the red tests are in place or if I hit a blocker.

acabrera04:
**[AI Agent — Complete]**

[...]

PR #379, opened 2026-04-16
[codex] Fix private channel locked pane UX

acabrera04:
## What changed
- kept the standard channel shell mounted for locked private channels and swapped only the center chat pane for a private-channel lock panel
- added a shared `PrivateChannelLockedPane` used by authenticated and guest private-channel denial states
- added a regression test to lock in the pane-only behavior for issue #338

## Why
Private channel deep links were replacing the entire page with an access-denied screen, which removed the surrounding navigation at the moment a user needed it most. This change preserves the server rail, channel list, and top bar so users can immediately navigate away.

## Root cause
`ChannelPageContent` wrapped the entire `HarmonyShell` in `VisibilityGuard`, and the guard rendered a full-screen denial UI for private channels. That made access denial a page-level replacement instead of a message-pane state.

## Impact
- private-channel denial is now recoverable without relying on browser history
- server/channel navigation remains visible and interactive
- the regression test will fail if the shell is replaced again in the future

## Validation
- `npm test -- --runTestsByPath src/__tests__/issue-338-private-channel-pane-lock.test.tsx src/__tests__/VisibilityGuard.test.tsx`
- `npm run lint`
- `npm run build`


declanblanc (approved):
## PR Review Checklist

- [x] **1. Logic over Style:** Replacing a full-screen denial UI with a pane-level lock while keeping the shell mounted is the correct fix. `lockedMessagePane` prop is clean and the `isChannelLocked` derived boolean gates all affected behaviors (SSE, Ctrl+K, pins panel, Search modal) consistently.
- [x] **2. Security First:** Messages are not fetched for locked channels (`sortedMessages` short-circuits to `[]`). SSE subscription is disabled while locked. No content leaks through disabled UI paths.
- [x] **3. Architectural Alignment:** `PrivateChannelLockedPane` correctly consolidates `AccessDeniedPage` and `NoPermissionPage` from `VisibilityGuard`. The `mode` prop keeps the guest/member distinction without duplicating the lock icon and layout.
- [x] **4. Issue Completion:** Issue #338 AC (shell stays mounted for private channel denial) is met. Regression test locks in the behavior.
- [x] **5. No Nitpicking:** One correctness question below.
- [x] **6. Avoid Repetition:** First human review on this PR.
- [x] **7. Iterative Reviews:** N/A — first review.
- [x] **8. Prevent CI Failures:** New test file and updated `VisibilityGuard.test.tsx` pass per test plan.

---

## Review

The structural change is correct and the test coverage is solid. One thing worth confirming:

### [`ChannelPageContent.tsx`, lines 266–268] `currentUser?.isSystemAdmin` added to `isServerAdmin`



The original code deliberately excluded any global-role check here, relying only on the server-scoped member role. Adding `isSystemAdmin` means system admins will bypass the locked pane on all private channels, which is probably the right behavior — but it's a behavioral change from the original intent. Please confirm this is intentional and that `currentUser.isSystemAdmin` is reliably set by the auth path before this component runs.

declanblanc:
> Re: comment 4123821758

Confirmed intentional — system admins should bypass the lock on all private channels. The fix in 2e9cac6 also addresses the reliability concern: the original code called authService.getCurrentUser() (client-side axios client reading in-memory token + localStorage) from a Server Component where both are always null, so currentUser was always null server-side. The locked pane was showing for everyone, including admins — this was the root cause of the failing E2E test. Fixed by adding getSessionUser() to trpc-client.ts (reads the httpOnly auth_token cookie set by AuthContext.login()) and using that in ChannelPageContent.tsx instead.

PR #387, opened 2026-04-16
feat: wire up frontend file attachment upload UI

### LINKED ISSUE #386: Sprint 5: Wire up frontend file attachment upload UI
## Summary

The attachment upload button exists in the message input UI but is currently non-functional — clicking it does nothing. Now that the backend attachment endpoint and production storage provider (R2) are implemented (#319), the frontend needs to be wired up to complete the feature.

## Acceptance Criteria

- Clicking the attachment button opens a file picker
- Selecting a file uploads it to `POST /api/attachments/upload` using the user's auth cookie
- The returned URL is inserted into the message input (or sent as a message automatically)
- Upload errors are surfaced to the user (e.g. unsupported file type, size limit exceeded)
- UI shows a loading/progress state while the upload is in flight

## Dependencies

- Blocked by: #319 (backend storage provider — merged)

declanblanc:
## Summary

Closes #386

- Added `AttachmentInput` type to the frontend types barrel for the upload response shape
- Created a Next.js route handler at `/api/attachments/upload` to proxy multipart POSTs server-side — necessary because the auth token lives in an httpOnly cookie that can't be read client-side, so the handler reads it via `cookies()` and forwards it as a Bearer header to the Express backend
- Extended `messageService.sendMessage` and `sendMessageAction` to accept and forward an optional `AttachmentInput[]` to the tRPC `message.sendMessage` mutation (which already supports attachments)
- Wired the attachment button in `MessageInput.tsx` with the full upload UX

## Behavior

1. Clicking the attachment button opens a native file picker (filtered to allowed MIME types)
2. Selecting a file immediately POSTs it to `/api/attachments/upload`; the button shows a spinner while the upload is in flight
3. On success, a dismissible filename chip appears above the input
4. On error (401, 413, unsupported type, network failure), a message appears in the existing error area
5. On send, pending attachments are forwarded to the backend and cleared after the message is created

## Test plan

- [ ] All 214 existing frontend tests pass (`npm test` in `harmony-frontend/`)
- [ ] TypeScript compiles cleanly (`npx tsc --noEmit`)
- [ ] Attach a supported file (image/PDF) → chip appears, message sends with attachment
- [ ] Attach an oversized or unsupported file → error message shown, no chip added
- [ ] Dismiss a chip via × → chip removed, not sent
- [ ] Send without attaching anything → works as before (no regression)
- [ ] Switch channels mid-compose → pending chips cleared

### DIFF
--- a/harmony-backend/src/trpc/routers/message.router.ts
+++ b/harmony-backend/src/trpc/routers/message.router.ts
@@ -37,7 +37,7 @@ export const messageRouter = router({
       z.object({
         serverId: z.string().uuid(),
         channelId: z.string().uuid(),
-        content: z.string().min(1).max(4000),
+        content: z.string().max(4000),
         attachments: z.array(AttachmentInputSchema).max(10).optional(),
       }),
     )
--- a/harmony-backend/src/trpc/routers/message.router.ts
+++ b/harmony-backend/src/trpc/routers/message.router.ts
@@ -39,7 +39,10 @@ export const messageRouter = router({
         channelId: z.string().uuid(),
         content: z.string().max(4000),
         attachments: z.array(AttachmentInputSchema).max(10).optional(),
-      }),
+      }).refine(
+        data => data.content.trim().length > 0 || (data.attachments?.length ?? 0) > 0,
+        { message: 'Message must have content or at least one attachment.' },
+      ),
     )
     .mutation(({ input, ctx }) =>
       messageService.sendMessage({
--- a/harmony-frontend/src/app/actions/sendMessage.ts
+++ b/harmony-frontend/src/app/actions/sendMessage.ts
@@ -1,6 +1,6 @@
 'use server';
 
-import type { Message } from '@/types';
+import type { Message, AttachmentInput } from '@/types';
 import { sendMessage as sendMessageService } from '@/services/messageService';
 
 /**
@@ -12,6 +12,7 @@ export async function sendMessageAction(
   channelId: string,
   content: string,
   serverId: string,
+  attachments?: AttachmentInput[],
 ): Promise<Message> {
-  return sendMessageService(channelId, content, serverId);
+  return sendMessageService(channelId, content, serverId, attachments);
 }
--- a/harmony-frontend/src/app/api/attachments/upload/route.ts
+++ b/harmony-frontend/src/app/api/attachments/upload/route.ts
@@ -17,6 +17,11 @@ export async function POST(request: NextRequest) {
     body: formData,
   });
 
-  const data = await res.json();
+  let data: unknown;
+  try {
+    data = await res.json();
+  } catch {
+    data = { error: 'Upload failed.' };
+  }
   return NextResponse.json(data, { status: res.status });
 }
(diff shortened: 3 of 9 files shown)


AvanishKulkarni (changes requested):

- [x] **Prioritize Logic over Style** — Code is DRY and logically sound. The `getMessages` refactor from try/catch fallback to explicit branching is cleaner and intentional.
- [ ] **Security First** — One unguarded `res.json()` call in the proxy route (see below).
- [x] **Architectural Alignment** — Proxy pattern for httpOnly cookie forwarding is the correct approach; tRPC mutation extension follows existing patterns.
- [x] **Issue Completion** — All five acceptance criteria from #386 are satisfied.
- [x] **No Nitpicking** — Only critical logic and correctness issues raised.
- [x] **Prevent CI Failures** — All 6 CI checks pass. TypeScript compiles cleanly.

---

## Review

The overall implementation is solid — the proxy pattern, state management, and UX flow are all well-reasoned. Two issues need to be addressed before merge.

### Issue 1 — Unguarded `res.json()` in proxy route (Bug)

**`harmony-frontend/src/app/api/attachments/upload/route.ts`, line 17:**

```ts
const data = await res.json();
return NextResponse.json(data, { status: res.status });
```

If the Express backend (or an intermediary like nginx) returns a non-JSON body — for example a plain-text `413 Request Entity Too Large` from nginx before the request reaches Express — `res.json()` throws, which propagates as an unhandled 500 to the client. The client-side error handler in `MessageInput` then shows "Upload failed. Please try again." instead of the more helpful "File is too large" message.

Fix:

```ts
let data: unknown;
try {
  data = await res.json();
} catch {
  data = { error: 'Upload failed.' };
}
return NextResponse.json(data, { status: res.status });
```

### Issue 2 — Server-side validation gap after removing `content: z.string().min(1)` (Logic)

**`harmony-backend/src/trpc/routers/message.router.ts`, line 40:**

```ts
- content: z.string().min(1).max(4000),
+ content: z.string().max(4000),
```

Removing `min(1)` is required to support attachment-only messages — correct. However, the schema now accepts `{ content: "", attachments: undefined }` (empty message, no attachments), which the frontend guards against in `handleSend` but the backend does not. A malformed or replayed request could create a blank message with no attachments.

Add a `.superRefine` or a `.refine` to enforce that at least one of `content` or `attachments` is non-empty at the schema level:

```ts
z.object({
  serverId: z.string().uuid(),
  channelId: z.string().uuid(),
  content: z.string().max(4000),
  attachments: z.array(AttachmentInputSchema).max(10).optional(),
}).refine(
  data => data.content.trim().length > 0 || (data.attachments?.length ?? 0) > 0,
  { message: 'Message must have content or at least one attachment.' },
)
```

---

Both issues are straightforward fixes. Everything else — the `AttachmentList` component, chip UI, channel-switch clearing, spinner state, and `getMessages` branching refactor — looks correct and well-implemented.

declanblanc:
Both issues addressed in f48dfc8:

**Issue 1** — Wrapped `res.json()` in a try/catch in the upload proxy route so non-JSON responses (e.g. nginx 413) return a safe `{ error: 'Upload failed.' }` body instead of propagating a 500.

**Issue 2** — Added `.refine()` to the `sendMessage` input schema enforcing that at least one of `content` (non-empty after trim) or `attachments` is present, closing the server-side gap.

AvanishKulkarni (approved):
Both issues from my previous review have been addressed in the latest commit (`f48dfc8`):

- **Proxy JSON parse guard**: `res.json()` is now wrapped in try/catch in `route.ts`, with a fallback `{ error: 'Upload failed.' }` for non-JSON responses (e.g. nginx 413).
- **Server-side validation**: `.refine()` added to the `sendMessage` input schema, rejecting requests with empty content and no attachments at the tRPC layer.

Approved.

PR #397, opened 2026-04-17
[codex] Fix sitemap host rewriting for issue #371

### LINKED ISSUE #371: sitemap.xml outputs hardcoded harmony.chat URLs in local environment
## Summary
`http://localhost:3000/sitemap.xml` emits `<loc>` entries under `https://harmony.chat/...` instead of the runtime host.

## Expected
Sitemap host generation should align with runtime/base URL configuration for the current environment.

## Actual
Sitemap index uses hardcoded production domain values in local testing.

## Repro Steps
1. Open `http://localhost:3000/sitemap.xml`.
2. Observe `<loc>` values (e.g. `https://harmony.chat/sitemap/harmony-hq`).
3. Compare against current local host context (`http://localhost:3000`).

## Artifacts
- Report: `dogfood-output/harmony-local-20260415/report.md` (ISSUE-005)
- Screenshot: `dogfood-output/harmony-local-20260415/screenshots/issue-005.png`
- Raw XML: `dogfood-output/harmony-local-20260415/screenshots/issue-005-sitemap.xml`

acabrera04:
## Summary
- rewrite crawler-facing sitemap XML `<loc>` entries to the frontend request host instead of leaking the backend's configured origin
- apply the same host rewrite to both the sitemap index route and per-server sitemap route
- add regression coverage for local-host sitemap rewriting so issue #371 stays fixed

## Root Cause
The frontend sitemap routes proxied backend XML unchanged. When the backend generated sitemap XML using its `BASE_URL` fallback of `https://harmony.chat`, local requests to `http://localhost:3000/sitemap.xml` still emitted production-host `<loc>` entries.

## Impact
Local and preview sitemap responses now match the actual frontend host that served the request, which keeps crawler-facing sitemap output aligned with the runtime environment.

## Verification
- `npm test -- --runTestsByPath src/__tests__/seo-routes.test.ts`
- `npx tsc --noEmit`
- `npx eslint src/app/sitemap.xml/route.ts 'src/app/sitemap/[serverSlug]/route.ts' src/lib/seo-sitemap.ts src/__tests__/seo-routes.test.ts`
- direct route repro before/after confirming `https://harmony.chat/...` sitemap `<loc>` entries rewrite to `http://localhost:3000/...` locally

Closes #371


declanblanc (commented):
## Review — Sitemap host rewriting (#371)

### Overview
Fixes a bug where sitemap XML returned by the frontend routes leaked the backend's configured `BASE_URL` (e.g. `https://harmony.chat`) into `<loc>` entries, even when the request actually hit `http://localhost:3000`. The fix adds a new `rewriteSitemapLocOrigins` helper that replaces the host/protocol of every `<loc>` value with the host from the incoming `Request`, and wires it into both `/sitemap.xml` and `/sitemap/[serverSlug]` route handlers.

### Correctness

**Overall approach is sound.** Fixing at the frontend layer is the right move since the frontend owns the public SEO surface and the backend shouldn't need to know about preview/local hosts.

**Potential issue: `request.url` behind a proxy**
On Vercel, `request.url` typically reflects the public host, but behind some proxy configurations the `Host` header (and thus `request.url`) may not. Consider checking `x-forwarded-host` / `x-forwarded-proto` as a fallback:
```ts
const host = request.headers.get('x-forwarded-host') ?? new URL(request.url).host;
const proto = request.headers.get('x-forwarded-proto') ?? new URL(request.url).protocol.replace(':', '');
```
Not strictly required if deployment is only Vercel, but worth a sanity check.

**`getRequestOrigin` is overengineered**
```ts
function getRequestOrigin(request: Request): string {
  const url = new URL(request.url);
  url.pathname = '';
  url.search = '';
  url.hash = '';
  return url.toString().replace(/\/$/, '');
}
```
Equivalent to `new URL(request.url).origin`. The current code also roundtrips back through `new URL(getRequestOrigin(request))` in the caller, which is wasteful. Simplified:
```ts
export function rewriteSitemapLocOrigins(xml: string, request: Request): string {
  const { protocol, host } = new URL(request.url);
  return xml.replace(/<loc>([^<]+)<\/loc>/g, (_match, rawUrl: string) => {
    try {
      const url = new URL(rawUrl);
      url.protocol = protocol;
      url.host = host;
      return `<loc>${url.toString()}</loc>`;
    } catch {
      return `<loc>${rawUrl}</loc>`;
    }
  });
}
```

**Regex XML parsing is fragile but acceptable here**
Sitemap XML is generator-controlled (your own backend) with no CDATA/entity encoding expected inside `<loc>` — regex is fine. If a future backend change wraps URLs in `<![CDATA[...]]>`, this silently fails; worth a comment noting the assumption.

**Stream → text conversion**
The original code passed `response.body` through as a stream; the new code reads full text. Sitemaps are bounded (50k URLs / 50MB per spec) so this is fine, but it does eliminate streaming. No action needed; just noting.

### Caching

The `Cache-Control: public, max-age=300` header is preserved. Since the response now varies by request host, ensure the CDN keys responses by host (Vercel does by default for separate domains, but preview URLs all share the same Vercel origin — caching across previews is unlikely here since each preview gets its own subdomain).

### Tests

- First existing test (`proxies the sitemap index...`) was repurposed to also test the rewrite. The name no longer matches its behavior — rename to something like `rewrites the sitemap index to the current frontend host`.
- Good coverage: one test for each route + one explicit local-host rewrite. Missing: a case where `<loc>` contains an invalid URL (exercises the `catch` fallback) and a case where the backend returns a non-2xx (currently the new code only runs the rewrite on success paths, which is correct — but no test confirms the pass-through error path wasn't broken by moving from `response.body` to `response.text()`).

### Security
No injection risk — the rewritten URL comes from parsing a `URL`, which throws on malformed input. The `Host` header from the incoming request is used for rewriting, which means an attacker setting a crafted `Host` could get the frontend to emit a sitemap with their host in `<loc>`. This is generally low-impact (crawlers verify origin), but worth considering whether to allow-list hosts for production.

### Recommendation
Approve with minor suggestions: simplify `getRequestOrigin`, rename the repurposed test, optionally consider `x-forwarded-*` headers if non-Vercel deployments are in scope. Functional fix is correct and well-tested.

acabrera04:
> Re: comment 4132592219

Fixed in fbfa44e. Simplified the sitemap host rewrite helper to use the request URL directly, renamed the sitemap-index test to match its rewritten-host behavior, and added an inline note documenting the generator-controlled XML assumption. I did not add x-forwarded-* handling because this frontend SEO surface is currently deployed on Vercel, where request.url is the public host for these route handlers.

declanblanc (approved):
Follow-ups look good —  removed in favor of direct , test renamed, and an inline comment now documents the assumption that backend XML uses raw absolute URLs (no CDATA). Approving.

declanblanc (approved):
Follow-ups look good — `getRequestOrigin` removed in favor of direct `new URL(request.url)`, test renamed, and an inline comment now documents the assumption that backend XML uses raw absolute URLs (no CDATA). Approving.

PR #398, opened 2026-04-17
[codex] Add signup empty-data regression coverage

### LINKED ISSUE #376: Sign-up form submit is a no-op in empty-data environment
## Summary
Sign-up submission is non-functional in an empty-data environment. Clicking **Continue** does not create an account, does not navigate, and does not surface any validation/server error feedback.

## Environment
- App URL: `http://localhost:3000`
- Backend DB state: reset with migrations, no mock seed
- Browser sessions used: `harmony-user-a` and `harmony-user-b`

## Steps to Reproduce
1. Open `http://localhost:3000/auth/signup`.
2. Enter valid values for required fields (EMAIL, USERNAME, PASSWORD).
3. Click **Continue**.

## Expected
A registration request is submitted and the user is either created (then redirected) or shown actionable validation/server error feedback.

## Actual
Nothing happens: page remains on `/auth/signup`, no visible feedback, and no auth submit request appears.

## Artifacts
- Repro video: [issue-001-repro.webm](https://github.com/CS485-Harmony/Harmony/blob/dogfood-artifacts-no-mock-2026-04-16/dogfood-output/no-mock-2026-04-16/videos/issue-001-repro.webm)
- Network capture: [issue-001-network.txt](https://github.com/CS485-Harmony/Harmony/blob/dogfood-artifacts-no-mock-2026-04-16/dogfood-output/no-mock-2026-04-16/screenshots/issue-001-network.txt)

### Step Screenshots
1. Issue 001 Step 1
2. Issue 001 Step 2
3. Issue 001 Step 3
4. Issue 001 Step 4
5. Issue 001 Step 5
6. Issue 001 Result

acabrera04:
## Summary
- add backend regression coverage for signup when no default server exists
- document the actual root cause behind issue #376: the attached repro left the required password field blank, so the browser never submitted the form
- keep the fix surface limited to test coverage because the product path already behaves correctly

## Root cause
Issue #376 reported signup as a no-op in an empty-data environment, but the attached screenshots show native browser validation blocking submit on an empty password field. No registration request fired because the form was invalid before submit.

## Impact
This locks down the real empty-data behavior on the backend so registration still succeeds even when there is no default server to auto-join.

## Validation
- `npm test -- --runInBand tests/auth.service.test.ts`
- `npx eslint tests/auth.service.test.ts`

Closes #376.

### DIFF
--- a/harmony-backend/tests/auth.service.test.ts
+++ b/harmony-backend/tests/auth.service.test.ts
@@ -249,6 +249,21 @@ describe('authService.register', () => {
     expect(mockJoinServer).toHaveBeenCalledWith(mockUserId, 'server-001');
   });
 
+  it('still returns tokens when no default server exists in an empty-data environment', async () => {
+    mockPrisma.server.findFirst.mockResolvedValue(null);
+
+    const result = await authService.register(
+      'user@example.com',
+      'testuser',
+      PASSWORD_SALT,
+      derivePasswordVerifier('SecurePass123!'),
+    );
+
+    expect(result.accessToken).toBeTruthy();
+    expect(result.refreshToken).toBeTruthy();
+    expect(mockJoinServer).not.toHaveBeenCalled();
+  });
+
   it('continues registration when joinServer fails', async () => {
     mockPrisma.server.findFirst.mockResolvedValue({ id: 'server-001' });
     mockJoinServer.mockRejectedValue(new Error('Server join failed'));
--- a/harmony-backend/tests/auth.service.test.ts
+++ b/harmony-backend/tests/auth.service.test.ts
@@ -249,7 +249,7 @@ describe('authService.register', () => {
     expect(mockJoinServer).toHaveBeenCalledWith(mockUserId, 'server-001');
   });
 
-  it('still returns tokens when no default server exists in an empty-data environment', async () => {
+  it('returns tokens when no default server exists and skips joinServer', async () => {
     mockPrisma.server.findFirst.mockResolvedValue(null);
 
     const result = await authService.register(


declanblanc (commented):
## Review — Signup empty-data regression coverage (#376)

### Overview
Issue #376 (already closed) reported signup being a no-op in an empty-data environment. The PR author's root-cause analysis concludes the repro actually showed native browser validation blocking submit on an empty password field — no registration request ever fired — rather than a real backend bug. This PR adds one regression test to `auth.service.test.ts` to lock down the empty-data path: when `server.findFirst` returns no default server, `register` must still return tokens and must not call `joinServer`.

### Code quality
- Test fits cleanly into the existing `describe('authService.register')` block and mirrors the style of the adjacent `calls joinServer when the default server exists` and `continues registration when joinServer fails` tests.
- Assertions are appropriate: tokens returned + `joinServer` not called. Nothing missing for this scope.
- No production code changes — zero blast radius.

### Concerns

**Root-cause claim is plausible but not directly verified here**
The PR body says the repro video shows the browser blocking submit on an empty password field. This can't be confirmed from the diff alone. The issue is already closed and a regression test is always welcome, but if the team hasn't independently reviewed the repro video, there's a residual risk that the real bug is in the frontend form submission and this test gives false confidence. A frontend test asserting the form does POST `/api/auth/register` with valid credentials in an empty-data env would close the loop more rigorously. Not required for this PR, but worth a follow-up if no such coverage exists.

**Test name is slightly aspirational**
`still returns tokens when no default server exists in an empty-data environment` — the \"empty-data environment\" framing describes the motivation, not the mechanism. Consider `returns tokens when no default server exists and skips joinServer`. Minor.

### Tests / security / performance
- No perf implications.
- No security implications — test-only change with no new codepaths.
- No new dependencies.

### Recommendation
Approve. The test is small, correct, and valuable as a regression guard. Suggest a follow-up (separate PR or issue) to add a frontend test that asserts the signup form actually submits when fields are validly filled, which is the layer where the reported UX symptom actually lives.

acabrera04:
> Re: comment 4132593565

Fixed in e9e2ec8. I renamed the test to 'returns tokens when no default server exists and skips joinServer' so it describes the actual mechanism under test. I kept the frontend-submit coverage as a follow-up rather than broadening this PR, since this diff is intentionally scoped to the backend no-default-server regression guard.

declanblanc (approved):
Test rename applied — name now describes the mechanism instead of the motivation. Approving.

PR #401, opened 2026-04-18
[codex] Fix server settings authorization bypass

### LINKED ISSUE #400: Client-side server settings auth can be bypassed by spoofing user.getCurrentUser isSystemAdmin
## Summary
Any authenticated user can bypass the `/settings/[serverSlug]` client-side redirect by tampering with the `user.getCurrentUser` response and forcing `isSystemAdmin: true` in the frontend session state.

This does **not** appear to grant real backend admin privileges by itself, but it does expose owner-only settings UI to unauthorized users.

## Reproduction
1. Log in as a normal user.
2. Navigate directly to a server settings URL such as `/settings/<serverSlug>`.
3. Intercept or override the `GET /trpc/user.getCurrentUser` response in the browser and change `isSystemAdmin` to `true`.
4. Refresh the page.
5. The client-side redirect no longer triggers, and the server settings UI becomes accessible.

## Observed Impact
- Unauthorized users can load the **Overview**, **Privacy**, and **Danger Zone** settings UI.
- If the attacker is already a member of the server, they can also load the **Members** section because member-list reads are separately allowed by `server:read`.
- Owner/admin mutations still appear to be enforced server-side, so this looks like a **frontend authorization bypass / privileged UI exposure**, not a full backend privilege escalation.

## Likely Root Cause
- The settings route server-renders for any authenticated user via `getServerAuthenticated(serverSlug)` without enforcing owner/admin authorization first.
- `ServerSettingsPage` relies on a **client-side** redirect using `useAuth().isAdmin(server.ownerId)`.
- `AuthContext.isAdmin()` returns `true` whenever `user.isSystemAdmin` is truthy.
- `authService.getCurrentUser()` maps `user.getCurrentUser` directly into frontend state, including `isSystemAdmin`.
- `harmony-backend/src/lib/admin.utils.ts` still documents the system-admin bypass as dev-only, but the frontend trusts the returned flag for access control.

## Relevant Code Paths
- `harmony-frontend/src/context/AuthContext.tsx`
- `harmony-frontend/src/services/authService.ts`
- `harmony-frontend/src/components/settings/ServerSettingsPage.tsx`
- `harmony-frontend/src/app/settings/[serverSlug]/page.tsx`
- `harmony-backend/src/lib/admin.utils.ts`

## Suggested Fix
- Enforce settings-page authorization on the **server** before rendering the route.
- Do not trust a client-visible `isSystemAdmin` flag for gating privileged UI.
- Prefer a server-verified owner/admin permission check for the settings route and server actions.

acabrera04:
## Summary
- move server settings authorization into a shared server-side helper
- enforce that helper in the settings route and related server actions
- preserve access for server owners, system admins, and server admins
- add regression coverage for unauthorized and admin settings access

## Root Cause
The `/settings/[serverSlug]` flow server-rendered for any authenticated user and then relied on a client-side redirect that trusted the frontend `user.isSystemAdmin` state. A tampered `user.getCurrentUser` response could therefore suppress the redirect and expose privileged settings UI.

The concrete impact was privileged UI exposure and missing server-side enforcement for the settings page flow, not a backend mutation privilege escalation: the backend already enforced ownership and permission checks on the underlying mutations.

## Fix
This change introduces a server-side settings access guard that resolves the server and verifies the current user is allowed to manage settings before rendering the page or running the related settings actions. The guard now preserves the intended behavior for:
- server owners
- real system admins
- server admins

It also removes the unreachable null check in the settings page route and adds regression coverage for the admin path.

## Verification
- `npx jest --runInBand src/__tests__/server-settings-access.test.ts src/__tests__/authService.test.ts src/__tests__/VisibilityGuard.test.tsx`
- `npx eslint src/app/settings/[serverSlug]/settings-access.ts src/app/settings/[serverSlug]/page.tsx src/app/settings/[serverSlug]/actions.ts src/components/settings/MembersSection.tsx src/components/settings/ServerSettingsPage.tsx src/__tests__/server-settings-access.test.ts`
- `npx tsc --noEmit`

Closes #400.

### DIFF
--- a/harmony-frontend/src/__tests__/server-settings-access.test.ts
+++ b/harmony-frontend/src/__tests__/server-settings-access.test.ts
@@ -1,6 +1,6 @@
 import { requireServerSettingsAccess } from '@/app/settings/[serverSlug]/settings-access';
 import { getCurrentUser } from '@/services/authService';
-import { getServerAuthenticated } from '@/services/serverService';
+import { getServerAuthenticated, getServerMembersWithRole } from '@/services/serverService';
 
 const mockRedirect = jest.fn((path: string) => {
   throw new Error(`REDIRECT:${path}`);
@@ -20,11 +20,15 @@ jest.mock('@/services/authService', () => ({
 
 jest.mock('@/services/serverService', () => ({
   getServerAuthenticated: jest.fn(),
+  getServerMembersWithRole: jest.fn(),
 }));
 
 const mockGetCurrentUser = getCurrentUser as jest.MockedFunction<typeof getCurrentUser>;
 const mockGetServerAuthenticated = getServerAuthenticated as jest.MockedFunction<
   typeof getServerAuthenticated
+>;
+const mockGetServerMembersWithRole = getServerMembersWithRole as jest.MockedFunction<
+  typeof getServerMembersWithRole
 >;
 
 describe('requireServerSettingsAccess', () => {
@@ -37,6 +41,7 @@ describe('requireServerSettingsAccess', () => {
       ReturnType<typeof getServerAuthenticated>
     >;
     mockGetServerAuthenticated.mockResolvedValue(server);
+    mockGetServerMembersWithRole.mockResolvedValue([]);
     mockGetCurrentUser.mockResolvedValue({
       id: 'owner-1',
       username: 'owner',
@@ -55,6 +60,7 @@ describe('requireServerSettingsAccess', () => {
       ReturnType<typeof getServerAuthenticated>
     >;
     mockGetServerAuthenticated.mockResolvedValue(server);
+    mockGetServerMembersWithRole.mockResolvedValue([]);
     mockGetCurrentUser.mockResolvedValue({
       id: 'admin-1',
       username: 'admin',
@@ -67,11 +73,48 @@ describe('requireServerSettingsAccess', () => {
     await expect(requireServerSettingsAccess('test-server')).resolves.toBe(server);
   });
 
+  it('returns the server for a server admin', async () => {
+    const server = { id: 'server-1', ownerId: 'owner-1', slug: 'test-server' } as Awaited<
+      ReturnType<typeof getServerAuthenticated>
+    >;
+    mockGetServerAuthenticated.mockResolvedValue(server);
+    mockGetServerMembersWithRole.mockResolvedValue([
+      {
+        userId: 'admin-member-1',
+        username: 'admin-member',
+        displayName: 'Admin Member',
+        avatarUrl: null,
+        role: 'admin',
+        joinedAt: '2026-04-17T00:00:00.000Z',
+      },
+    ]);
+    mockGetCurrentUser.mockResolvedValue({
+      id: 'admin-member-1',
+      username: 'admin-member',
+      displayName: 'Admin Member',
+      role: 'member',
+      status: 'online',
+      isSystemAdmin: false,
+    });
+
+    await expect(requireServerSettingsAccess('test-server')).resolves.toBe(server);
+  });
+
   it('redirects unauthorized users to the channel view', async () => {
     const server = { id: 'server-1', ownerId: 'owner-1', slug: 'test-server' } as Awaited<
       ReturnType<typeof getServerAuthenticated>
     >;
     mockGetServerAuthenticated.mockResolvedValue(server);
+    mockGetServerMembersWithRole.mockResolvedValue([
+      {
+        userId: 'member-1',
+        username: 'member',
+        displayName: 'Member',
+        avatarUrl: null,
+        role: 'member',
+        joinedAt: '2026-04-17T00:00:00.000Z',
+      },
+    ]);
     mockGetCurrentUser.mockResolvedValue({
       id: 'member-1',
       username: 'member',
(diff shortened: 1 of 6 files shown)


declanblanc (commented):
## Review — Server settings authorization (#400)

### Overview
Introduces `requireServerSettingsAccess(serverSlug, mode)` — a shared server-side guard that resolves a server by slug, calls `getCurrentUser()`, and gates access to owner-or-system-admin only. Wires it into `/settings/[serverSlug]/page.tsx` and all five server actions (`saveServerSettings`, `deleteServerAction`, `getServerMembersAction`, `changeMemberRoleAction`, `removeMemberAction`). `MembersSection` now passes `serverSlug` instead of `serverId`, forcing server-side resolution. Adds 5 unit tests covering owner / system-admin / unauthorized / throw-mode / not-found paths.

### Severity / framing

**The "authorization bypass" framing slightly overstates the impact.** The backend already enforces ownership in `serverService.updateServer` / `deleteServer` (lines 132/159 of `server.service.ts`) and `withPermission('server:manage_members')` on the role/remove tRPC procedures. So a tampered client could *render* the settings UI but every mutation it issued would 403 at the backend. The real risk addressed here is **information disclosure** (privileged UI surfaces shown to non-admins) plus **defense in depth** at the server-action layer. Worth correcting in the PR description so the threat model is accurate.

### Strengths
- Centralizing the auth check in one helper that both the page and the actions call is the right shape — single source of truth, no chance of an action forgetting to check.
- `mode: 'redirect' | 'throw'` cleanly separates page-render UX (redirect) from action invocation (throw).
- Switching `MembersSection` to pass `serverSlug` is a real hardening step — the client can no longer pick an arbitrary `serverId`.
- Tests cover the four behavioral branches (owner, sysadmin, unauthorized-redirect, unauthorized-throw, not-found). Good coverage for a small surface.

### Issues

**Server `ADMIN` role loses settings UI access (intentional? worth confirming)**
Backend role permissions grant `server:update`, `settings:update`, `server:manage_members` to the `ADMIN` role (per `permission.service.ts`). The new guard only admits `ownerId === user.id || user.isSystemAdmin`. So a server `ADMIN` (non-owner, non-sysadmin) who is empowered to manage members at the API layer can no longer view the settings page that exposes that functionality. The previous code's client-side redirect was also `isSystemAdmin`-only, so this *matches* the prior FE behavior — not a regression — but the FE/BE access models are inconsistent. Worth a follow-up issue to align (either expand the guard to include server `ADMIN`s by querying `permissionService`, or document the OWNER+SYSADMIN-only intent for the settings page).

**Dead code in `page.tsx`**
```ts
const server = await requireServerSettingsAccess(serverSlug);
if (!server) notFound();
```
`requireServerSettingsAccess` calls `notFound()` itself when the server is missing and never returns null — the second check is unreachable. Drop it.

**No error handling around `getCurrentUser()`**
If the auth fetch fails (expired token, backend hiccup), `getCurrentUser` likely throws, propagating a 500 instead of redirecting to login. Minor, but worth a try/catch that redirects to `/auth/login` on auth-shaped errors.

**Mixed concern: large amount of Prettier reformatting noise**
`MembersSection.tsx` and `ServerSettingsPage.tsx` have lots of unrelated whitespace/wrapping changes mixed with the functional ones. Hard to tell at a glance what's a behavior change vs. a format change. Consider splitting reformatting into its own commit (or PR) in the future.

**Test mocking pattern**
Throwing `new Error('REDIRECT:...')` to simulate `next/navigation`'s `redirect` works, but it's fragile if `next/navigation` ever changes its contract. Consider a more idiomatic mock that asserts `mockRedirect` was called with the expected path rather than catching a thrown sentinel.

### Security
- The guard correctly handles `user === null` (would-be unauthenticated) via `Boolean(user && ...)`.
- No new privilege escalation surface introduced.
- The slug-resolved `server.id` flow prevents client-controlled `serverId` injection in member actions — strictly better than the prior code.

### Recommendation
Approve with two small follow-ups:
1. Remove the dead `if (!server) notFound()` in `page.tsx`.
2. Open a follow-up issue clarifying whether server `ADMIN`s should access the settings page (FE/BE access-model alignment).

Reformatting noise and PR description framing are non-blocking.

acabrera04:
> Re: comment 4133287976

Fixed in 4d2ef77. I removed the unreachable null check in the settings page route, updated the server-side guard so server ADMINs retain settings access, and added regression coverage for that admin path. I also tightened the PR description to frame the issue as privileged UI exposure / server-side settings-flow hardening rather than backend mutation privilege escalation.

declanblanc (approved):
Follow-ups applied cleanly:
- Dead `if (!server) notFound()` removed in `page.tsx` ✓
- Server `ADMIN` access added via `hasServerAdminAccess`, matching backend's permission model ✓
- New test case for server admin, additional fixture data for existing cases ✓

One non-blocking observation for future cleanup: `hasServerAdminAccess` fetches **all** members via `getServerMembersWithRole(serverId)` just to look up the current user's role. On servers with many members this is wasteful. A dedicated "get my role" / `getMyMembership` endpoint (or reusing the backend's `permissionService.getMemberRole`) would be O(1) instead of O(n). Worth a follow-up issue if perf becomes an issue; fine as-is for now.

Approving.

PR #402, opened 2026-04-18
fix: eliminate MaxListenersExceededWarning in eventBus (#396)

### LINKED ISSUE #396: Bug: backend-api emits MaxListenersExceededWarning after horizontal scaling
## Summary
The backend API logs a runtime warning after scaling to 2 replicas on Railway:

`MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 message listeners added to [Commander]. MaxListeners is 10.`

## Evidence
- Service: `harmony` (backend-api)
- Environment: `production`
- Deployment status: `SUCCESS`
- Warning observed in Railway logs around `2026-04-17T19:32:49Z`

## Impact
No immediate request failures were observed, but this indicates potential listener accumulation and may lead to memory/performance issues under sustained traffic.

## Repro Context
1. Scale backend-api to 2 replicas on Railway.
2. Hit API health endpoint repeatedly/concurrently.
3. Check `harmony` service logs for warning.

## Expected
No `MaxListenersExceededWarning` should be emitted during normal operation under horizontal scaling.

## Actual
Warning appears in runtime logs for backend-api.

AvanishKulkarni:
## Summary

- Each SSE connection called `client.on('message', messageListener)` on the shared ioredis subscriber client — adding one listener per active connection per subscribed channel
- With 11+ concurrent SSE connections the default Node.js MaxListeners limit (10) was exceeded, producing `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 message listeners added to [Commander]`
- Replace per-subscription `client.on()` calls with a single internal dispatcher: a `Map<channel, Set<handler>>` registry + one permanent `'message'` listener on the ioredis client that routes to registered handlers
- The emitter's listener count is now permanently **1** regardless of how many SSE connections or replicas are active

Closes #396

## Test plan

- [x] All 74 event-bus / SSE tests pass (`npm test -- --testPathPattern="eventBus|events|sse"`)
- [x] Full backend suite: 641/641 tests pass
- [x] `unsubscribe()` still removes only the target handler and issues Redis UNSUBSCRIBE only when the last handler for a channel is removed

### DIFF
--- a/harmony-backend/src/events/eventBus.ts
+++ b/harmony-backend/src/events/eventBus.ts
@@ -30,10 +30,12 @@ export type {
 let subscriberClient: Redis | null = null;
 const logger = createLogger({ component: 'event-bus' });
 
-// Per-channel handler count — tracks how many JS handlers are registered for
-// each Redis channel so we can unsubscribe at the Redis level precisely when
-// the last handler for a given channel is removed.
-const channelHandlerCounts = new Map<string, number>();
+// Central handler registry — maps each Redis channel to its active JS handlers.
+// A single 'message' listener on the ioredis client dispatches to these sets,
+// avoiding per-subscription client.on() calls that accumulate and trigger
+// MaxListenersExceededWarning under concurrent SSE connections.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const channelHandlers = new Map<string, Set<(payload: any) => void>>();
 
 // Per-channel ready promise — resolves when Redis confirms the subscription.
 // All subscribers on the same channel share this promise so latecomers don't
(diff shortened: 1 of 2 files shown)


declanblanc (approved):
## Review — eliminate MaxListenersExceededWarning in eventBus

### Overview
Replaces per-subscription `client.on('message', listener)` calls on the shared ioredis subscriber with a single module-level dispatcher + `Map<channel, Set<handler>>` registry. ioredis emitter listener count stays at 1 regardless of SSE connection volume, fixing the `MaxListenersExceededWarning` at 11+ concurrent connections.

### Correctness
- Dispatcher attaches once, inside the `subscriberClient === null` guard in `getSubscriberClient()`. `closeSubscriberClient` nulls the client and clears the registry, so re-init re-attaches correctly. ✓
- `channelHandlers.get(ch)` efficiently selects only handlers for the delivered channel — removes the per-listener `if (ch !== channel) return` branch as a side benefit. ✓
- `unsubscribe()` uses handler identity in a `Set` and only issues Redis `UNSUBSCRIBE` when the last handler is removed. Preserves prior semantics. ✓
- `isFirstSubscriber` captured before the `Set.add()`, avoiding a subtle off-by-one. ✓
- For-of iteration over a `Set` mutated mid-iteration (if a handler unsubscribes itself synchronously) is well-defined in JS — no crash risk.

### Minor observations (non-blocking)

**Subtle behavior change: duplicate-handler subscriptions are now deduped**
Previously, `channelHandlerCounts` incremented on every `subscribe()` call, so subscribing the *same function reference* twice tracked as 2 handlers. Now, `Set.add(handler)` dedupes by reference — subscribing the same fn twice registers once, and the second `unsubscribe()` would be a no-op on an already-removed entry. Unlikely to matter in practice (callers typically create fresh closures per subscription) but worth noting.

**`any` at the registry type boundary**
The three `// eslint-disable-next-line @typescript-eslint/no-explicit-any` annotations could be replaced with `Set<(p: unknown) => void>` — the callsite casts preserve type safety, and `unknown` is strictly more sound than `any`. Pure style, not blocking.

**No new test asserting the listener-count invariant**
A focused test like `subscribe(channel, h) × 20; expect(client.listenerCount('message')).toBe(1)` would pin down the fix and catch future regressions. Existing 641 tests confirm no regressions in observable behavior, but the specific invariant the PR targets isn't directly asserted. Would be a nice follow-up.

### Security / performance
- No new external input paths; dispatcher JSON.parses the same payloads as before.
- Performance improves slightly — one handler lookup + iteration per message instead of N linear filters.

### Recommendation
Approve. Clean, well-targeted fix with accurate root-cause analysis and minimal surface area.

PR #409, opened 2026-04-18
[codex] Make cloud integration frontend origin configurable

### LINKED ISSUE #330: Sprint 4: Railway Multi-Replica Validation and Deployment Evidence
## Summary
Scale `backend-api` to 2 or more replicas on Railway. Keep `backend-worker` at 1 replica.

Verify and capture evidence for:
- public page loads through the deployed frontend
- authenticated API behavior through multiple API replicas
- shared rate limiting across replicas
- attachment access independent of serving replica
- cache invalidation / indexing behavior via singleton worker
- SSE/realtime smoke verification in deployed environment

Run the cloud-target integration/smoke suite against the deployed system. Keep deployed-environment smoke verification read-only unless a separate isolated cloud test environment is explicitly provisioned. Use the replica observability added in #320 to prove requests were served across 2+ API replicas via headers, health output, and/or structured logs.

Capture logs/screenshots/test output needed for submission. Validate against the expected topology in `docs/deployment/deployment-architecture.md` and the replica-safety checklist in `docs/deployment/replica-readiness-audit.md`.

## Acceptance Criteria
- Live deployment is stable with `backend-api` at 2+ replicas
- No replica-specific failures are observed for required paths
- Proxy-aware rate limiting is validated against the documented one-hop trust model
- Cloud-target tests pass against deployed URLs
- Evidence clearly distinguishes deployed-system validation from localhost validation

## Ownership
- Assignee: `declanblanc`
- Backup owner: `Aiden-Barrera`
- Due: `April 18, 2026`

## Dependencies
- Blocked by: #318, #319, #320, #322, #324, #326, #329
- Unblocks: #331

acabrera04:
## Summary
- make the cloud integration auth and CORS checks use `FRONTEND_URL` instead of assuming `https://harmony.chat`
- keep the cloud suite runnable against the actual deployed frontend origin, whether that is the custom domain or the generated Vercel production URL

## Why
The cloud integration suite had hardcoded production-origin assumptions that did not match the live deployment host used during validation. This change keeps the tests environment-driven and avoids baking temporary hostname assumptions into the suite.

## Validation
- `TEST_TARGET=cloud BACKEND_URL=https://harmony-production-13e3.up.railway.app FRONTEND_URL=https://harmony-dun-omega.vercel.app CLOUD_TEST_SERVER_SLUG=testserver CLOUD_TEST_PUBLIC_CHANNEL=new-channel npm --prefix tests/integration run test:cloud`

### DIFF
--- a/tests/integration/auth.test.ts
+++ b/tests/integration/auth.test.ts
@@ -7,7 +7,14 @@
  */
 
 import crypto from 'crypto';
-import { BACKEND_URL, isCloud, LOCAL_SEEDS, localOnlyDescribe, localOnlyTest } from './env';
+import {
+  BACKEND_URL,
+  FRONTEND_URL,
+  isCloud,
+  LOCAL_SEEDS,
+  localOnlyDescribe,
+  localOnlyTest,
+} from './env';
 import { login } from './helpers/auth';
 
 function derivePasswordVerifier(password: string, saltHex: string): string {
@@ -34,7 +41,7 @@ describe('Auth Smoke', () => {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
-        Origin: isCloud ? 'https://harmony.chat' : 'http://localhost:3000',
+        Origin: FRONTEND_URL,
       },
       body: JSON.stringify({ email: 'smoke@example.invalid' }),
     });
@@ -77,7 +84,10 @@ localOnlyDescribe('Auth (local-only)', () => {
     const res = await fetch(`${BACKEND_URL}/api/auth/login`, {
       method: 'POST',
       headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify({ email: nonExistentEmail, passwordVerifier: verifier }),
+      body: JSON.stringify({
+        email: nonExistentEmail,
+        passwordVerifier: verifier,
+      }),
     });
     expect(res.status).toBe(401);
   });
@@ -99,7 +109,9 @@ localOnlyDescribe('Auth (local-only)', () => {
       headers: { Authorization: `Bearer ${accessToken}` },
     });
     expect(res.status).toBe(200);
-    const body = (await res.json()) as { result?: { data?: { email?: string } } };
+    const body = (await res.json()) as {
+      result?: { data?: { email?: string } };
+    };
     expect(body.result?.data?.email).toBe(email);
   });
 
@@ -111,7 +123,10 @@ localOnlyDescribe('Auth (local-only)', () => {
       body: JSON.stringify({ refreshToken: first.refreshToken }),
     });
     expect(res.status).toBe(200);
-    const body = (await res.json()) as { accessToken?: string; refreshToken?: string };
+    const body = (await res.json()) as {
+      accessToken?: string;
+      refreshToken?: string;
+    };
     expect(typeof body.accessToken).toBe('string');
     expect(typeof body.refreshToken).toBe('string');
 
--- a/tests/integration/cors.test.ts
+++ b/tests/integration/cors.test.ts
@@ -6,10 +6,7 @@
 import { BACKEND_URL, FRONTEND_URL, isCloud, LOCAL_SEEDS } from './env';
 import { login } from './helpers/auth';
 
-const PRODUCTION_FRONTEND_ORIGIN = 'https://harmony.chat';
-const localFrontendOrigin = 'http://localhost:3000';
-
-const allowedOrigin = isCloud ? PRODUCTION_FRONTEND_ORIGIN : localFrontendOrigin;
+const allowedOrigin = FRONTEND_URL;
 
 describe('CORS Header Verification', () => {
   test('CORS-1: OPTIONS preflight returns correct CORS headers for allowed origin', async () => {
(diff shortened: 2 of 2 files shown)


declanblanc (dismissed):
## PR Review Checklist

- [x] **1. Logic over Style** — The core logic change (replacing hardcoded `'https://harmony.chat'` with `FRONTEND_URL`) is correct, DRY, and consistent across both changed test files.
- [x] **2. Security First** — This PR only modifies integration test files; no production application code or secrets handling is affected. The change reduces implicit security assumptions in tests rather than introducing new ones.
- [x] **3. Architectural Alignment** — The change is test-infrastructure only and does not touch backend CORS configuration, which remains the source of truth.
- [ ] **4. Issue Completion** — Issue #330 requires cloud-target tests to pass against deployed URLs. The PR partially addresses this, but there is a consistency gap described below that would cause the CORS test to fail when `FRONTEND_URL` is overridden on the command line.
- [x] **5. No Nitpicking** — N/A (reviewer role).
- [x] **6. Avoid Repetition** — This is the first review on this PR.
- [x] **8. Prevent CI Failures** — One issue identified below that will cause `CORS-1` to fail silently or incorrectly in cloud mode with a non-default `FRONTEND_URL`.

---

## Review

### Blocking Issue: `cors.test.ts` does not use `FRONTEND_URL` for the local path in CORS-3

The change to `cors.test.ts` correctly replaces `PRODUCTION_FRONTEND_ORIGIN` with `FRONTEND_URL` for the `allowedOrigin` used in `CORS-1` and `CORS-2`. However, in `CORS-3`, the `Origin` header on the local-mode authenticated fetch still reads from `FRONTEND_URL` (already the case on `main`). This is fine — but the `allowedOrigin` variable now diverges from the local path:

```ts
// cors.test.ts (PR branch)
const localFrontendOrigin = 'http://localhost:3000';
const allowedOrigin = isCloud ? FRONTEND_URL : localFrontendOrigin;
```

When running in **local mode**, `allowedOrigin` is hardcoded to `'http://localhost:3000'` while `FRONTEND_URL` defaults to the same value — so local tests pass. But when `FRONTEND_URL` is overridden (e.g. to point at a Vercel preview URL) in local mode, `CORS-1` will send the overridden origin as the preflight `Origin` header via the `allowedOrigin = isCloud ? FRONTEND_URL : localFrontendOrigin` expression — wait, actually the inverse is the problem: in **cloud mode** specifically, `allowedOrigin` now equals `FRONTEND_URL`, which is correct. The real gap is that the `localFrontendOrigin` constant is now dead code that adds confusion — it is only ever used when `isCloud` is false, but it is a hardcoded string rather than falling back to `FRONTEND_URL`. This means that if someone runs the local suite against a non-standard local frontend port (by setting `FRONTEND_URL=http://localhost:3001`), `CORS-1` will construct its `Origin` header as `http://localhost:3000` and then assert the response `Access-Control-Allow-Origin` equals `http://localhost:3001` (`FRONTEND_URL`), causing a spurious failure.

The fix is straightforward — remove the special-casing entirely and always use `FRONTEND_URL`:

```ts
// suggested replacement
const allowedOrigin = FRONTEND_URL;
```

`FRONTEND_URL` already defaults to `'http://localhost:3000'` in local mode and `'https://harmony.chat'` in cloud mode (per `env.ts`), so removing the `isCloud` ternary is both simpler and fully consistent with how `auth.test.ts` was updated in this same PR.

---

### Secondary Issue: `ALLOWED_ORIGIN` export in `env.ts` on `main` is now stale / divergent

`main`'s `env.ts` exports `ALLOWED_ORIGIN` (line 28):

```ts
export const ALLOWED_ORIGIN = process.env.ALLOWED_ORIGIN ?? (IS_CLOUD ? 'https://harmony.chat' : 'http://localhost:3000');
```

The PR branch's `env.ts` does **not** export `ALLOWED_ORIGIN`. The `cors.test.ts` on `main` does not import it either (it derives `allowedOrigin` locally), so there is no import breakage — but the exported symbol in `main`'s `env.ts` and the env-var documentation in `tests/integration/README.md` (which lists `ALLOWED_ORIGIN` as a supported variable) become misleading once this PR lands. This is not a blocker, but the `ALLOWED_ORIGIN` row should be removed from the README table and the export removed from `env.ts` to keep the documentation truthful.

---

### Summary

The intent of this PR is correct and the cloud-mode path for `auth.test.ts` is properly fixed. The one change that must be made before merge is removing the `localFrontendOrigin` / `isCloud` ternary in `cors.test.ts` and replacing it with a direct reference to `FRONTEND_URL`, matching what was done in `auth.test.ts`. The stale `ALLOWED_ORIGIN` documentation is a secondary cleanup item.

acabrera04:
> Re: comment 4136471855

Fixed in 7206187. CORS-1 now uses FRONTEND_URL directly instead of special-casing local mode to http://localhost:3000, so local and cloud preflight checks both follow the same env-driven origin contract. Re-ran the cloud integration suite against the deployed Vercel/Railway hosts after the change.

FardeenI (approved):
Looks good — switching the integration suite to use FRONTEND_URL keeps cloud runs environment-driven (and matches env.ts defaults), and CI is green.

Non-blocking nits:
- In tests/integration/auth.test.ts and cors.test.ts, isCloud may now be unused; consider removing the import.
- If you want extra robustness when someone sets FRONTEND_URL with a path, you could derive the Origin via new URL(FRONTEND_URL).origin (and likewise for allowedOrigin).

PR #413, opened 2026-04-19
[codex] Fix SSE subscription readiness window

### LINKED ISSUE #412: SSE: eliminate first-connection subscription readiness window on fresh API replicas
## Summary
The deployed multi-replica SSE design has a known startup-window issue: on the first SSE connection to a freshly started `backend-api` replica, the connection can be accepted before Redis confirms the `SUBSCRIBE` handshake. During that small window, events published by another replica can be missed.

## Current Behavior
`eventBus.subscribe()` already returns `{ unsubscribe, ready }`, but the SSE router does not await `ready` before treating the stream as live.

Relevant references:
- `docs/deployment/replica-readiness-audit.md` §4.2 and §6.1
- `harmony-backend/src/events/eventBus.ts`
- `harmony-backend/src/routes/events.router.ts`

## Why This Matters
At demo scale this is low-probability, but it is still a real cross-replica correctness gap:
- User A can connect to replica 1
- User B can publish on replica 2
- If replica 1 has just opened its first SSE subscription and Redis has not confirmed it yet, User A can miss the event

## Desired Fix
Implement one of these approaches:
1. Await `eventBus.subscribe().ready` before the SSE route reports the connection as established / live.
2. Add replay support using `Last-Event-ID` so events missed during reconnect or startup windows can be recovered.

Option 1 is the smaller immediate fix. Option 2 is the more durable production-grade solution.

## Acceptance Criteria
- No event can be missed solely because the client connected to a freshly started replica before Redis subscription readiness completed.
- SSE behavior remains replica-safe with 2+ `backend-api` replicas.
- Regression coverage is added for the startup-window case or the replay contract.
- The deployment docs are updated if the chosen fix changes the runtime guarantees.

## Context
This was called out as a known limitation during issue #330 deployment validation and intentionally left unfixed there because that task was focused on read-only production validation rather than a code change.

acabrera04:
## Summary
- wait for all Redis SSE subscription handshakes before marking `/api/events/channel/:channelId` and `/api/events/server/:serverId` as live
- buffer events that arrive during the handshake window and flush them once the stream is ready
- add regression tests for delayed readiness and update the replica readiness audit to mark the gap as resolved

## Why
Fresh `backend-api` replicas could accept an SSE connection before Redis confirmed the underlying `SUBSCRIBE` calls. If another replica published during that startup window, the first client on the fresh replica could miss the event even though cross-replica fan-out was otherwise correct.

## Impact
This removes the first-connection readiness gap without changing the external SSE contract. Multi-replica deployments keep the same Redis Pub/Sub fan-out model, but a connection is not treated as live until all subscriptions are actually ready.

## Validation
- `npm test -- --runInBand --runTestsByPath tests/events.router.test.ts tests/events.router.server.test.ts`
- `npm run build`
- `npm run lint -- src/routes/events.router.ts tests/events.router.test.ts tests/events.router.server.test.ts`
  - existing unrelated warning remains in `harmony-backend/tests/events.router.sse-server-updated.test.ts`

Closes #412


declanblanc (changes requested):
## PR Review Checklist

- [x] **Prioritize Logic over Style** — The fix is logically sound. The buffered-state machine correctly sequences subscription creation, readiness awaiting, header flushing, and buffer draining.
- [x] **Security First** — No new security surface. Auth and authorization checks are unchanged and happen before any subscription is created. The 503 error on subscription failure does not leak internals.
- [x] **Architectural Alignment** — Aligns with the SSE fan-out model documented in `docs/unified-backend-architecture.md §4.6` and `docs/deployment/replica-readiness-audit.md §4.2`. No architectural drift.
- [x] **Issue Completion** — Closes #412. All four acceptance criteria are addressed: events cannot be missed during the startup window; replica-safety is preserved; regression coverage is added; deployment docs are updated.
- [ ] **No Nitpicking** — One concern flagged is a genuine DRY violation across test files, not a style preference.
- [ ] **Prevent CI Failures** — One gap in test coverage identified that could mask a regression in the server-scoped SSE route.

---

## Review

The core implementation is correct and well-structured. `finalizeSseSetup` cleanly encapsulates the readiness gate, and the `BufferedSseState` machine is idempotent against early client disconnect: `cleanupSseConnection` guards via `state.closed`, the cleanup listener is registered before `await Promise.all(...)` so early disconnects are never missed, and the subsequent `if (state.closed)` check prevents writing to a dead stream. The deployment doc updates are accurate and complete.

Two issues require changes before merge.

### 1. Missing buffering regression test for the server-scoped SSE route

The channel route (`/api/events/channel/:channelId`) has two readiness tests: one that verifies headers are withheld until `ready` resolves, and a second that verifies events arriving *during* the setup window are buffered and flushed after readiness. The server route (`/api/events/server/:serverId`) only has the header-withholding test — the buffering behavior for server-scoped events (`channel:created`, `member:joined`, etc.) is not covered.

This matters because `finalizeSseSetup` is shared, so buffering itself is implicitly tested for both routes. However, issue #412 acceptance criterion 3 explicitly requires regression coverage for the startup-window case. Without a buffering test for the server route, a future refactor that breaks buffering for server events only would not be caught by the test suite.

Add a buffering test to `tests/events.router.server.test.ts` parallel to the one in `tests/events.router.test.ts`.

### 2. `createDeferred` and `waitFor` are duplicated across both test files

`createDeferred<T>` and `waitFor` are copied verbatim into both `tests/events.router.test.ts` and `tests/events.router.server.test.ts`. This is a DRY violation: if the timeout, polling interval, or error message needs to change, both files must be updated in sync. Extract these into a shared test utility (e.g., `tests/helpers/async.ts`) and import them in both files.

acabrera04:
> Re: comment 4136472569

Fixed in ba621a3. Added a server-scoped buffering regression test for the startup window and extracted the shared async test helpers into tests/helpers/async.ts so the SSE route tests no longer duplicate createDeferred and waitFor. Re-ran the targeted Jest suite and backend build; scoped lint still shows the existing unrelated warning in tests/events.router.sse-server-updated.test.ts.

declanblanc (approved):
Reviewed latest commits and validated the SSE readiness fix. The buffering + delayed flush behavior closes the handshake window, and regression coverage for both channel and server routes is solid. Ran targeted tests locally: tests/events.router.test.ts and tests/events.router.server.test.ts (pass).

PR #414, opened 2026-04-19
ci: automate cloud integration smoke tests

acabrera04:
## Summary
- add a dedicated GitHub Actions job for cloud-target integration smoke tests
- require deployed frontend/backend URLs via repository variables so CI does not silently hit the wrong host
- keep the existing local integration workflow intact alongside the new deployed smoke coverage

## Verification
- ruby -e "require 'yaml'; YAML.load_file('.github/workflows/run-integration-tests.yml'); puts 'YAML OK'"\n\n## Notes\n- configure `CLOUD_TEST_BACKEND_URL` and `CLOUD_TEST_FRONTEND_URL` repository variables before expecting the new job to pass\n- optional authenticated smoke coverage can be enabled with `CLOUD_TEST_ACCESS_TOKEN` and the related optional cloud test variables

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -1,25 +1,32 @@
-# Integration test workflow — runs the local-target integration suite in CI.
+# Integration test workflow — runs both the local-target CI suite and the
+# deployed cloud-read-only smoke suite in CI.
 #
 # Job names are stable so this workflow can be used as a required status check
 # in branch protection rules (Settings → Branches → Require status checks →
-# "Run Integration Tests").
+# "Run Integration Tests" / "Run Cloud Integration Tests").
 #
 # ─── Cloud-target usage (outside CI) ────────────────────────────────────────
-# To run read-only smoke tests against the deployed URLs, export the required
-# env vars and run from the repo root:
+# To run read-only smoke tests against deployed URLs locally, export the
+# required env vars and run from the repo root:
 #
 #   export TEST_TARGET=cloud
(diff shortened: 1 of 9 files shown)


AvanishKulkarni (approved):
## Review

Overall this is a solid, focused PR. The core change is correct and the fail-fast config validation before running tests is a nice touch. The hardcoded `https://harmony.chat` origin fix in `auth.test.ts` and `cors.test.ts` is a meaningful correctness improvement.

### Suggestions (non-blocking)

**Medium — missing failure artifact upload in cloud job**
The local job uploads logs on failure (`retention-days: 3`). The new cloud job has no equivalent. If the cloud smoke run fails, there is nothing to diagnose. Consider adding:
\`\`\`yaml
- name: Upload cloud test logs
  if: failure()
  uses: actions/upload-artifact@v4
  with:
    name: cloud-integration-logs
    path: tests/integration/
    retention-days: 3
\`\`\`

**Low — `test:cloud` missing `--runInBand`**
`test` and `test:local` both pass `--runInBand`; `test:cloud` does not. Low risk now but worth aligning for consistency.

**Low — validation comment**
`env.ts` has fallback defaults for `BACKEND_URL`/`FRONTEND_URL` in cloud mode, so the shell validation is the only real guard. A short comment in the workflow explaining why this check exists (i.e., `env.ts` would silently fall back to production defaults) would help future readers.

**Info — GitHub Environment must exist**
The `environment: cloud-integration-tests` block references a GitHub Environment that needs to be created in repo settings before this job can run. Worth calling out in the PR description or a follow-up issue.

Approvable as-is; the artifact upload gap is worth a follow-up issue if not addressed here.

PR #417, opened 2026-04-20
test: stabilize cloud integration fixtures

acabrera04:
## Summary
- discover a live public server/channel fixture from the deployed public API when cloud fixture env vars are unset
- update the cloud-read-only integration tests to use the discovered fixture instead of stale hard-coded slugs
- add a local Prettier config for tests/integration so formatting there matches the repo single-quote style

## Root cause
The cloud integration workflow was falling back to `harmony-hq/general` when `CLOUD_TEST_SERVER_SLUG` and `CLOUD_TEST_PUBLIC_CHANNEL` were unset. The live Railway data no longer serves that fixture, so the cloud smoke suite hit 404s even though public content still existed.

## Verification
- `cd tests/integration && TEST_TARGET=cloud BACKEND_URL=https://harmony-production-13e3.up.railway.app FRONTEND_URL=https://harmony-dun-omega.vercel.app npm test`

### DIFF
--- a/docs/test-specs/integration-test-spec.md
+++ b/docs/test-specs/integration-test-spec.md
@@ -164,9 +164,12 @@ Until an isolated environment is provisioned, all write-path test cases in this 
 
 **Backend routes:** `harmony-backend/src/routes/attachment.router.ts`
 **Storage backend:** `harmony-backend/src/lib/storage/`
-**Classification:** local-only (upload is a write path; production storage is `STORAGE_PROVIDER=local` until issue #319 implements S3)
+**Classification:** mixed
 
-> **Production storage note:** The current backend only supports `STORAGE_PROVIDER=local`, which writes files to the instance's local disk. This is not safe for 2+ replicas (see `docs/deployment/replica-readiness-audit.md §3.4`). These tests are local-only until issue #319 implements `S3StorageProvider`. Once S3 is in place, upload and retrieval tests may be promoted to `cloud-isolated-env-only`.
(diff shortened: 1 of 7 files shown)


Aiden-Barrera (approved):
## PR Review — ✅ Approved

Swaps the hard-coded `CLOUD_KNOWN` pair for runtime discovery via `/api/public/servers`, and promotes the pre-storage validation paths of ATT to cloud-safe. Sound test-infra improvement — no security or correctness concerns in the production code path (changes are test-only).

### Issues
✅ None blocking.

### Non-blocking
- MEDIUM: [tests/integration/env.ts] — `cloudFixturePromise` caches the promise permanently; if the first resolve rejects (transient network blip) every subsequent call reuses the same rejected promise for the process lifetime. Consider clearing the cache on rejection so a retry can succeed.
- MEDIUM: [tests/integration/attachments.test.ts:54,68,82] — `if (!accessToken) return;` silently passes the test when `CLOUD_TEST_ACCESS_TOKEN` is missing. `test.skip`/`pending` would make the skip visible in CI output instead of masquerading as a pass.
- LOW: [tests/integration/env.ts] — `channelsBody.channels?.find((channel) => channel.slug)?.slug` just takes the first slug-bearing channel; if the public-channels endpoint ever returns PUBLIC_NO_INDEX entries, tests asserting indexable behavior could pick the wrong one. Fine for now since the endpoint filters to public, but worth a comment.

---
*Reviewed by claude-opus (diff-only)*

PR #442, opened 2026-04-20
Rate-limit joinServer to prevent mass-join attacks

### LINKED ISSUE #428: [HIGH] Rate-limit joinServer to prevent mass-join attacks
## Description
`serverMember.joinServer` is exposed via an `authedProcedure` with no per-user or per-IP rate limiting or CAPTCHA. Any authenticated user can mass-join every public server in rapid succession.

**Relevant file:** `harmony-backend/src/services/serverMember.service.ts` lines 49–80

## Exploitability
An attacker can inflate member counts on public servers, exhaust member-sidebar rendering resources for legitimate users, and scrape membership metadata.

## Fix
Add per-user/per-IP rate-limit middleware to membership mutation endpoints, or enforce a short cooldown per `(userId, serverId)` pair in the service layer.

declanblanc:
Closes #428

## Summary
- Adds a per-user Redis rate limit to `joinServer` (10 joins per 60-second window)
- Uses an atomic Lua `INCR`+`EXPIRE` script so the counter is replica-safe with no non-atomic race conditions
- The check runs before the server lookup, so an over-limit call throws `TOO_MANY_REQUESTS` immediately
- Limit is configurable via `JOIN_RATE_LIMIT` env var for easy tuning

## Test plan
- [x] Existing 22 `serverMember` integration tests still pass
- [x] New test: at the limit (`count == JOIN_RATE_MAX`) the call is allowed and proceeds to `NOT_FOUND` (rate check passed)
- [x] New test: over the limit (`count > JOIN_RATE_MAX`) throws `TOO_MANY_REQUESTS` before any DB access
- [x] New test: after window reset (key deleted) the counter restarts and the call is allowed


AvanishKulkarni (changes requested):
## Review Checklist

- [x] **1. Logic over Style** — Rate-limiting logic is sound: Lua INCR+EXPIRE is atomic, boundary conditions (at-limit vs. over-limit) are correctly handled.
- [x] **2. Security First** — Closes a real HIGH-severity attack vector. No secrets exposed. Per-user key scoping is appropriate.
- [x] **3. Architectural Alignment** — Service-layer enforcement is a reasonable departure from HTTP middleware since `joinServer` is a tRPC procedure. Consistent with the existing `rl:` key prefix convention.
- [x] **4. Issue Completion** — Issue #428 is satisfied (per-user limit, short cooldown).
- [x] **5. No Nitpicking** — Only critical items below.
- [x] **7. Iterative Reviews** — N/A, first review.
- [ ] **8. Prevent CI Failures** — One silent failure mode identified (NaN env var parsing).

---

## Review

This PR adds a Redis-backed per-user rate limit (10 joins/60 s) to `serverMemberService.joinServer` via an atomic Lua `INCR`+`EXPIRE` script. It's a well-scoped, targeted fix for the mass-join attack vector in issue #428. The approach is sound and the tests cover the three meaningful boundary conditions.

---

### 🔴 Issue 1 — Silent rate-limit disable on invalid `JOIN_RATE_LIMIT` env var (Blocking)

**`harmony-backend/src/services/serverMember.service.ts`, line 9**

```ts
export const JOIN_RATE_MAX = parseInt(process.env.JOIN_RATE_LIMIT ?? '10', 10);
```

If `JOIN_RATE_LIMIT` is set to a non-numeric string (e.g. a misconfigured deploy), `parseInt` returns `NaN`. The comparison `count > NaN` is always `false`, which **silently disables the rate limit entirely** — the exact vulnerability this PR fixes would be re-openable via a bad env var.

**Suggested fix:**

```ts
const parsed = parseInt(process.env.JOIN_RATE_LIMIT ?? '', 10);
export const JOIN_RATE_MAX = Number.isFinite(parsed) && parsed > 0 ? parsed : 10;
```

---

### 🟡 Issue 2 — `redis.eval` always runs, even in test/dev environments (Minor)

The existing `rate-limit.middleware.ts` explicitly opts out of Redis in non-production:

```ts
function buildProductionStore(): Store | undefined {
  if (process.env.NODE_ENV !== 'production') return undefined;
  ...
}
```

`enforceJoinRateLimit` unconditionally calls `redis.eval`, making all `joinServer` calls depend on a live Redis connection. The current tests are integration tests with Redis available so CI is unaffected, but it diverges from the project's pattern of keeping non-integration tests hermetic. Worth tracking for future unit tests of `joinServer`.

---

### Summary

The implementation is correct and well-tested. **Issue 1 is the only blocking concern** — the NaN guard is a one-liner fix that closes a silent failure mode which could re-open the exact vulnerability this PR addresses.

declanblanc:
**Issue 1 (blocking) — fixed in 0cf138f.**

Replaced the bare `parseInt(... ?? '10')` with:
```ts
const _parsedLimit = parseInt(process.env.JOIN_RATE_LIMIT ?? '', 10);
export const JOIN_RATE_MAX = Number.isFinite(_parsedLimit) && _parsedLimit > 0 ? _parsedLimit : 10;
```
Any non-numeric or zero/negative value now falls back to 10, so a misconfigured env var can never silently disable the rate limit.

**Issue 2 (minor) — acknowledged, not changed.**
The unconditional `redis.eval` is intentional: `joinServer` is an integration-tested mutation that already depends on Prisma and the event bus. Adding a production guard here would make the security control untestable in CI. Tracked as a future unit-test ergonomics improvement.

AvanishKulkarni (approved):
The blocking concern (NaN guard on `JOIN_RATE_MAX`) is addressed — the updated parsing with `Number.isFinite` and a positive-value check correctly falls back to 10 on any invalid input. Issue 2 (unconditional `redis.eval`) remains a minor non-blocking note; fine to carry as-is. LGTM.

PR #444, opened 2026-04-20
test(integration): explicit skip for auth-dependent attachment cases

### LINKED ISSUE #420: test(integration): mark cloud attachment auth-dependent cases as skipped when token is absent
Follow-up from PR #417 approval review.

Problem:
`tests/integration/attachments.test.ts` currently uses `if (!accessToken) return;` for cloud attachment checks that require `CLOUD_TEST_ACCESS_TOKEN`. That makes the test appear to pass instead of surfacing a visible skip in CI output.

Requested change:
- convert those early returns into explicit skips or a visibly pending outcome
- keep the unauthenticated attachment smoke case running without a token

Context:
Raised as a non-blocking review comment on PR #417.

AvanishKulkarni:
## Summary

- Adds `cloudTokenTest` helper to `tests/integration/env.ts` (mirrors existing `localOnlyTest` pattern): calls `test.skip` when `isCloud && !CLOUD_TEST_ACCESS_TOKEN`, otherwise calls `test` normally
- Replaces `if (!accessToken) return;` silent early-returns in ATT-3, ATT-4, ATT-6 with `cloudTokenTest(...)`, making the skip visible in CI output
- ATT-2 (unauthenticated 401 smoke) is untouched and always runs

Closes #420

## Test plan

- [ ] Local run (`TEST_TARGET=local`): ATT-3, ATT-4, ATT-6 run normally (token comes from `login()`)
- [ ] Cloud run without token (`TEST_TARGET=cloud`, no `CLOUD_TEST_ACCESS_TOKEN`): ATT-3, ATT-4, ATT-6 appear as **skipped** in Jest output, ATT-2 still passes
- [ ] Cloud run with token: all four tests run and pass
- [ ] TypeScript: `tsc --noEmit` passes cleanly

### DIFF
--- a/llm-logs/2026-04-20-100616-session-12bb34ec.md
+++ b/llm-logs/2026-04-20-100616-session-12bb34ec.md
@@ -1,0 +1,468 @@
+# Session Log 2026-04-20-100616
+
+## User
+
+complete https://github.com/CS485-Harmony/Harmony/issues/420 in a new branch
+
+## Assistant
+
+**Tool call: `ToolSearch`**
+```json
+{
+  "query": "select:WebFetch",
+  "max_results": 1
+}
+```
+
+## Assistant
+
+**Tool call: `Bash`**
+```json
+{
+  "command": "gh issue view 420 --repo CS485-Harmony/Harmony"
+}
+```
+
+## Assistant
+
+Let me read the relevant test file.
+
+## Assistant
+
+**Tool call: `Read`**
+```json
+{
+  "file_path": "/Users/bigfa/Documents/GitHub/Harmony/tests/integration/attachments.test.ts"
+}
+```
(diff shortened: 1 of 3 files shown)


declanblanc (changes requested):
Requesting changes for one blocker: this PR includes an auto-generated transcript file under llm-logs/ that should not be committed to source control. Please remove the log file from the PR and keep only the integration test changes in tests/integration/*. The actual test wrapper changes look good.

declanblanc (approved):
Per maintainer guidance, llm-logs in this repo are intentional and not a blocker. Re-reviewed the functional change: explicit cloud token gating via cloudTokenTest and replacing silent early returns is correct and improves CI visibility.

PR #447, opened 2026-04-20
[codex] Disable automatic frontend E2E workflow runs

acabrera04:
## Summary
- stop running the frontend E2E workflow on `push`
- stop running the frontend E2E workflow on `pull_request`
- keep the workflow available through manual `workflow_dispatch`

## Why
The frontend E2E workflow was adding noisy failing checks to PRs even though it is not a required status check. This keeps the job available for manual runs without having it execute on every branch update and PR.

## Validation
- inspected the workflow diff to confirm only the trigger block changed
- did not run tests because this change only adjusts GitHub Actions triggers

### DIFF
--- a/.github/workflows/run-frontend-e2e.yml
+++ b/.github/workflows/run-frontend-e2e.yml
@@ -1,10 +1,6 @@
 name: Frontend E2E
 
 on:
-  push:
-    branches: [main, master]
-  pull_request:
-    branches: [main, master]
   workflow_dispatch:
 
 jobs:


Aiden-Barrera (approved):
## PR Review — ✅ Approved

Removes automatic `push`/`pull_request` triggers from the Frontend E2E workflow, keeping `workflow_dispatch` so it can still be run manually. Safe, scoped change.

### Issues
✅ None

### Non-blocking
- Confirm `Frontend E2E` is not configured as a required status check on `main` — if it is, PRs will stall waiting for a check that no longer auto-runs. Per `docs/deployment/team-workflow.md` the required checks are `Backend Lint and Build`, `Frontend Lint and Build`, `Run Backend Tests`, `Run Frontend Tests`, and `Run Integration Tests`, so this should be fine, but worth verifying in branch protection settings.

---
*Reviewed by claude (diff-only, manual review — diff was 4 lines, not worth spawning a sub-agent)*

PR #448, opened 2026-04-20
Fix SSE subscription readiness race dropping cross-user messages

declanblanc:
## Summary
- wait for Redis Pub/Sub subscription readiness before considering SSE streams live on both `/api/events/channel/:channelId` and `/api/events/server/:serverId`
- fail fast with cleanup and a deterministic `500 { error: \"Failed to initialize event stream\" }` when subscription setup fails, instead of silently opening a stream that can miss early events
- add regression coverage for readiness failures on both SSE endpoints to prevent reintroducing startup race behavior

## Root cause
`eventBus.subscribe()` returns both `unsubscribe` and `ready`, but the SSE routes previously opened streams without awaiting `ready`. On initial connect/reconnect, events published during the Redis subscribe handshake window could be missed. Users would then only see those messages after a page refresh reloaded persisted data.

## Implementation details
- introduced `awaitSubscriptionReadiness(...)` helper in `events.router.ts`
- channel endpoint now:
  - registers subscriptions
  - awaits all relevant `.ready` promises (`MESSAGE_CREATED`, `MESSAGE_EDITED`, `MESSAGE_DELETED`, `SERVER_UPDATED`)
  - only then sends SSE headers and starts heartbeats
- server endpoint now:
  - awaits early channel membership-set subscriptions (`CHANNEL_CREATED`, `CHANNEL_DELETED`) before preload
  - awaits the remaining stream subscriptions before continuing heartbeat/live streaming
  - reuses existing cleanup flow on readiness failure

## Validation
- targeted backend tests pass:
  - `npm test -- events.router.test.ts events.router.server.test.ts --runInBand`

## Test plan
- [x] run backend SSE route tests for channel and server endpoints
- [x] verify readiness-failure behavior returns HTTP 500 instead of hanging/opening stream
- [ ] validate on deployed environment with two authenticated users in same channel and confirm cross-user messages render without refresh

Made with [Cursor](https://cursor.com)


harmony-backend/src/routes/events.router.ts
```diff
@@ -608,8 +659,28 @@ eventsRouter.get('/server/:serverId', async (req: Request, res: Response) => {
...
+      channelUpdatedSub.ready,
+      statusChangedSub.ready,
+      memberJoinedSub.ready,
+      memberLeftSub.ready,
+      visibilityChangedSub.ready,
+    ],
+    cleanup,
+  );
+  if (!serverReady) {
+    if (!res.headersSent) res.status(500).json({ error: 'Failed to initialize event stream' });
```
AvanishKulkarni: This guard is unreachable. SSE headers are flushed at line 422 (`res.flushHeaders()`), before the second batch of subscriptions is even started. By the time `serverReady` can be `false`, `res.headersSent` is already `true`, so this branch can never send the 500.

The channel endpoint avoids this by deferring `res.flushHeaders()` until after all subscriptions are confirmed ready. The server endpoint should do the same: move the SSE header flush block (currently at ~line 422) to after the `serverReady` guard. That also makes the two-phase `preloadedChannelsReady` check redundant for the purpose of 500 reporting — you could await all subscriptions at once before flushing, then run the `findMany` preload after (since the channel-ID handlers only need the Redis subscription to be live, not headers to be sent).

If the current two-phase split is intentional for timing reasons, at minimum replace the unreachable 500 with an explicit SSE error event (`sendEvent(res, "error", {...})`) followed by `res.end()`, so the client gets a signal rather than a silently dropped stream.

harmony-backend/tests/events.router.server.test.ts
```diff
@@ -230,3 +230,24 @@ describe('GET /api/events/server/:serverId — authorisation', () => {
...
+  it('returns 500 when SSE subscriptions fail to become ready', async () => {
+    const failingReady = Promise.reject(new Error('redis subscribe failed'));
+    // Mark as handled immediately so Jest doesn't flag an unhandled rejection
+    // before the route awaits the readiness promise.
+    failingReady.catch(() => undefined);
+    mockSubscribe.mockReturnValueOnce({
+      unsubscribe: jest.fn(),
+      ready: failingReady,
+    });
+    mockSubscribe.mockReturnValue({ unsubscribe: jest.fn(), ready: Promise.resolve() });
```
AvanishKulkarni, line 588: This test covers only the `preloadedChannelsReady` failure path (first `subscribe` call → `CHANNEL_CREATED`), which fires before SSE headers are flushed and can legitimately return a 500. There is no test for the `serverReady` failure path (second batch of subscriptions: `MESSAGE_CREATED`, `MESSAGE_EDITED`, …), where headers are already sent and the 500 response cannot be delivered. Add a second case that injects a failing `ready` promise into one of those later subscriptions and asserts that the connection closes gracefully (e.g. `res.end()` called, no heartbeat leak).

AvanishKulkarni (changes requested):
## PR Review Checklist

- [x] **Logic over Style** — Core race-condition fix is sound; one unreachable code path flagged below.
- [x] **Security First** — No new surface area or vulnerabilities introduced.
- [x] **Architectural Alignment** — Follows existing `eventBus.subscribe` / cleanup patterns in the file.
- [x] **Issue Completion** — No linked issue; PR description goals are mostly met (see inline note on server endpoint 500 path).
- [x] **No Nitpicking** — Only structural/correctness comments included.
- [x] **Iterative Reviews** — First review.
- [x] **Prevent CI Failures** — Tests should pass for the covered cases; gap noted below.

---

The readiness fix is the right move and the channel endpoint is correctly restructured. Two issues on the server endpoint worth addressing before merge.

acabrera04 (changes requested):
## PR Review Checklist

- [x] **Prioritize Logic over Style** - Reviewed the behavioral SSE readiness changes and focused on correctness rather than style.
- [x] **Security First** - No new auth bypass, secret exposure, or input-trust issue found in the changed surface.
- [x] **Architectural Alignment** - The change follows the existing `eventBus.subscribe` and cleanup structure, but the server endpoint readiness sequencing still has a correctness issue.
- [ ] **Issue Completion** - The PR goal is not fully satisfied because the server-scoped stream can still fail readiness after SSE headers have already been flushed, so the deterministic 500 path is not actually deliverable for that batch.
- [x] **No Nitpicking** - Feedback is limited to a blocking runtime behavior and the matching regression-test gap.
- [x] **Avoid Repetition** - Existing unresolved inline threads already capture the exact line-level issue, so I am not adding duplicate inline comments.
- [x] **Iterative Reviews** - This latest head was rechecked; the only new commit adds LLM logs and does not address the prior requested changes.
- [x] **Prevent CI Failures** - GitHub reports the Vercel checks passing; I also checked for obvious backend test-coverage gaps in the PR diff.

---

Requesting changes because the unresolved prior review findings still apply to the latest head (`7b3808e`). In `harmony-backend/src/routes/events.router.ts`, the server endpoint flushes SSE headers at line 426 before awaiting the second batch of subscription readiness promises. If one of those later `ready` promises rejects, the guard at line 680 cannot send `500 { error: "Failed to initialize event stream" }` because `res.headersSent` is already true. That leaves the server-scoped endpoint inconsistent with the PR contract and with the channel endpoint, which waits to flush headers until after readiness succeeds.

The current server readiness test only covers the first subscription batch, before headers are sent. It still needs coverage for a later subscription failure, or the implementation needs to move the header flush until after all readiness promises that are meant to fail fast have resolved. The existing unresolved threads on `events.router.ts` line 680 and `events.router.server.test.ts` line 244 describe the concrete fixes, so I am relying on those rather than reposting duplicate inline feedback.

CI evidence: `gh pr checks 448 --repo CS485-Harmony/Harmony` reports `Vercel` and `Vercel Preview Comments` passing, but no backend Jest check is shown in the current rollup.

Aiden-Barrera (changes requested):
The channel-scoped readiness fix is correct, but the server-scoped stream still cannot deliver the advertised fail-fast behavior after headers are flushed.

### Issues
- `harmony-backend/src/routes/events.router.ts`: the server endpoint flushes SSE headers before awaiting the second readiness batch, so a later `ready` rejection cannot return the documented `500 { error: "Failed to initialize event stream" }`. That leaves the server path inconsistent with the channel path and with the PR description.
- `harmony-backend/tests/events.router.server.test.ts`: coverage only exercises the early readiness failure path before headers are sent. There is still no regression test for a failure in the later subscription batch after headers are already committed.

### Non-blocking
- `llm-logs/...` is unrelated to the bugfix and should stay out of follow-up functional discussion.

Aiden-Barrera (dismissed):

Moves SSE header flush after Redis subscription readiness, eliminating the race that dropped early messages. The core fix is correct and the tests confirm the failure path, but two issues need addressing.

### Issues
- **HIGH:** `harmony-backend/src/routes/events.router.ts` — `awaitSubscriptionReadiness` logs `err` but not which subscription failed. If one of N Redis channels errors, operators get no signal about which one. Add the failing subscription channel name to the log.
- **HIGH:** `harmony-backend/src/routes/events.router.ts` — After `res.status(500).json(...)`, there is no explicit `res.end()`. The response closes implicitly via Express, but this is fragile. Call `res.end()` or chain it after `.json()` to be explicit.

### Non-blocking
- Tests don't assert that `unsubscribe` callbacks are called on readiness failure. Worth adding `expect(unsubFn).toHaveBeenCalled()` to prove cleanup is wired end-to-end.

---
*Reviewed by claude-haiku*

declanblanc:
> Re: comment 4177624355

Fixed in 8ae6ec95. (HIGH 1) awaitSubscriptionReadiness now accepts named entries and logs failedSubscriptions with the Redis channel names of whichever subscriptions rejected. (HIGH 2) res.json() internally calls res.end() in Express — adding an explicit res.end() after would throw. The implicit close is correct and not fragile. (Non-blocking) The existing first-batch test now asserts unsubscribe was called; the new second-batch test does the same.

declanblanc:
> Re: comment 4156003879

Fixed in 8ae6ec95. The server endpoint no longer flushes SSE headers before awaiting the second subscription batch. res.flushHeaders() is now called only after serverReady resolves successfully, so a failure in the second batch returns a proper 500 JSON response. A new test (second-batch failure returns 500 before headers are flushed) confirms this path end-to-end.

declanblanc:
> Re: comment 4155953837

Fixed in 8ae6ec95. res.flushHeaders() is now deferred until after all subscriptions in the second batch confirm readiness. The 500 guard at serverReady is now reachable and returns proper JSON. The new test verifies this: second-batch failure → 500 response, no SSE headers sent.

harmony-backend/src/routes/events.router.ts
declanblanc: Fixed in 8ae6ec95. res.flushHeaders() is now called after the serverReady guard, making the 500 branch reachable. The comment about the guard being unreachable no longer applies.

harmony-backend/tests/events.router.server.test.ts
declanblanc: Fixed in 8ae6ec95. Added a second test case that injects a failing ready promise into the third subscribe call (MESSAGE_CREATED, first in the second batch). The test asserts res.status === 500, correct JSON body, no SSE headers sent, and all registered unsubscribe callbacks called.

declanblanc:
> Re: comment 4150374751

Fixed in 8ae6ec95. Both server endpoint issues are addressed: the unreachable 500 branch is now reachable (header flush deferred past serverReady), and awaitSubscriptionReadiness logs failedSubscriptions with Redis channel names for diagnosability.

acabrera04 (approved):
## PR Review Checklist

- [x] **Logic over Style** - Latest head moves `res.flushHeaders()` until second-batch readiness succeeds, so fail-fast behavior now matches the intended runtime contract.
- [x] **Security First** - No new auth or data-exposure paths were introduced; changes are limited to SSE startup sequencing and diagnostics.
- [x] **Architectural Alignment** - Keeps the existing `eventBus.subscribe` plus centralized cleanup pattern and improves readiness logging with failed channel names.
- [x] **Issue Completion** - The previously blocking server-stream race is addressed at head `8ae6ec95b098ab0fa693b024b7ae5c9d41906d6d`, including regression coverage for second-batch readiness failure.
- [x] **No Nitpicking** - Review focused on correctness, cleanup behavior, and observable failure handling.
- [x] **Avoid Repetition** - I did not restate prior inline findings; those threads were already addressed in the latest commit.
- [x] **Iterative Reviews** - Re-reviewed against the newest commit on April 29, 2026 and validated that prior requested changes were incorporated.
- [x] **Prevent CI Failures** - Current PR checks are green (Vercel plus Vercel Preview Comments), and the new tests cover both early and second-batch readiness failure paths.

Approving. The latest update resolves the previously reported server-endpoint readiness and header-order issue and adds matching regression assertions for cleanup and deterministic 500 behavior before SSE headers are sent.

declanblanc:
> Re: comment 4177624355

Both issues addressed in 8ae6ec95: (1) awaitSubscriptionReadiness now collects failed channel names via Promise.allSettled and logs them as failedSubscriptions — operators get the exact channel names that failed; (2) res.flushHeaders() is now moved after all subscriptions succeed, so a failure in any batch sends res.status(500).json() before headers are committed and the response ends via Express's json() implementation. Regression tests for second-batch readiness failure were also added.

declanblanc:
> Re: comment 4156003879

Both issues addressed in 8ae6ec95: (1) res.flushHeaders() is now moved to after awaitSubscriptionReadiness succeeds for the server endpoint's second batch — if any subscription fails, res.status(500).json() fires before headers are committed, consistent with the channel endpoint and the PR description; (2) regression test for a second-batch readiness failure was added to events.router.server.test.ts.

declanblanc:
> Re: comment 4155953837

Addressed in 8ae6ec95 — the server endpoint's header flush was moved to after all subscription readiness checks pass, making the fail-fast 500 path deterministic. Second-batch regression test added. See acabrera04's approval at that head.

declanblanc:
> Re: comment 4150374751

Addressed in 8ae6ec95 — server endpoint header flush sequencing fixed and second-batch test coverage added.

acabrera04 (approved):
## PR Review Checklist

- [x] **Prioritize Logic over Style** - Re-reviewed the latest head for behavioral correctness of SSE readiness failure handling and corresponding regression tests.
- [x] **Security First** - No new auth, secret exposure, or unsafe input-handling risks were introduced in the updated changes.
- [x] **Architectural Alignment** - Test updates remain aligned with the existing backend event-stream architecture and established route behavior.
- [x] **Issue Completion** - The latest head includes the readiness regression coverage updates and reflects the intended failure contract for stream initialization.
- [x] **No Nitpicking** - Feedback is limited to merge-critical correctness and verification outcomes.
- [x] **Avoid Repetition** - I am not repeating prior inline comments; previous blockers were already addressed and dismissed.
- [x] **Iterative Reviews** - This is an iterative re-review on latest head `45af726855cb4de6829ac7ec0ac3f84609ea139b`.
- [x] **Prevent CI Failures** - Current checks are green, including backend/frontend lint-build and backend/frontend/integration tests.

Approving on latest head. No new blocking issues found in this revision.

PR #457, opened 2026-04-23
fix(scripts): rename build:e2e/start:e2e to local-only names (#423)

### LINKED ISSUE #423: Clarify or harden local-only frontend e2e startup scripts
## Summary
The new `build:e2e` and `start:e2e` frontend scripts inline `NEXT_PUBLIC_API_URL=http://localhost:4000` and `NEXT_PUBLIC_BASE_URL=http://localhost:3000`. That is correct for the documented local integration path, but the scripts are intentionally local-only and could be misused in other contexts.

## Context
Follow-up from PR #418 approval:
- `harmony-frontend/package.json` now includes `build:e2e` and `start:e2e`
- They are meant to mirror the local GitHub Actions-style integration setup from the README
- If reused in a deploy pipeline or other automation without context, they would bake localhost URLs into the bundle

## Goal
Decide whether the current scripts should remain as-is, gain clearer naming/docs/guards, or be refactored so the localhost env values are set outside the package scripts.

## Options to evaluate
- Keep the scripts and document explicitly that they are local-only
- Rename them to make local-only scope even more obvious
- Replace inline env values with a wrapper script or task runner that sets them externally

## Acceptance criteria
- The local integration README path stays simple to run
- The chosen approach reduces the risk of accidental reuse in non-local environments
- Any updated scripts or docs still support the passing local integration workflow from PR #418

## Source
Non-blocking review comment on PR #418

AvanishKulkarni:
## Summary
- Renames `build:e2e` → `build:local-e2e` and `start:e2e` → `start:local-e2e` in `harmony-frontend/package.json`
- Adds a `LOCAL-ONLY` callout in the README terminal block for the frontend commands
- Updates all README references to use the new script names

## Why
These scripts hardcode `http://localhost:4000` and `http://localhost:3000` and are intentionally for local integration testing only. The new names make the local-only scope obvious at the call-site, reducing the risk of accidental use in CI or deploy pipelines. CI already uses `npm run build` with explicit env vars, so no CI changes are needed.

## Test plan
- [ ] `npm run build:local-e2e` runs from `harmony-frontend/` and builds with localhost env vars
- [ ] `npm run start:local-e2e` starts the frontend on port 3000
- [ ] The local integration README section (`Start the app for local integration testing`) reflects the new command names

Closes #423


acabrera04 (approved):
- [x] **Prioritize Logic over Style:** The change is a pure rename plus README alignment; the local-only intent is clearer and the script behavior stays identical.
- [x] **Security First:** No new execution path or secret handling is introduced, and the README warning reduces the chance of accidentally baking localhost URLs into non-local workflows.
- [x] **Architectural Alignment:** This does not alter application architecture; it stays within the existing local integration-test workflow.
- [x] **Issue Completion:** The PR satisfies issue #423's goal by making the frontend scripts explicitly local-only in both `harmony-frontend/package.json` and the README instructions.
- [x] **No Nitpicking:** I limited the review to correctness, risk, and workflow impact.
- [x] **Avoid Repetition:** There were no prior human review findings on this PR for me to duplicate.
- [x] **Iterative Reviews:** This is my first review on the current head, so there were no earlier concerns from me to re-check or resolve.
- [x] **Prevent CI Failures:** The renamed commands are updated consistently in the touched README references, and the current PR checks are passing.

Bundled review:

Approved. The rename is internally consistent, preserves the exact localhost wiring of the existing scripts, and makes the local-only scope much harder to miss at the call site and in the setup docs. I also checked the current head state and the relevant CI runs are green, so I don't see a blocking logic or workflow regression here.

PR #458, opened 2026-04-23
[codex] Retry failed cloud fixture discovery in integration tests

### LINKED ISSUE #419: test(integration): retry cloud fixture discovery after transient failure
Follow-up from PR #417 approval review.

Problem:
`tests/integration/env.ts` caches `cloudFixturePromise` permanently. If the first fixture discovery attempt fails due to a transient network error, every later call in the same test process reuses the rejected promise and cannot recover.

Requested change:
- clear the cached promise on rejection so later calls can retry
- preserve the current single-flight behavior for successful discovery

Context:
Raised as a non-blocking review comment on PR #417.

acabrera04:
## Summary
- clear the cached cloud fixture discovery promise after a rejected lookup so later calls can retry in the same process
- add integration coverage for both retry-after-failure and successful single-flight behavior

## Root Cause
`tests/integration/env.ts` cached the first discovery promise permanently. If that first lookup failed due to a transient network error, every later `getCloudFixture()` call in the same Jest process reused the rejected promise and could never recover.

## Validation
- `cd tests/integration && npm test -- env.test.ts`
- `cd tests/integration && npx tsc --noEmit -p tsconfig.json`
- `cd harmony-frontend && npm test -- --runInBand`
- `cd harmony-backend && npm test -- --runInBand` *(blocked locally in this worktree because `DATABASE_URL` is unset)*

Closes #419.

### DIFF
--- a/llm-logs/2026-04-23-122348-session-6d495841.md
+++ b/llm-logs/2026-04-23-122348-session-6d495841.md
@@ -1,0 +1,308 @@
+# Session Log 2026-04-23-122348
+
+## User
+
+<command-message>review</command-message>
+<command-name>/review</command-name>
+<command-args>458 for issue completion and approve if and only if the issue is complete. otherwise request changes</command-args>
+
+## User
+
+You are an expert code reviewer. Follow these steps:
+
+      1. If no PR number is provided in the args, run `gh pr list` to show open PRs
+      2. If a PR number is provided, run `gh pr view <number>` to get PR details
+      3. Run `gh pr diff <number>` to get the diff
+      4. Analyze the changes and provide a thorough code review that includes:
+         - Overview of what the PR does
+         - Analysis of code quality and style
+         - Specific suggestions for improvements
+         - Any potential issues or risks
+
+      Keep your review concise but thorough. Focus on:
+      - Code correctness
+      - Following project conventions
+      - Performance implications
+      - Test coverage
+      - Security considerations
+
+      Format your review with clear sections and bullet points.
+
+      PR number: 458 for issue completion and approve if and only if the issue is complete. otherwise request changes
+
+## Assistant
+
+I'll start by fetching the PR details and diff for PR #458.
+
+## Assistant
+
+**Tool call: `Bash`**
+```json
+{
+  "command": "gh pr view 458 2>&1",
+  "description": "Get PR #458 details"
+}
+```
+
+## Assistant
+
+**Tool call: `Bash`**
+```json
+{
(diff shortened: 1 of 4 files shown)


AvanishKulkarni (approved):

- [x] **Logic over Style** — The `.catch()` handler correctly nulls `cloudFixturePromise` synchronously before re-throwing, so any subsequent caller will always see `null` and start a fresh attempt. No logic issues.
- [x] **Security First** — No security implications; this is an internal test utility touching no auth paths, user data, or external-facing APIs.
- [x] **Architectural Alignment** — Integration-test utility change only; unified-backend-architecture.md is not applicable here.
- [x] **Issue Completion** — Issue #419 requested exactly two things: (1) clear cached promise on rejection so later calls retry, and (2) preserve single-flight behavior for successful discovery. Both are delivered and verified by the new tests.
- [x] **No Nitpicking** — N/A.
- [x] **Iterative Reviews** — First review; N/A.
- [x] **Prevent CI Failures** — All checks pass: Backend Lint/Build, Frontend Lint/Build, Backend Tests, Frontend Tests, Integration Tests, Vercel.

---

## Review

The PR is minimal, targeted, and fully satisfies issue #419.

**`tests/integration/env.ts` (core fix, +4 lines)**

The original code cached the raw `resolveCloudFixtureFromPublicApi()` promise forever. The fix chains a `.catch()` onto that promise and assigns *the chained promise* to `cloudFixturePromise`. When the inner promise rejects, the `.catch()` callback executes synchronously (as a microtask), nulling `cloudFixturePromise` before the rejection propagates to any awaiting caller — so the next `getCloudFixture()` invocation always finds `null` and starts fresh. The single-flight guarantee for in-flight successful requests is unchanged because the assignment still happens before `return cloudFixturePromise`.

**`tests/integration/env.test.ts` (new, 86 lines)**

Two tests covering both issue requirements:

1. *Retry-after-failure* — mocks fetch to 503 on the first call, then returns valid data on the next two calls. Confirms the first call rejects with the right message, the second call resolves correctly, and exactly 3 fetch invocations occurred. This directly validates the fix.

2. *Single-flight for concurrent callers* — holds the servers response unresolved while two concurrent `getCloudFixture()` calls are in flight, confirms only one fetch was issued, then resolves both to the same fixture. This validates the preserved behavior.

The use of `jest.resetModules()` + dynamic `import('./env')` in `loadCloudEnvModule()` is the correct pattern for resetting module-level singletons between test cases.

No issues found. Issue #419 is complete.

PR #470, opened 2026-04-25
fix(security): enforce isPublic filter on public server slug endpoints (CWE-862)

### LINKED ISSUE #464: [Security] Public endpoint leaks private server metadata by slug
## Summary
Public REST endpoint `GET /api/public/servers/:serverSlug` returns server data by slug without enforcing `isPublic=true`.

## Severity
High

## CWE
CWE-862 (Missing Authorization)

## Affected code
- `harmony-backend/src/routes/public.router.ts` (`/servers/:serverSlug` route)

## Reproduction
1. Identify/guess a private server slug.
2. Request `GET /api/public/servers/<private-slug>`.
3. Observe `200 OK` with server metadata.

## Expected
Private servers should not be retrievable through a public endpoint.

## Actual
Endpoint returns metadata for any existing slug regardless of public/private visibility.

## Impact
Unauthenticated private server enumeration and metadata disclosure.

## Suggested fix
Query with `where: { slug: req.params.serverSlug, isPublic: true }` and return `404` when not matched.

AvanishKulkarni:
## Summary

Closes #464

- `GET /api/public/servers/:serverSlug` was querying by slug alone, allowing unauthenticated callers to retrieve metadata for private servers.
- Same gap existed in `GET /api/public/servers/:serverSlug/channels` and `GET /api/public/servers/:serverSlug/channels/:channelSlug`.
- Fix: added `isPublic: true` to the Prisma `where` clause in all three routes so Prisma returns `null` (→ 404) for any private server, preventing enumeration and metadata disclosure.

## Test plan

- [x] Added `CWE-862` regression test for each affected route asserting that a private server slug returns 404 and that `findUnique` is called with `isPublic: true` in the where clause.
- [x] All 55 existing `public.router.test.ts` tests still pass (`npm test -- --testPathPattern="public.router"`).

### DIFF
--- a/harmony-backend/src/routes/public.router.ts
+++ b/harmony-backend/src/routes/public.router.ts
@@ -169,7 +169,7 @@ export function createPublicRouter(store?: Store) {
   router.get('/servers/:serverSlug', async (req: Request, res: Response) => {
     try {
       const server = await prisma.server.findUnique({
-        where: { slug: req.params.serverSlug },
+        where: { slug: req.params.serverSlug, isPublic: true },
         select: {
           id: true,
           name: true,
@@ -224,7 +224,7 @@ export function createPublicRouter(store?: Store) {
   router.get('/servers/:serverSlug/channels', async (req: Request, res: Response) => {
     try {
       const server = await prisma.server.findUnique({
-        where: { slug: req.params.serverSlug },
+        where: { slug: req.params.serverSlug, isPublic: true },
         select: { id: true },
       });
 
@@ -276,7 +276,7 @@ export function createPublicRouter(store?: Store) {
   router.get('/servers/:serverSlug/channels/:channelSlug', async (req: Request, res: Response) => {
     try {
       const server = await prisma.server.findUnique({
-        where: { slug: req.params.serverSlug },
+        where: { slug: req.params.serverSlug, isPublic: true },
         select: { id: true },
       });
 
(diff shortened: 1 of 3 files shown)


acabrera04 (approved):
- [x] Prioritize Logic over Style: The fix closes the server-slug authorization gap at the query boundary with minimal, targeted changes and no unnecessary stylistic churn.
- [x] Security First: The changed routes now require `isPublic: true` alongside the unique `slug`, which prevents unauthenticated retrieval of private-server metadata through these public slug endpoints.
- [x] Architectural Alignment: The changes stay within the existing `public.router` pattern and preserve the route-level access model already used for public resources.
- [x] Issue Completion: The linked issue's expected fix is present on `GET /api/public/servers/:serverSlug`, and the PR also applies the same guard to the related slug-based channel endpoints to close the broader leak surface.
- [x] No Nitpicking: No blocking style-only concerns identified.
- [x] Avoid Repetition: Existing review comments only raise a Prisma typing concern that does not apply on Prisma 5.x when a unique field (`slug`) is included in `findUnique(...where)` with additional filters, so I am not repeating them.
- [x] Iterative Reviews: I inspected the current review state and inline threads first; there are no prior human review findings from me to re-litigate.
- [x] Prevent CI Failures: Live GitHub checks are green, and the Prisma usage matches the current documented `WhereUniqueInput` behavior for additional non-unique filters.

Approved. I verified the live diff, checks, existing review comments, and the linked issue. The fix is appropriately scoped, closes the reported unauthorized metadata disclosure path, and includes regression coverage for the affected public slug endpoints.

PR #471, opened 2026-04-26
fix: block reserved admin email/username at registration (#466)

### LINKED ISSUE #466: [Security] Prevent reserved admin email registration privilege escalation
## Summary
A new user can register the reserved email `admin@harmony.dev` via the normal public registration endpoint, and the system-admin check is based on email equality. This can grant effective system-admin permissions to an attacker.

## Severity
Critical

## CWE
- CWE-287 (Improper Authentication)
- CWE-266 (Incorrect Privilege Assignment)

## Affected code
- `harmony-backend/src/routes/auth.router.ts` (public registration accepts arbitrary email)
- `harmony-backend/src/services/auth.service.ts` (`register` persists attacker-controlled email)
- `harmony-backend/src/lib/admin.utils.ts` (`ADMIN_EMAIL = 'admin@harmony.dev'`, `isSystemAdmin` email comparison)
- `harmony-backend/src/services/permission.service.ts` (system-admin short-circuit in `checkPermission`)

## Reproduction (safe)
1. If no account currently owns `admin@harmony.dev`, register a new account using that email through `/api/auth/register`.
2. Authenticate with that account.
3. Invoke an admin-gated path and observe authorization behavior consistent with system-admin bypass.

## Expected
Reserved/admin identities must not be claimable through public registration, and privileged access must be derived from explicit role/claims instead of raw email string matching.

## Actual
Public registration can mint the reserved admin identity, and authorization logic trusts that identity for elevated access.

## Impact
Potential full privilege escalation from unauthenticated user to broad administrative capabilities.

## Suggested fix
1. Block/deny reserved admin emails at registration (`admin@harmony.dev` and any future reserved identifiers).
2. Remove email-string-based admin privilege checks.
3. Use explicit persisted admin role/claim and strict environment gating for any dev-only override.
4. Add regression tests covering reserved-email registration and admin check behavior.

## Notes
Related to existing dev-admin hardening work, but this report is specifically about the **public registration path** claiming the reserved admin identity.

AvanishKulkarni:
## Summary

Fixes #466 — privilege escalation via public registration claiming `admin@harmony.dev`.

- **Block reserved emails at registration**: `auth.service.ts` now rejects any email in `RESERVED_EMAILS` (case-insensitive) with HTTP 403 before the DB is consulted. `admin@harmony.dev` and any future reserved addresses can be added to the set in `admin.utils.ts`.
- **Block reserved usernames at registration**: Similarly rejects username `admin` (and future reserved names) so the dev-admin bootstrap path in `ensureAdminUser` can never be blocked by a pre-claimed username.
- **Production env guard on `isSystemAdmin`**: Returns `false` immediately when `NODE_ENV === 'production'`, ensuring the email-based admin check is never reachable in production even if another write path bypasses the registration block.

## Test plan

- [ ] `POST /api/auth/register` with `admin@harmony.dev` → 403 Reserved
- [ ] `POST /api/auth/register` with `Admin@Harmony.Dev` (mixed case) → 403 Reserved
- [ ] `POST /api/auth/register` with username `admin` → 403 Reserved
- [ ] `isSystemAdmin` returns `false` in production env even for a user whose email matches `ADMIN_EMAIL`
- [ ] `isSystemAdmin` still returns `true` in test/dev env for the admin user
- [ ] Existing auth flow tests (register, login, logout, refresh) continue to pass
- [ ] Run `npm test` in `harmony-backend` — all non-infrastructure tests pass

### DIFF
--- a/harmony-backend/src/lib/admin.utils.ts
+++ b/harmony-backend/src/lib/admin.utils.ts
@@ -10,14 +10,22 @@ import { prisma } from '../db/prisma';
 
 export const ADMIN_EMAIL = 'admin@harmony.dev';
 
+/** Lowercase emails that must not be claimable via public registration. */
+export const RESERVED_EMAILS = new Set<string>([ADMIN_EMAIL]);
+
+/** Lowercase usernames that must not be claimable via public registration. */
+export const RESERVED_USERNAMES = new Set<string>(['admin']);
+
 /** Cached admin user ID to avoid repeated DB lookups. */
 let _adminUserId: string | null = null;
 
 /**
  * Returns true if the given userId belongs to the system admin account.
+ * Never returns true in production — this utility is dev/test only.
  * Caches the result after the first positive lookup.
  */
 export async function isSystemAdmin(userId: string): Promise<boolean> {
+  if (process.env.NODE_ENV === 'production') return false;
   if (_adminUserId === userId) return true;
   const user = await prisma.user.findUnique({
     where: { id: userId },
--- a/harmony-backend/src/services/auth.service.ts
+++ b/harmony-backend/src/services/auth.service.ts
@@ -4,7 +4,7 @@ import crypto from 'crypto';
 import { Prisma } from '@prisma/client';
 import { TRPCError } from '@trpc/server';
 import { serverMemberService } from './serverMember.service';
-import { ADMIN_EMAIL } from '../lib/admin.utils';
+import { ADMIN_EMAIL, RESERVED_EMAILS, RESERVED_USERNAMES } from '../lib/admin.utils';
 import { userRepository } from '../repositories/user.repository';
 import { serverRepository } from '../repositories/server.repository';
 import { serverMemberRepository } from '../repositories/serverMember.repository';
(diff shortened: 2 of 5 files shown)


acabrera04 (approved):
- [x] Prioritize Logic over Style: Reserved email/username checks are added at the top of registration, and the production-only guard prevents the dev-only system-admin check from being reachable in prod.
- [x] Security First: The change closes the public-registration path for reserved admin identifiers and adds defense in depth around `isSystemAdmin`.
- [x] Architectural Alignment: The patch stays within the existing backend auth/admin utility patterns and does not introduce a parallel authorization path.
- [x] Issue Completion: The PR addresses the issue scope described in the PR body with both enforcement and regression coverage.
- [x] No Nitpicking: I did not find a style-only issue worth commenting on.
- [x] Avoid Repetition: I reviewed the existing PR comments and am not repeating prior bot suggestions.
- [x] Iterative Reviews: This is my first review on PR #471, so there are no prior comments from me to resolve.
- [x] Prevent CI Failures: The current GitHub checks on the latest head are green, and I did not find an obvious pipeline-breaking regression in the diff.

Security fix looks good on the current head. The reserved-identity checks are in the right place, the production guard matches the existing dev-only admin flow, and the latest GitHub checks are green. I did not find an additional blocker that would justify holding this PR.

PR #474, opened 2026-04-26
Finalize Sprint 5 SEO README and deployer guide

### LINKED ISSUE #363: Sprint 5: S14 README + deployer guide final pass
## Summary
S14: README + deployer guide final pass
- Fold SEO feature usage, override flow, and regeneration behavior into README
- Update deployer guide with any new env vars introduced in Phase A–C (NLP keys, R2 bucket for OG images if any, etc.)

## Acceptance Criteria
- README documents: how to view generated SEO tags, how a server admin overrides them, and how regeneration is triggered
- Deployer guide lists every new env var added during Sprint 5 with purpose and example value
- README links to the deployed frontend URL, backend URL, and submission evidence bundle
- A fresh contributor can follow README instructions to run the app locally and see SEO tags on a public channel

## Ownership
- Assignee: `acabrera04`
- Due: `May 2, 2026`

## Dependencies
- Blocked by: #359
- Unblocks: #364

acabrera04:
## Summary
- update the root README with the real Sprint 5 SEO usage flow: local public-channel tag inspection, server-admin override UI, regeneration behavior, live deployment links, and an SEO evidence-bundle entrypoint
- document the Sprint 5 SEO env-var inventory in the Vercel deploy guide and make explicit that the shipped feature reuses existing deploy-time vars rather than adding hidden new ones
- fix the frontend SEO regeneration request path and add a regression test discovered during the docs audit

## Reviewer note about the bug fix
This PR is primarily issue #363 documentation work, but the README audit exposed a small real product bug in the server-admin SEO flow.

The docs now explain that a server admin can click `Regenerate Metadata` from the SEO Preview settings page. While verifying that path against the implementation, I found the frontend was posting regeneration requests to:
- `/api/admin/channels/:channelId/meta-tags`

But the backend route that actually exists is:
- `/api/admin/channels/:channelId/meta-tags/jobs`

So without this fix, the documented admin regeneration flow would have been inaccurate and the button would not hit the intended backend endpoint. The code change keeps the docs and shipped behavior aligned rather than expanding scope beyond the issue.

## Testing
- code-path verification against the current frontend/backend SEO implementation
- attempted targeted frontend Jest verification, but this worktree is missing frontend test deps and `npm test` failed with `sh: jest: command not found`

Closes #363


acabrera04:
Reviewer note on the README indexing section:

The new `robots.txt` note is not saying Harmony's current SEO implementation is wrong. It is clarifying the search-engine distinction between crawl control and index control.

Current Harmony behavior is the intended pattern:
- `PUBLIC_INDEXABLE`: crawlable, included in sitemap, normal indexable metadata
- `PUBLIC_NO_INDEX`: still crawlable so Google can read the page-level `noindex`, but excluded from the sitemap
- `PRIVATE`: not part of the public indexable surface

Why the README calls this out: a `robots.txt` disallow by itself is not a reliable "keep this out of Google" mechanism, because Google can still index a discovered URL without fetching the page. For the public-but-unindexed case, Harmony needs the page to remain fetchable so crawlers can actually see the `noindex` metadata. So the current implementation is the correct one; the README note is there to prevent future deployers from trying to replace that behavior with a `robots.txt` block.

AvanishKulkarni (changes requested):
## Review

The bug fix is correct and verified — the backend router confirms  is the right path, and the old  path would have 404'd. Good catch while doing the docs audit.

**One must-fix before merge:**

README section ordering is broken:  (Server-admin override) appears *after*  (Search Console registration), which breaks alphabetical order. The sections should read  in the rendered doc. This looks like a copy/paste ordering mistake when the sections were assembled.

**Other minor notes (non-blocking):**
- The new test file couldn't be run locally (`jest: command not found`). Please confirm the tests pass in CI before merging.
- Smoke-check commands use `rg` (ripgrep) which may not be on all machines — consider noting `grep -E` as a fallback.

Fix the section ordering and this is good to go.

AvanishKulkarni (changes requested):
## Review

The bug fix is correct and verified — the backend router confirms `POST .../meta-tags/jobs` is the right path, and the old `.../meta-tags` path would have 404d. Good catch while doing the docs audit.

**One must-fix before merge:**

README section ordering is broken: `3b` (Server-admin override) appears *after* `3d` (Search Console registration), which breaks alphabetical order. The sections should read `3a → 3b → 3c → 3d` in the rendered doc. This looks like a copy/paste ordering mistake when the sections were assembled.

**Other minor notes (non-blocking):**
- The new test file could not be run locally (`jest: command not found`). Please confirm the tests pass in CI before merging.
- Smoke-check commands use `rg` (ripgrep) which may not be on all machines — consider noting `grep -E` as a fallback.

Fix the section ordering and this is good to go.

acabrera04:
> Re: comment 4177250650

Fixed in 97176ba. Reordered the README SEO subsections so they now render as 3a -> 3b -> 3c -> 3d. I also reran the targeted frontend Jest coverage for the SEO admin flow and channel metadata paths successfully.

acabrera04:
> Re: comment 4177250566

Fixed in 97176ba. Reordered the README SEO subsections so they now render as 3a -> 3b -> 3c -> 3d. I also reran the targeted frontend Jest coverage for the SEO admin flow and channel metadata paths successfully.

acabrera04:
> Re: comment 4321146382

Already addressed. The README keeps the indexing clarification, and the reviewer-reported subsection ordering issue has now been fixed in 97176ba so the SEO docs read in the intended 3a -> 3b -> 3c -> 3d order.

Aiden-Barrera (dismissed):

Fixes the regeneration endpoint path (`/meta-tags` → `/meta-tags/jobs`) and adds a unit test for `metaTagAdminService`. The path fix is correct, but the service has unhandled async errors.

### Issues
- **CRITICAL:** `harmony-frontend/src/services/metaTagAdminService.ts` — `triggerMetaTagRegeneration` and `getMetaTagRegenerationStatus` make async fetch calls with no try/catch or error boundary; network failures and non-2xx responses will throw unhandled rejections in the caller.

### Non-blocking
- `metaTagAdminService.test.ts` — The test demonstrates user-controlled channel IDs being URL-encoded (good), but the service layer has no validation to reject structurally invalid channel identifiers before constructing the request.

---
*Reviewed by claude-haiku*

acabrera04:
> Re: comment 4177624119

Wont fix: this service intentionally propagates request failures to its callers rather than swallowing them locally. requestJson already turns non-2xx responses into Errors, the server actions await those promises, and the client SEO Preview component wraps regeneration and status polling in try/catch and shows the error state to the user. So network and non-2xx failures are already handled at the action and UI boundary rather than becoming unhandled rejections. The non-blocking channel-id note also does not apply here because the client does not send arbitrary IDs directly; the server actions resolve channel IDs from trusted route slugs before calling this service.

acabrera04:
> Re: comment 4177250650

Already addressed in a prior follow-up. The README subsection order was fixed in 97176ba, and the targeted frontend Jest coverage for the SEO admin flow and metadata paths was rerun successfully.

acabrera04:
> Re: comment 4177250566

Already addressed in a prior follow-up. The README subsection order was fixed in 97176ba, and the targeted frontend Jest coverage for the SEO admin flow and metadata paths was rerun successfully.

acabrera04:
> Re: comment 4321146382

Already addressed. The README indexing clarification stays as-is, and the related README ordering issue was fixed in 97176ba in the previous follow-up pass.

AvanishKulkarni (approved):
All review issues are resolved:

- Section ordering (3a → 3b → 3c → 3d) confirmed fixed in 97176ba.
- The CRITICAL error-handling concern is a false positive: `requestJson` throws on non-2xx, server actions propagate, and `SeoPreviewSection.tsx` wraps every call in try/catch with `setError` — there are no unhandled rejections.
- Non-blocking notes acknowledged; not blocking.

Code is correct, tests are solid, docs are accurate. ✅

PR #510, opened 2026-04-28
[codex] Implement lease-based presence tracking

### LINKED ISSUE #490: feat: Real-time presence status (online, idle, dnd, offline)
## Summary
Implement automatic presence tracking that detects user activity and broadcasts status changes in real-time via SSE.

## Expected Behavior
- User status automatically updates to ONLINE when active
- Status changes to IDLE after 5 minutes of inactivity
- Status changes to OFFLINE when user closes browser
- Status changes broadcast to all connected peers in real-time

acabrera04:
## Summary
- add authenticated presence heartbeat endpoint and Redis-backed lease expiry
- track browser activity with online/idle heartbeats instead of pagehide offline writes
- wire real-time member status updates into the visible members sidebar

## Why
Issue #490 asks for automatic online, idle, and offline presence broadcast over existing SSE. Offline is handled by server-side lease expiry so minimizing or switching tabs does not immediately mark a user offline.

## Verification
- npm --prefix harmony-backend test -- presence.router.test.ts presence.service.test.ts
- npm --prefix harmony-frontend test -- usePresenceTracker.test.tsx
- npx tsc --noEmit in harmony-frontend

Notes: frontend production build was previously blocked locally by sandbox network access to Google Fonts, not by this change.

### DIFF
--- a/harmony-backend/src/app.ts
+++ b/harmony-backend/src/app.ts
@@ -10,11 +10,13 @@ import { authRouter } from './routes/auth.router';
 import { createPublicRouter } from './routes/public.router';
 import { seoRouter } from './routes/seo.router';
 import { eventsRouter } from './routes/events.router';
+import { presenceRouter } from './routes/presence.router';
 import { attachmentRouter } from './routes/attachment.router';
 import { adminMetaTagRouter } from './routes/admin.metaTag.router';
 import { instanceId } from './lib/instance-identity';
 import { createLogger } from './lib/logger';
 import { redis } from './db/redis';
+import { presenceService } from './services/presence.service';
 
 const logger = createLogger({ component: 'app', instanceId });
 
@@ -56,6 +58,8 @@ export interface CreateAppOptions {
 }
 
 export function createApp(options: CreateAppOptions = {}) {
+  presenceService.startSweeper();
+
   const isE2E = process.env.NODE_ENV === 'e2e';
   // Each limiter calls makeStore() independently so it gets its own instance.
   const makeStore = (prefix: string): Store | undefined =>
@@ -154,6 +158,9 @@ export function createApp(options: CreateAppOptions = {}) {
 
   // Real-time SSE endpoints
   app.use('/api/events', eventsRouter);
+
+  // Presence updates from authenticated browser clients
+  app.use('/api/presence', presenceRouter);
 
   // Attachment upload + file serving
   app.use('/api/attachments', attachmentRouter);
--- a/harmony-backend/src/routes/presence.router.ts
+++ b/harmony-backend/src/routes/presence.router.ts
@@ -1,0 +1,59 @@
+import { NextFunction, Router, Request, Response } from 'express';
+import { UserStatus } from '@prisma/client';
+import { z } from 'zod';
+import { authService } from '../services/auth.service';
+import { presenceService } from '../services/presence.service';
+
+export const presenceRouter = Router();
+
+const PresenceStatusSchema = z.enum([UserStatus.ONLINE, UserStatus.IDLE]);
+const PresenceBodySchema = z.object({ status: PresenceStatusSchema });
+type ActivePresenceStatus = z.infer<typeof PresenceStatusSchema>;
+
+function getBearerToken(req: Request): string | null {
+  const authHeader = req.headers.authorization;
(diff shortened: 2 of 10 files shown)


AvanishKulkarni (commented):
## Code Review: PR #510 — Lease-Based Presence Tracking

### Overview

The overall design is sound — a Redis sorted-set lease model is a good fit here. It avoids false offline signals from tab-switching, handles multi-replica deployment via a Lua atomic-remove race, and the `MembersSidebar` is correctly wired to `localMembers` (SSE-updated) rather than the stale prop. Three issues warrant attention before merge.

---

### Issues

#### 1. Duplicates `requireAuth` middleware (significant)

`presence.router.ts` re-implements `getBearerToken` + `verifyToken` inline, but an identical `requireAuth` middleware already exists in `auth.middleware.ts`. This duplicates ~20 lines and diverges from every other authenticated route.

The router should use `requireAuth` directly:

```ts
import { requireAuth, AuthenticatedRequest } from '../middleware/auth.middleware';

presenceRouter.post('/status', requireAuth, async (req, res, next) => {
  const userId = (req as AuthenticatedRequest).userId;
  // ...
});
```

#### 2. `setPresence` throttle can suppress IDLE→ONLINE transition (bug)

In `usePresenceTracker.ts`, the ONLINE throttle check fires before the status-change check:

```ts
if (!force && status === 'ONLINE' && now - lastActivePostRef.current < ACTIVE_THROTTLE_MS) return; // ← runs first
if (!force && statusRef.current === status) return;
```

Scenario: user is active (ONLINE posted at t=0), goes idle at t=4m59s (via the idle timer), immediately moves mouse at t=5m+5s. `now - lastActivePost = 5m+5s > 30s` — fine. But if the gap is shorter (e.g., ONLINE posted at t=28s ago, then IDLE fires at t=29s, user moves mouse at t=30s), the throttle fires first and swallows the IDLE→ONLINE transition. `statusRef` stays `IDLE` for up to another 30s.

Fix: dedup same-status first, *then* throttle:

```ts
if (!force && statusRef.current === status) return;
if (!force && status === 'ONLINE' && now - lastActivePostRef.current < ACTIVE_THROTTLE_MS) return;
```

#### 3. `updateUserStatusIfChanged` updates even when user is not found

```ts
const user = await userRepository.findSelf(userId);
if (!user || user.status !== status) {   // ← calls updateUser even when user is null
  await userService.updateUser(userId, { status });
}
```

If `findSelf` returns `null` (deleted/deactivated user), `updateUser` is still called with a non-existent ID. Should be:

```ts
if (!user) return;
if (user.status !== status) {
  await userService.updateUser(userId, { status });
}
```

---

### Minor

- **`updatePresence` helper in router**: The wrapper that calls `renewLease` then `res.status(204).send()` adds indirection for trivial logic — can be inlined.
- **Missing router test for invalid token**: Tests cover a missing Authorization header but not a present-but-invalid token (where `verifyAccessToken` throws). The 401 branch in `verifyToken` is untested.
- **`ACTIVE_THROTTLE_MS === HEARTBEAT_INTERVAL_MS`**: Both are `30_000`. A single named constant (e.g. `PRESENCE_INTERVAL_MS`) would make the intentional coupling explicit.

---

### Positives

- Lua atomic-remove correctly handles the multi-replica race in `expireStaleLeases`.
- `sweeper.unref?.()` prevents the interval from blocking process exit.
- `currentUser` now derives `status`/`role` from `currentMemberRecord` — fixes the stale auth snapshot in the shell.
- Test coverage is solid for happy paths and key edge cases (race condition, DND rejection, pagehide no-op, heartbeat).

acabrera04:
> Re: comment 4191693937

Fixed in 5a2c1b2. Reused requireAuth in the presence router, inlined the trivial route helper, fixed the IDLE-to-ONLINE throttle ordering, skipped status writes for missing users, consolidated the presence interval constant, and added invalid-token/deleted-user/throttle regression coverage.

AvanishKulkarni (approved):
All review items addressed. The router now uses  correctly, the  guard order is fixed (status-change dedup before throttle), the null-user guard is in place, and test coverage fills the gaps called out. Good to merge.

AvanishKulkarni (approved):
All review items addressed. The router now uses requireAuth correctly, the setPresence guard order is fixed (status-change dedup before throttle), the null-user guard is in place, and test coverage fills the gaps called out. Good to merge.

PR #511, opened 2026-04-28
Make frontend builds sandbox safe

acabrera04:
## Summary

- remove `next/font/google` from the frontend root layout so production builds no longer fetch Google Fonts
- define the Inter font token through local CSS fallbacks
- use Next's webpack build mode for frontend build scripts so sandboxed builds avoid Turbopack's internal process/port restriction
- update the `.xml` sitemap route handler and tests for the stricter generated route type exposed by the webpack build

## Verification

- `rtk npx prettier --write package.json src/__tests__/seo-routes.test.ts src/app/sitemap/[serverSlug].xml/route.ts src/app/layout.tsx src/app/globals.css`
- `./node_modules/.bin/tsc --noEmit`
- `rtk npm run lint`
- `rtk npm test -- --runInBand`
- `rtk npm run build`

Note: the build still emits existing warnings for the deprecated `middleware` convention and the module type of `tailwind.config.ts`; it exits successfully.

### DIFF
--- a/harmony-frontend/package.json
+++ b/harmony-frontend/package.json
@@ -4,9 +4,11 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build --webpack",
+    "build": "next build",
+    "build:sandbox": "next build --webpack",
     "start": "next start",
-    "build:local-e2e": "NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next build --webpack",
+    "build:local-e2e": "NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next build",
+    "build:local-e2e:sandbox": "NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next build --webpack",
     "start:local-e2e": "PORT=3000 NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next start",
     "test": "jest --forceExit",
     "test:e2e": "playwright test",
--- a/harmony-frontend/package.json
+++ b/harmony-frontend/package.json
@@ -4,9 +4,9 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "next build --webpack",
     "start": "next start",
-    "build:local-e2e": "NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next build",
+    "build:local-e2e": "NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next build --webpack",
     "start:local-e2e": "PORT=3000 NEXT_PUBLIC_API_URL=http://localhost:4000 NEXT_PUBLIC_BASE_URL=http://localhost:3000 next start",
     "test": "jest --forceExit",
     "test:e2e": "playwright test",
--- a/harmony-frontend/src/__tests__/seo-routes.test.ts
+++ b/harmony-frontend/src/__tests__/seo-routes.test.ts
@@ -99,7 +99,7 @@ describe('frontend SEO route handlers', () => {
     );
 
     const response = await getServerSitemap(new Request('https://harmony.chat/sitemap/demo.xml'), {
-      params: Promise.resolve({}),
+      params: Promise.resolve({ serverSlug: 'demo' }),
     });
 
     expect(global.fetch).toHaveBeenCalledWith('https://api.harmony.chat/sitemap/demo.xml', {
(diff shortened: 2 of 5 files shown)


harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts
```diff
@@ -5,18 +5,20 @@ import { proxySitemapXml } from '@/lib/sitemap-response';
...
+  params: Promise<Record<string, never>>;
 }
 
 /**
  * Per-server sitemap entrypoints stay on the frontend host and serve a cached
  * proxy of the backend XML generator, revalidating periodically, so crawlers
  * never need the API domain as the primary SEO surface.
  */
-export async function GET(request: Request, context?: RouteContext) {
-  const params = await context?.params;
```
Comment (AvanishKulkarni) [inline harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts:17]: `Record<string, never>` is a lie — this dynamic segment *does* inject `serverSlug` into params at runtime. Using `never` silences the type error but misrepresents the contract. If Next 16 + webpack generates a stricter type here, use the generated type from `.next/types`, or at minimum use `Record<string, string | string[] | undefined>` as before and cast where needed.

AvanishKulkarni (changes requested):
The font and build changes are clean. The sitemap route workaround for the webpack type mismatch works but has a few rough edges — see inline comments.

harmony-frontend/package.json
```diff
@@ -4,9 +4,9 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "next build --webpack",
     "start": "next start",
```
AvanishKulkarni, line 9: Forcing `--webpack` on the default `build` script changes behavior for everyone, not just sandbox CI. Turbopack is the default for a reason (faster). Consider scoping this to a dedicated `build:ci` or `build:sandbox` script and only using it where the sandbox port restriction actually applies.

harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts
```diff
@@ -5,18 +5,20 @@ import { proxySitemapXml } from '@/lib/sitemap-response';
...
-  const serverSlug = params?.serverSlug;
-  if (typeof serverSlug !== 'string') {
+export async function GET(request: Request, context: RouteContext) {
+  await context.params;
+  const pathname = new URL(request.url).pathname;
+  const slugSegment = pathname.split('/').at(-1) ?? '';
+  const serverSlug = slugSegment.endsWith('.xml') ? slugSegment.slice(0, -4) : slugSegment;
+  if (!serverSlug) {
     return new Response('Sitemap not found.', {
```
Comment (AvanishKulkarni) [inline harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts:22]: URL parsing is fragile compared to reading `params.serverSlug` directly. If the URL is URL-encoded (e.g. `my%20server.xml`), `.at(-1)` gives `my%20server.xml` and the slug will be percent-encoded. The original params approach handles this correctly via Next.js's own routing layer. If the webpack-generated type truly won't accept `serverSlug`, cast it: `const { serverSlug } = await context.params as { serverSlug: string }`.

harmony-frontend/src/__tests__/seo-routes.test.ts
```diff
@@ -119,7 +119,7 @@ describe('frontend SEO route handlers', () => {
     );
 
     const response = await getServerSitemap(new Request('http://localhost:3000/sitemap/demo.xml'), {
-      params: Promise.resolve({ serverSlug: 'demo' }),
+      params: Promise.resolve({}),
     });
```
AvanishKulkarni, line 124: Now that the route reads `serverSlug` from the URL rather than params, this test no longer exercises the slug extraction path — both test cases pass `{}` and rely on the URL string to carry the slug. Add a test that passes a URL with a different slug than `demo` to confirm the extraction isn't hardcoded.

harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts
Comment (acabrera04) [reply harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts:17]: Fixed in 082d529. Replaced the misleading Record<string, never> route context with an unknown params payload and a local cast at the read site, so the handler no longer misrepresents the runtime params contract.

harmony-frontend/package.json
acabrera04: Fixed in 082d529. Restored the default build scripts to Next/Turbopack and added dedicated build:sandbox and build:local-e2e:sandbox scripts for the sandbox webpack path.

harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts
Comment (acabrera04) [reply harmony-frontend/src/app/sitemap/[serverSlug].xml/route.ts:22]: Fixed in 082d529. The route now reads serverSlug from context.params again and only guards missing build-time params, avoiding fragile URL segment parsing.

harmony-frontend/src/__tests__/seo-routes.test.ts
acabrera04: Fixed in 082d529. The tests pass serverSlug through params again because the handler no longer extracts it from the URL.

AvanishKulkarni (approved):
All four concerns addressed in 082d529 — params type contract corrected, webpack flag scoped to dedicated sandbox scripts, URL-based slug extraction replaced with params read, and tests updated accordingly. Good to merge.

PR #522, opened 2026-04-28
[codex] Fix cloud SEO smoke coverage

acabrera04:
## What changed
- harden the public SEO meta-tags route so it falls back to on-the-fly preview generation instead of returning `500` when persisted or cached meta-tag data is broken in deployment state
- expand cloud fixture discovery to collect up to 3 crawler targets across public servers while keeping a primary server/channel for the other cloud smoke tests
- add backend and integration regression coverage for the fallback path and the richer cloud fixture selection behavior

## Why
The post-merge cloud integration failure after PR #510 was not caused by presence tracking directly. The live failure was isolated to the SEO cloud smoke suite:
- the public `meta-tags` endpoint returned `500` for live public channels
- the crawler-UA smoke expected one live server to expose at least 3 public indexable channels, but the current production data spreads them across multiple servers

## Impact
- cloud SEO smoke is resilient to malformed or stale persisted SEO metadata in the deployed database
- crawler-UA coverage still exercises 3 public channels in cloud mode without assuming a single server contains all of them
- local and backend regression tests now cover both behaviors

## Verification
- `npm --prefix harmony-backend test -- public.router.test.ts`
- `npm --prefix tests/integration test -- env.test.ts --runInBand`
- `env BACKEND_URL=http://localhost:4001 FRONTEND_URL=http://localhost:3001 npm --prefix tests/integration run test:cloud`


harmony-backend/src/services/metaTag/metaTagService.ts
```diff
@@ -259,6 +259,94 @@ async function loadGenerationInputs(channelId: string): Promise<{
   };
 }
 
+async function loadPreviewFallbackInputs(channelId: string): Promise<{
```
AvanishKulkarni, line 262: **Duplication concern — `loadPreviewFallbackInputs` vs `loadGenerationInputs`**

This function is nearly identical to `loadGenerationInputs` above it (same channel select shape, same message query). The only difference is that it omits the `GeneratedMetaTags` and custom-overrides selects.

If the reason for a separate function is to deliberately avoid touching the `generated_meta_tags` table (since that's exactly the table that may be broken in the failing state), that constraint should be documented with a comment. If there's no such constraint, consider whether `loadGenerationInputs` can simply be reused here (dropping the `persisted` result). Two copies will diverge silently as the schema evolves.

harmony-backend/src/services/metaTag/metaTagService.ts
```diff
@@ -259,6 +259,94 @@ async function loadGenerationInputs(channelId: string): Promise<{
...
+  persisted: GeneratedMetaTags | null,
+): MetaTagPreview {
+  return {
+    title: tags.title,
+    description: tags.description,
+    ogTitle: tags.openGraph.ogTitle,
+    ogDescription: tags.openGraph.ogDescription,
+    ogImage: tags.openGraph.ogImage,
+    keywords: tags.keywords,
+    generatedAt: persisted?.generatedAt.toISOString() ?? new Date().toISOString(),
```
AvanishKulkarni, line 331: **Misleading `generatedAt` timestamp in fallback path**

When `persisted` is `null` (the fallback case), this resolves to `new Date().toISOString()` — i.e. the wall-clock time of the fallback call. Callers and observability tooling have no way to distinguish "was generated right now via the real pipeline" from "was synthesised on-the-fly as a best-effort fallback". Low priority, but easy to fix: use a distinct field name or a sentinel value (e.g. `null`) for the fallback case so the distinction is visible in logs/API responses.

harmony-backend/src/routes/public.router.ts
```diff
@@ -348,7 +348,21 @@ export function createPublicRouter(store?: Store) {
...
+          logger.warn(
+            {
+              err: previewErr,
+              serverSlug: req.params.serverSlug,
+              channelSlug: req.params.channelSlug,
+              channelId: channel.id,
+            },
+            'Primary public meta tag preview failed; falling back to on-the-fly generation',
+          );
+          preview = await metaTagService.getFallbackMetaTagsForPreview(channel.id);
```
AvanishKulkarni: **No log entry if the fallback itself also throws**

If `getFallbackMetaTagsForPreview` throws (e.g. DB fully down), the error propagates to the Express error handler as an unhandled 500, with no warn-level log identifying this as a *double-fault* (primary failed, fallback also failed). A wrapping warn log before re-throwing would make on-call triage significantly easier. Low priority for the happy path, but worth adding before this goes to production.

tests/integration/env.ts
```diff
@@ -128,11 +144,30 @@ async function resolveCloudFixtureFromPublicApi(): Promise<CloudFixture> {
       .map((ch) => ch.slug);
     if (!publicChannels.length) continue;
 
+    if (primaryFixture === null || publicChannels.length > primaryFixture.publicChannels.length) {
```
AvanishKulkarni: **Target-collection and primary-fixture selection strategies can disagree**

`primaryFixture` is updated on every server in iteration order (picks the server with the most channels), while `publicChannelTargets` fills globally up to 3 on a first-come, first-served basis. If a small server appears earlier in the API response, it can occupy the first slot(s) of `publicChannelTargets` before the eventual primary server is even reached.

The existing test at `env.test.ts:407` exercises exactly this case and passes only because the total target count (1 + 2 = 3) happens to include both servers. In production, if a large primary server appears *second* in the API response and a small server has already consumed 2 of the 3 slots, the primary server would contribute only 1 target — the crawler-UA test could end up exercising channels that aren't on the primary fixture's server at all.

**Suggested fix:** collect all targets in the first pass (or without a cap), determine the primary, then build `publicChannelTargets` in a second pass starting from the primary server to ensure its channels are prioritised.

harmony-backend/tests/public.router.test.ts
```diff
@@ -864,6 +877,89 @@ describe('GET /api/public/servers/:serverSlug/channels/:channelSlug', () => {
...
+
+  it('returns 404 when the channel is missing or private', async () => {
+    mockPrisma.channel.findFirst.mockResolvedValue(null);
+
+    const res = await request(app).get(
+      `/api/public/servers/${SERVER.slug}/channels/${CHANNEL.slug}/meta-tags`,
+    );
+
+    expect(res.status).toBe(404);
+    expect(res.body).toHaveProperty('error', 'Channel not found');
```
AvanishKulkarni, line 960: **Missing negative assertions on the 404 path**

The test correctly asserts `res.status === 404` and `body.error === 'Channel not found'`, but it does not assert that `getMetaTagsForPreview` and `getFallbackMetaTagsForPreview` were *not* called. Without those assertions, a regression that accidentally invokes the service on a missing/private channel would go undetected here. Consider adding:
```ts
expect(mockMetaTagService.getMetaTagsForPreview).not.toHaveBeenCalled();
expect(mockMetaTagService.getFallbackMetaTagsForPreview).not.toHaveBeenCalled();
```

AvanishKulkarni (commented):
## Review Guidelines Checklist

- [x] **1. Prioritize Logic over Style** — Core changes are logically sound: the fallback try/catch is minimal and targeted; cross-server fixture collection is algorithmically correct.
- [x] **2. Security First** — No new attack surface. The fallback path reads the DB directly with a typed `findUnique`; no user-controlled data is echoed unsanitized.
- [x] **3. Architectural Alignment** — `getFallbackMetaTagsForPreview` follows the same service/repository layering pattern. `loadPreviewFallbackInputs` mirrors `loadGenerationInputs` appropriately.
- [ ] **4. Issue Completion** — No linked issue number in the PR body to cross-check ACs against, though the root cause is well described.
- [x] **5. No Nitpicking** — Applying.
- [x] **8. Prevent CI Failures** — Tests cover the new code paths; mock setup is correct; no syntax issues.

---

Overall the two core fixes are correct and well-targeted. The main actionable concern is the **target-collection ordering issue** in `env.ts` (see inline). See inline comments for the other items.

harmony-backend/src/services/metaTag/metaTagService.ts
acabrera04: Fixed in 207e821. I surfaced a distinct isFallbackPreview flag in the preview payload so callers can tell a best-effort fallback response from persisted/generated metadata instead of inferring from generatedAt alone.

tests/integration/env.ts
acabrera04: Fixed in 207e821. Cloud fixture discovery now collects all candidate public-channel sets first, chooses the primary server, and only then builds crawler targets with the primary server prioritized so the target list and primary fixture strategy cannot drift apart.

harmony-backend/src/routes/public.router.ts
acabrera04: Fixed in 207e821. The public meta-tags route now logs a second warning if the fallback preview generation also fails before re-throwing, so double-fault 500s are explicit in logs.

acabrera04:
> Re: comment 4192454255

Addressed in 207e821. I implemented the primary-server target prioritization from the inline note and also tightened the fallback preview observability and route-test coverage from the other review comments.

harmony-backend/src/services/metaTag/metaTagService.ts
acabrera04: Fixed in 207e821. I kept the fallback loader separate and added an inline rationale because this path must avoid reading generated_meta_tags when the persisted relation itself may be the failing state.

harmony-backend/tests/public.router.test.ts
acabrera04: Fixed in 207e821. I added negative assertions on the 404 path so the test now proves neither preview service is invoked for missing/private channels.

AvanishKulkarni (approved):
All five issues addressed — approving.

-  separation is now documented with an intent comment
-  field added to  resolves the misleading  ambiguity
- Double-fault warn log added before re-throw
- Target-collection rewritten to two-pass (primary determined first, then targets filled starting from primary server)
- Negative assertions added to the 404 test case

AvanishKulkarni (approved):
All five issues addressed. Approving.

acabrera04:
> Re: comment 4192515104

Thanks for the re-review and approval.

acabrera04:
> Re: comment 4192514603

Thanks for the detailed re-review and approval.

PR #529, opened 2026-04-29
ci: gate cloud smoke on deploys and narrow required checks

acabrera04:
## Summary
- wait for successful Vercel and Railway deployment statuses before running cloud integration smoke on `main`
- keep required check names stable while fast-pathing them when their owned code paths were untouched
- narrow workflow ownership so frontend checks watch `harmony-frontend/`, backend checks watch `harmony-backend/`, and integration keeps the broader frontend/backend/tests scope

## Verification
- `ruby -e 'require "yaml"; %w[.github/workflows/ci.yml .github/workflows/run-backend-tests.yml .github/workflows/run-frontend-tests.yml .github/workflows/run-integration-tests.yml].each { |f| YAML.load_file(f); puts "YAML OK #{f}" }'`
- inspected live commit status contexts on `main` to wire the deployment wait step to the existing Vercel and Railway signals

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -98,23 +98,24 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Detect non-log changes
+      - name: Detect relevant code/test changes
         id: changes
         uses: dorny/paths-filter@v3
         with:
           filters: |
-            non_log:
-              - '!llm-logs/**'
-              - '**'
+            relevant:
+              - 'harmony-frontend/**'
+              - 'harmony-backend/**'
+              - 'tests/**'
 
-      - name: Skip integration tests for llm-logs-only changes
-        if: steps.changes.outputs.non_log != 'true'
-        run: echo "llm-logs-only change detected; skipping local integration tests."
+      - name: Skip integration tests for non-code changes
+        if: steps.changes.outputs.relevant != 'true'
+        run: echo "No changes under harmony-frontend/, harmony-backend/, or tests/; skipping local integration tests."
 
       # ── Backend setup ──────────────────────────────────────────────────────
 
       - name: Set up Node.js (backend)
-        if: steps.changes.outputs.non_log == 'true'
+        if: steps.changes.outputs.relevant == 'true'
         uses: actions/setup-node@v4
         with:
           node-version: 20
@@ -122,27 +123,27 @@ jobs:
           cache-dependency-path: harmony-backend/package-lock.json
 
       - name: Install backend dependencies
-        if: steps.changes.outputs.non_log == 'true'
+        if: steps.changes.outputs.relevant == 'true'
         working-directory: harmony-backend
         run: npm ci
 
       - name: Generate Prisma client
-        if: steps.changes.outputs.non_log == 'true'
+        if: steps.changes.outputs.relevant == 'true'
         working-directory: harmony-backend
         run: npx prisma generate
 
       - name: Run database migrations
-        if: steps.changes.outputs.non_log == 'true'
+        if: steps.changes.outputs.relevant == 'true'
         working-directory: harmony-backend
         run: npx prisma migrate deploy
 
       - name: Seed mock dataset
-        if: steps.changes.outputs.non_log == 'true'
+        if: steps.changes.outputs.relevant == 'true'
         working-directory: harmony-backend
         run: npm run db:seed:mock
 
       - name: Start backend API
-        if: steps.changes.outputs.non_log == 'true'
+        if: steps.changes.outputs.relevant == 'true'
         working-directory: harmony-backend
         env:
           PORT: "4000"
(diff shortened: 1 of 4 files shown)


.github/workflows/run-integration-tests.yml
```diff
@@ -96,32 +98,52 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Detect relevant code/test changes
+        id: changes
+        uses: dorny/paths-filter@v3
```
AvanishKulkarni, line 103: **Request for change — `workflow_dispatch` silently skips all local integration tests**

`dorny/paths-filter@v3` has no diff context on `workflow_dispatch` events (no base commit to compare against), so it outputs `relevant=false` for all filters. Every subsequent `if: steps.changes.outputs.relevant == 'true'` step is then skipped, making a manually-triggered run a no-op.

Fix: guard each gated step with an additional `|| github.event_name == 'workflow_dispatch'` clause, or add a short override step right after the filter that forces `relevant=true` on dispatch:

```yaml
- name: Override relevant for workflow_dispatch
  if: github.event_name == 'workflow_dispatch'
  run: echo "relevant=true" >> $GITHUB_OUTPUT
```

The same issue applies in `ci.yml`, `run-backend-tests.yml`, and `run-frontend-tests.yml`, but those workflows don't have `workflow_dispatch` triggers, so only this file is currently broken.

.github/workflows/run-integration-tests.yml
```diff
@@ -268,6 +303,50 @@ jobs:
...
+        if: github.event_name != 'workflow_dispatch'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          SHA: ${{ github.sha }}
+        run: |
+          required_contexts=(
+            "Vercel"
+            "overflowing-luck - Harmony backend-api"
+            "overflowing-luck - backend-worker"
```
AvanishKulkarni: **Request for change — hardcoded Railway context names will silently time out if project is renamed**

These two context strings are derived from the Railway project slug (`overflowing-luck`). If the Railway project is ever renamed, redeployed under a new name, or the service names change, the `while true` loop will spin for the full 30-minute `deadline` and then fail every `main` push with a timeout error — there is no early-exit for a "context not found after N minutes" case distinct from a genuine deploy failure.

Minimum fix: document in a comment here that these strings must exactly match the Railway commit status contexts (visible in the GitHub Checks UI on any `main` commit) and how to update them. Better fix: expose them as workflow-level variables (e.g. `vars.RAILWAY_BACKEND_API_CONTEXT` and `vars.RAILWAY_WORKER_CONTEXT`) so they can be updated without a code change when the Railway project changes.

AvanishKulkarni (changes requested):
## Review Checklist

- [x] **Logic over Style** — Correctness of path-gating logic and deployment wait loop examined
- [ ] **Security First** — No secrets exposed, but hardcoded deployment context names create a silent-failure risk (see inline comment)
- [x] **Architectural Alignment** — CI changes don't touch backend/frontend source; no architecture doc applies
- [x] **Issue Completion** — PR summary goals (deploy-gating, path narrowing) are structurally addressed
- [x] **No Nitpicking** — Only blocking/functional issues flagged
- [x] **Avoid Repetition** — First review; no prior comments to check
- [x] **Iterative Reviews** — N/A (first review)
- [ ] **Prevent CI Failures** — Two issues found that will cause silent failures in specific trigger paths (see inline comments)

---

## Overview

This PR does three things: (1) adds `dorny/paths-filter` scoping so lint/test jobs only execute when their owned directory tree changes, (2) adds a deployment-wait gate to the cloud smoke job on `main` so tests don't fire against a stale deploy, and (3) adds `permissions: statuses: read` to the cloud job. The intent is correct and the general structure is sound. Two functional issues need to be fixed before merge.

---

Two inline comments below with required changes.

.github/workflows/run-integration-tests.yml
acabrera04: Fixed in 6f3bbdb. Manual workflow_dispatch runs now bypass diff-based gating so the local integration workflow still executes even when paths-filter has no compare base.

acabrera04:
> Re: comment 4193457723

Fixed in 6f3bbdb. Addressed both inline issues: manual workflow_dispatch runs no longer no-op under path gating, and the deploy-gating status contexts are now documented and configurable via vars with current defaults.

.github/workflows/run-integration-tests.yml
acabrera04: Fixed in 6f3bbdb. The deploy wait step now documents that the commit-status labels must match the live Vercel/Railway contexts and exposes them as configurable vars (with current defaults) instead of relying on unchangeable hardcoded Railway labels.

AvanishKulkarni (approved):
## Review Checklist

- [x] **Logic over Style** — Path-gating and deployment-wait logic is correct
- [x] **Security First** — No secrets exposed; Railway context names are now configurable rather than hardcoded
- [x] **Architectural Alignment** — CI-only changes, no architecture doc concerns
- [x] **Issue Completion** — Both deploy-wait gating and path narrowing are fully implemented
- [x] **No Nitpicking** — Only functional blockers flagged previously
- [x] **Avoid Repetition** — No new issues to add
- [x] **Iterative Reviews** — Both prior change requests resolved (see below)
- [x] **Prevent CI Failures** — No remaining issues

---

Both previously requested changes are resolved in commit `6f3bbdb`:

1. **`workflow_dispatch` bypass** — All gated steps now use `steps.changes.outputs.relevant == 'true' || github.event_name == 'workflow_dispatch'`, so manual dispatches correctly run the full suite instead of silently skipping it. Applied consistently across all four workflow files.

2. **Configurable deployment context names** — The three Railway/Vercel context strings are now exposed as `VERCEL_DEPLOYMENT_CONTEXT`, `RAILWAY_BACKEND_API_CONTEXT`, and `RAILWAY_BACKEND_WORKER_CONTEXT` env vars with sensible defaults, and the wait step includes a comment explaining they must match the live commit-status contexts. No more silent 30-minute timeout on a rename.

:white_check_mark: Approved.

PR #538, opened 2026-04-29
fix(#493): add client-side AuthGuard to /channels and /settings layouts

### LINKED ISSUE #493: fix: Add authentication guard to /channels and /settings layouts
## Summary
Unauthenticated users can currently access protected routes /channels and /settings.

## Expected Behavior
- Unauthenticated users redirected to /auth/login
- Session checked on layout mount
- Only authenticated users access protected routes

## Steps to Reproduce
1. Clear cookies/logout
2. Navigate to /channels/harmony-hq/general
3. Observe: page loads (should redirect to login)

declanblanc:
## Summary
- Creates a new `AuthGuard` client component that checks authentication state on mount and redirects unauthenticated users to `/auth/login` (with `returnUrl` preserved)
- Mounts `AuthGuard` in both `harmony-frontend/src/app/channels/layout.tsx` and `harmony-frontend/src/app/settings/layout.tsx`
- Removes the stale TODO comments from both layouts

The Next.js middleware (`src/middleware.ts`) already provides server-side protection on initial navigation. `AuthGuard` adds a client-side defense-in-depth layer that catches session expiration during an active session — a scenario the middleware cannot catch after the first page load.

Closes #493

## Test plan
- [ ] Log out, navigate to `/channels/...` — should redirect to `/auth/login?returnUrl=...`
- [ ] Log out, navigate to `/settings/...` — should redirect to `/auth/login?returnUrl=...`
- [ ] After login via the redirected URL, verify the `returnUrl` lands you back on the original page
- [ ] Verify authenticated users are unaffected (no redirect occurs)

### DIFF
--- a/harmony-frontend/src/components/layout/AuthGuard.tsx
+++ b/harmony-frontend/src/components/layout/AuthGuard.tsx
@@ -1,0 +1,27 @@
+'use client';
+
+import { useEffect } from 'react';
+import { useRouter, usePathname } from 'next/navigation';
+import { useAuth } from '@/hooks/useAuth';
+
+/**
+ * Client-side authentication guard for protected route layouts.
+ *
+ * Defense-in-depth layer that catches session expiration during an active
+ * session — the Next.js middleware already handles the server-side check on
+ * initial navigation. Redirects unauthenticated users to /auth/login with a
+ * returnUrl so they land back where they were after signing in.
+ */
+export function AuthGuard() {
+  const { isAuthenticated, isLoading } = useAuth();
+  const router = useRouter();
+  const pathname = usePathname();
+
+  useEffect(() => {
+    if (!isLoading && !isAuthenticated) {
+      router.replace(`/auth/login?returnUrl=${encodeURIComponent(pathname)}`);
+    }
+  }, [isAuthenticated, isLoading, pathname, router]);
+
+  return null;
+}
--- a/harmony-frontend/src/components/layout/AuthGuard.tsx
+++ b/harmony-frontend/src/components/layout/AuthGuard.tsx
@@ -1,8 +1,9 @@
 'use client';
 
 import { useEffect } from 'react';
-import { useRouter, usePathname } from 'next/navigation';
+import { useRouter, usePathname, useSearchParams } from 'next/navigation';
 import { useAuth } from '@/hooks/useAuth';
+import type { ReactNode } from 'react';
 
 /**
  * Client-side authentication guard for protected route layouts.
@@ -11,17 +12,25 @@ import { useAuth } from '@/hooks/useAuth';
  * session — the Next.js middleware already handles the server-side check on
  * initial navigation. Redirects unauthenticated users to /auth/login with a
  * returnUrl so they land back where they were after signing in.
+ *
+ * Suppresses children during the loading window to prevent a flash of
+ * protected content when middleware misses an invalid cookie.
  */
-export function AuthGuard() {
+export function AuthGuard({ children }: { children: ReactNode }) {
   const { isAuthenticated, isLoading } = useAuth();
   const router = useRouter();
   const pathname = usePathname();
+  const searchParams = useSearchParams();
 
   useEffect(() => {
     if (!isLoading && !isAuthenticated) {
-      router.replace(`/auth/login?returnUrl=${encodeURIComponent(pathname)}`);
+      const fullPath = searchParams.toString()
+        ? `${pathname}?${searchParams.toString()}`
+        : pathname;
+      router.replace(`/auth/login?returnUrl=${encodeURIComponent(fullPath)}`);
     }
-  }, [isAuthenticated, isLoading, pathname, router]);
+  }, [isAuthenticated, isLoading, pathname, router, searchParams]);
 
-  return null;
+  if (isLoading || !isAuthenticated) return null;
+  return <>{children}</>;
 }
--- a/harmony-frontend/src/__tests__/AuthGuard.test.tsx
+++ b/harmony-frontend/src/__tests__/AuthGuard.test.tsx
@@ -1,0 +1,82 @@
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+const mockReplace = jest.fn();
+const mockUseAuth = jest.fn();
+let mockSearchParams = new URLSearchParams('');
+
+jest.mock('@/hooks/useAuth', () => ({
+  useAuth: () => mockUseAuth(),
+}));
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({ replace: mockReplace }),
+  usePathname: () => '/channels/server-1/general',
+  useSearchParams: () => mockSearchParams,
+}));
+
+// Import after mocks are set up
+import { AuthGuard } from '@/components/layout/AuthGuard';
+
+describe('AuthGuard', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockSearchParams = new URLSearchParams('');
+  });
+
+  it('suppresses children while loading', () => {
+    mockUseAuth.mockReturnValue({ isAuthenticated: false, isLoading: true });
+
+    render(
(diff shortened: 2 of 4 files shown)


harmony-frontend/src/components/layout/AuthGuard.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const pathname = usePathname();
+
+  useEffect(() => {
+    if (!isLoading && !isAuthenticated) {
+      router.replace(`/auth/login?returnUrl=${encodeURIComponent(pathname)}`);
+    }
+  }, [isAuthenticated, isLoading, pathname, router]);
+
+  return null;
+}
```
AvanishKulkarni, line 36: `AuthGuard` always returns `null`, so children in the parent layout render unconditionally during the `isLoading` window. Convert this to a wrapper component so the guard can conditionally suppress protected content:

```tsx
export function AuthGuard({ children }: { children: React.ReactNode }) {
  const { isAuthenticated, isLoading } = useAuth();
  const router = useRouter();
  const pathname = usePathname();

  useEffect(() => {
    if (!isLoading && !isAuthenticated) {
      router.replace(`/auth/login?returnUrl=${encodeURIComponent(pathname)}`);
    }
  }, [isAuthenticated, isLoading, pathname, router]);

  if (isLoading || !isAuthenticated) return null;
  return <>{children}</>;
}
```

Then update both layouts to `<AuthGuard>{children}</AuthGuard>`.

AvanishKulkarni (dismissed):
## Review

The intent here is correct — a client-side defense-in-depth layer for session expiration is the right call given the middleware only runs on initial navigation. The implementation is clean and focused, but there is one meaningful architectural issue and a couple of smaller concerns.

**Main issue: children always render during loading**

`AuthGuard` returns `null` and is mounted as a sibling to `{children}` in both layouts:

```tsx
<>
  <AuthGuard />   // always null
  {children}       // always renders
</>
```

This means the protected content flashes briefly during the `isLoading` window (while `getCurrentUser()` is in flight). For the intended session-expiration use case the user is already on the page so this does not matter, but for any edge case where the middleware misses an invalid cookie the content is momentarily visible before the redirect fires. The standard pattern is to make `AuthGuard` a wrapper that conditionally renders its children:

```tsx
// AuthGuard.tsx
export function AuthGuard({ children }: { children: ReactNode }) {
  // …same useEffect for the redirect…
  if (isLoading || !isAuthenticated) return null;
  return <>{children}</>;
}

// layout.tsx
<AuthGuard>{children}</AuthGuard>
```

**Secondary: `returnUrl` loses the query string**

`usePathname()` returns only the pathname segment — query parameters are dropped. A user at `/channels/foo/bar?highlight=123` would be redirected to `/auth/login?returnUrl=%2Fchannels%2Ffoo%2Fbar` and land back without the `?highlight=123`. Combine `pathname` with `useSearchParams()` to preserve the full URL:

```tsx
const searchParams = useSearchParams();
const fullPath = searchParams.toString()
  ? `${pathname}?${searchParams.toString()}`
  : pathname;
router.replace(`/auth/login?returnUrl=${encodeURIComponent(fullPath)}`);
```

**Minor: no tests for the new component**

The `AuthGuard` has three distinct branches (`isLoading`, authenticated, unauthenticated) that are straightforward to unit-test with a mocked `useAuth`. The project has a frontend test suite in `harmony-frontend/src/__tests__/` — a small test file here would close the coverage gap introduced by this PR.

harmony-frontend/src/components/layout/AuthGuard.tsx
declanblanc: Fixed in 5933ca23. Converted AuthGuard to a wrapper component that accepts children — layouts now use `<AuthGuard>{children}</AuthGuard>`. Children are suppressed (returns null) during isLoading or when unauthenticated, preventing any flash of protected content.

declanblanc:
> Re: comment 4197231770

Fixed in 5933ca23. All three issues addressed: (1) AuthGuard is now a wrapper that suppresses children when isLoading or unauthenticated — both layouts updated to `<AuthGuard>{children}</AuthGuard>`; (2) returnUrl now preserves query params via useSearchParams, so `?highlight=123` survives the login redirect; (3) added unit tests covering all three branches (loading, authenticated, unauthenticated) plus the query-param preservation case.

acabrera04 (changes requested):
## 8-Point Checklist
- [x] Prioritize Logic over Style: Core logic is focused and avoids style-only churn; the wrapper guard pattern is correct.
- [x] Security First: Redirect logic does not expose secrets and preserves intended auth boundary for protected routes.
- [x] Architectural Alignment: The change is scoped to frontend auth guard behavior and matches the route-layout protection approach used in this codebase.
- [x] Issue Completion: The PR addresses the reported session-expiration guard gap and includes branch coverage tests for new logic.
- [x] No Nitpicking: Feedback below is only about a concrete build/runtime blocker.
- [x] Avoid Repetition: Prior feedback (content flash/query param/test coverage) was addressed at this head; not repeated.
- [x] Iterative Reviews: Reviewed latest head commit `5933ca235a33aa080bb3d0aefe712d6627e0e9ae` and validated against prior unresolved concerns.
- [ ] Prevent CI Failures: Current head still introduces a failing frontend build in CI.

Bundled review:
The remaining blocker is a real CI break introduced by this revision: `next build` fails while prerendering `/settings` with `useSearchParams() should be wrapped in a suspense boundary at page "/settings"`. This comes from `AuthGuard` calling `useSearchParams()` without a surrounding `Suspense` boundary in the render tree.

Please fix by either:
1. Wrapping `AuthGuard` usage in a `Suspense` boundary in the affected layouts/pages, or
2. Refactoring the guard so it does not require `useSearchParams` during prerender-sensitive paths (for example, reading query only after mount in a way compatible with the app router prerender constraints).

Evidence: GitHub Actions run `25111893250` -> job `Frontend Lint and Build` fails during `next build` on `/settings` with the missing suspense error.

declanblanc:
> Re: comment 4198262727

Fixed in 37a67b15. Wrapped `<AuthGuard>` in `<Suspense>` in both `/channels/layout.tsx` and `/settings/layout.tsx`. This satisfies Next.js App Router's requirement that any component calling `useSearchParams()` must be wrapped in a Suspense boundary, resolving the static prerender failure on `/settings`.

declanblanc:
> Re: comment 4197231770

All three concerns addressed in this and prior commits: (1) wrapper pattern — AuthGuard now conditionally renders children (returns null during loading/unauthenticated); (2) query string preservation — useSearchParams() is used to build the full returnUrl; (3) tests — AuthGuard.test.tsx added with 4 test cases covering isLoading, authenticated, unauthenticated, and query param preservation branches. The Suspense boundary required by Next.js for useSearchParams() was added in 37a67b15.

acabrera04 (approved):
## 8-Point Checklist
- [x] Prioritize Logic over Style: The latest changes remain focused on auth flow correctness and render behavior.
- [x] Security First: Guard still enforces unauthenticated redirect with encoded return URL and no new sensitive-data exposure.
- [x] Architectural Alignment: Client-side defense-in-depth guard in route layouts matches intended frontend auth pattern.
- [x] Issue Completion: Latest head addresses the previously blocking Next.js prerender/Suspense requirement and keeps prior fixes intact.
- [x] No Nitpicking: No style-only concerns raised; only functional/blocking checks considered.
- [x] Avoid Repetition: Prior requested-change items were revalidated and are not repeated.
- [x] Iterative Reviews: Re-reviewed at latest head commit `37a67b1562f0c62e3c205c2d8c1a54b12818a5b2` and compared against prior blockers.
- [x] Prevent CI Failures: Frontend/backend lint/tests and integration checks are passing at current head.

Bundled review:
Re-review complete on latest head. The previously reported blocker (`useSearchParams()` suspense requirement during prerender) is resolved by wrapping `AuthGuard` with `Suspense` in both `/channels` and `/settings` layouts, and CI is green on this revision. I do not see any new unresolved issues.

PR #545, opened 2026-04-29
fix: messages no longer persist after page refresh

declanblanc:
## Summary

- Changed message query sort order from `asc` to `desc` in the backend so messages are returned newest-first directly from the DB, eliminating the need for a client-side reverse
- Converted fire-and-forget cache invalidation calls to `await`-based with try/catch in both `sendMessage` and `sendReply` — previously the cache was invalidated asynchronously, so a page refresh could race ahead of invalidation and serve stale data
- Removed the now-unnecessary `.reverse()` and its stale comment from the frontend `getMessages` service

## Test plan

- [ ] Send a message in a channel, refresh the page — new message should persist
- [ ] Send a reply in a thread, refresh the page — reply should persist
- [ ] Verify message order is correct (newest at bottom) after refresh
- [ ] Confirm no regression in pagination / `hasMore` behavior

### DIFF
--- a/harmony-backend/src/services/message.service.ts
+++ b/harmony-backend/src/services/message.service.ts
@@ -127,7 +127,7 @@ export const messageService = {
           { channelId, isDeleted: false },
           clampedLimit + 1,
           cursor,
-          { createdAt: 'asc' },
+          { createdAt: 'desc' },
         );
 
         const hasMore = messages.length > clampedLimit;
@@ -161,16 +161,13 @@ export const messageService = {
         }),
     });
 
-    cacheService
-      .invalidatePattern(
+    try {
+      await cacheService.invalidatePattern(
         `channel:msgs:${sanitizeKeySegment(serverId)}:${sanitizeKeySegment(channelId)}:*`,
-      )
-      .catch((err) =>
-        logger.warn(
-          { err, channelId, serverId },
-          'Failed to invalidate channel message cache after send',
-        ),
       );
+    } catch (err) {
+      logger.warn({ err, channelId, serverId }, 'Failed to invalidate channel message cache after send');
+    }
 
     eventBus
       .publish(EventChannels.MESSAGE_CREATED, {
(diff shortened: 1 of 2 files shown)


acabrera04 (approved):
- [x] **Prioritize Logic over Style:** The changes are logic-focused and remove an ordering/caching race without adding stylistic churn; query order and cache invalidation flow are coherent with intended behavior.
- [x] **Security First:** No secrets, auth, or trust-boundary regressions introduced; inputs used in cache keys remain sanitized and error handling stays non-fatal.
- [x] **Architectural Alignment:** Changes stay within existing service/repository boundaries and follow current backend message retrieval + cache invalidation architecture.
- [x] **Issue Completion:** PR summary scope (message persistence after refresh + ordering consistency) is implemented in both backend and frontend paths touched by this bug.
- [x] **No Nitpicking:** Review focuses only on correctness/regression risk; no style-only feedback.
- [x] **Avoid Repetition:** There are no prior substantive review threads from me or others to duplicate; this review adds only net-new assessment.
- [x] **Iterative Reviews:** Verified latest PR head commit `3b6bd2f63bccd01c0ac05e6bf676c97e499d9ad2`; no unresolved prior review concerns to re-litigate.
- [x] **Prevent CI Failures:** Checked latest status checks on this head; required lint/build/tests are passing and the patch introduces no obvious syntax/import/runtime breakage.

Bundled review:

Reviewed against latest head `3b6bd2f63bccd01c0ac05e6bf676c97e499d9ad2`.

The fix correctly addresses the persistence race by awaiting cache invalidation in `sendMessage` and `createReply` while preserving non-blocking behavior via localized `try/catch` logging. The backend message query order change to `createdAt DESC` aligns the first page with newest data, and removing the frontend service-level `.reverse()` prevents double-order inversion with current channel rendering logic. CI on the current head is green for backend/frontend lint/build, backend/frontend tests, and integration tests.

APPROVE.

PR #553, opened 2026-04-29
Add one-command local dev startup

acabrera04:
## Summary
- update `dev.sh` to start Docker services, run `prisma migrate deploy`, and launch backend API + backend worker + frontend together
- add root `npm run dev` that invokes `./dev.sh`
- document one-command local startup in the root README
- make `dev.sh` process monitoring macOS-compatible by replacing `wait -n` with a portable PID-monitor loop

## Verification
- `bash -n dev.sh`
- `git diff --stat`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,16 @@ password: HarmonyAdmin123!
 
 ### 3. Start the app for normal local development
 
+One-command startup from repo root:
+
+```bash
+./dev.sh
+```
+
+This script starts Docker (`postgres`, `redis`), applies backend Prisma migrations, then launches backend API, backend worker, and frontend together with shared shutdown on `Ctrl+C`.
+
+Manual startup (three terminals):
+
 Use three terminals:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -75,10 +75,10 @@ password: HarmonyAdmin123!
 One-command startup from repo root:
 
 ```bash
-./dev.sh
+npm run dev
 ```
 
-This script starts Docker (`postgres`, `redis`), applies backend Prisma migrations, then launches backend API, backend worker, and frontend together with shared shutdown on `Ctrl+C`.
+This command runs `./dev.sh`, which starts Docker (`postgres`, `redis`), applies backend Prisma migrations, then launches backend API, backend worker, and frontend together with shared shutdown on `Ctrl+C`.
 
 Manual startup (three terminals):
 
--- a/dev.sh
+++ b/dev.sh
@@ -1,51 +1,72 @@
 #!/usr/bin/env bash
-# dev.sh — start Docker services, backend, and frontend for local development
+# dev.sh — start Docker services, backend API + worker, and frontend for local development
 set -euo pipefail
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 BACKEND_DIR="$REPO_ROOT/harmony-backend"
 FRONTEND_DIR="$REPO_ROOT/harmony-frontend"
 
-# ── Docker ────────────────────────────────────────────────────────────────────
+require_cmd() {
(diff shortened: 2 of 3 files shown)


declanblanc (commented):
## PR Review Checklist

- [x] **Prioritize Logic over Style** — Logic is sound; the polling loop correctly replaces `wait -n` and handles all three PIDs.
- [x] **Security First** — No security concerns; script only manages local processes and runs pre-existing npm/docker commands.
- [x] **Architectural Alignment** — Tooling-only change; no backend/frontend architecture involved.
- [x] **Issue Completion** — No linked issue; PR description clearly states what was done and includes manual verification output.
- [x] **No Nitpicking** — Review focused on logic and correctness only.
- [x] **Prevent CI Failures** — No CI-impacting changes; `package.json` addition is additive and correct.

---

## Review

**Overview:** This PR adds a `dev.sh` wrapper that starts the backend worker alongside the API and frontend, replacing the old two-process script. It also adds a root-level `npm run dev` convenience command and updates the README. The core change is the worker process addition and the replacement of `wait -n` with a polling loop.

### `dev.sh`

**Positive — `wait -n` replacement is a real fix.** The old code used `wait -n "$BACKEND_PID" "$FRONTEND_PID"`, which requires Bash 4.3+. macOS ships Bash 3.2 by default, meaning the old script silently broke on stock macOS. The new polling loop is compatible with Bash 3.2+ and is correct.

**Minor: polling loop has a 1-second detection lag.** If a service crashes, the script waits up to 1 second before detecting it. For a dev tool this is entirely acceptable, but worth documenting if the intent was to match the responsiveness of `wait -n`.

**Concern: `require_cmd` does not check if Docker daemon is running**, only that the `docker` binary exists. A user with Docker installed but daemon stopped will get a useful error from `docker compose up -d` itself, so this is not broken — but a `docker info >/dev/null 2>&1 || { echo "Docker daemon is not running"; exit 1; }` guard would give a cleaner message. Non-blocking.

**Concern: no `cd "$BACKEND_DIR"` guard before `npm run dev:worker`.** The working directory is `$BACKEND_DIR` when the worker is launched (set earlier for migrations), and this works correctly today. However, it is implicit — if the script is refactored in the future and a `cd` is added between the API start and worker start, the worker would silently run from the wrong directory. Explicitly `cd "$BACKEND_DIR"` before both `npm run dev &` and `npm run dev:worker &` would make the intent clear and resilient.

```diff
+echo "==> Starting backend API (npm run dev)..."
+cd "$BACKEND_DIR"
 npm run dev &
 BACKEND_PID=$!
 
 echo "==> Starting backend worker (npm run dev:worker)..."
+cd "$BACKEND_DIR"
 npm run dev:worker &
 WORKER_PID=$!
```

### `README.md`

The addition is clear. The flow — one-command option first, manual three-terminal option second — matches how most developers will use it.

### `package.json`

`"dev": "./dev.sh"` is correct and idiomatic. No issues.

---

**Summary:** The logic is correct and the `wait -n` fix is a genuine improvement for macOS compatibility. The only non-trivial suggestion is adding explicit `cd "$BACKEND_DIR"` guards before each backend process launch to make directory assumptions explicit rather than implicit.

declanblanc (approved):
The code is functionally correct and the `wait -n` fix is a genuine macOS compatibility improvement. The suggestions in my previous comment (explicit `cd "$BACKEND_DIR"` guards before each backend process launch) are non-blocking — the script works correctly today. Approving; the `cd` guards are worth addressing before merging but are not a blocker.

acabrera04:
> Re: comment 4200761951

Addressed in cf95f68. Added explicit `cd ""` guards before both backend API and worker launches, and updated README to use `npm run dev` as the primary one-command path.

acabrera04:
> Re: comment 4200772874

Thanks for the approval. Followed up in cf95f68 with explicit backend directory guards before each backend process launch for resilience during future refactors.

PR #560, opened 2026-04-29
fix(settings): eliminate React key warning in PermissionsSection

### LINKED ISSUE #556: Settings permissions table triggers React key warning in PermissionsSection
## Summary
The Server Settings page throws a React warning in development:

`Each child in a list should have a unique "key" prop.`

The warning points to:
- `harmony-frontend/src/components/settings/PermissionsSection.tsx:152:17`
- render path in `PermissionsSection` during `ACTION_GROUPS.map(...)`

## Reproduction
1. Run the frontend in dev mode.
2. Open Server Settings.
3. Observe the Next.js/React error overlay.

## Actual Behavior
React warns that list children are missing unique keys.

## Expected Behavior
No key warning should appear; each top-level element returned by list rendering should have a stable unique key.

## Suspected Cause
Inside `ACTION_GROUPS.map(group => (...))`, the top-level fragment `<>...</>` has no key.

Current pattern:
- `ACTION_GROUPS.map(group => (<> ... </>))`

Likely fix:
- use `<React.Fragment key={...}>` (or another keyed wrapper) around each group block.

## Impact
- Dev console/overlay noise
- Makes it harder to spot other real runtime issues

declanblanc:
## Summary
- Replaces keyless shorthand fragments (`<>`) with `<React.Fragment key={group.label}>` in `ACTION_GROUPS.map()`
- Adds `import React from 'react'` so `React.Fragment` is in scope
- Removes the redundant `key` prop from the inner `<tr>` (the key is now on the Fragment wrapper, which is the correct level)

Closes #556

## Test plan
- [ ] Open Server Settings in dev mode
- [ ] Confirm no React key warning appears in the console/overlay
- [ ] Verify permissions table renders correctly for all role columns

### DIFF
--- a/harmony-frontend/src/components/settings/PermissionsSection.tsx
+++ b/harmony-frontend/src/components/settings/PermissionsSection.tsx
@@ -4,6 +4,8 @@
  * always reflects the authoritative backend definition.
  */
 
+import React from 'react';
+
 // ─── Types ────────────────────────────────────────────────────────────────────
 
 export type PermissionMatrix = Record<string, string[]>;
@@ -149,8 +151,8 @@ export function PermissionsSection({ matrix }: PermissionsSectionProps) {
             </thead>
             <tbody>
               {ACTION_GROUPS.map(group => (
-                <>
-                  <tr key={`${group.label}-header`} className='border-t border-white/10 bg-[#25262a]'>
+                <React.Fragment key={group.label}>
+                  <tr className='border-t border-white/10 bg-[#25262a]'>
                     <td
                       colSpan={ROLE_ORDER.length + 1}
                       className='px-4 py-1.5 text-xs font-semibold uppercase tracking-wide text-gray-500'
@@ -182,7 +184,7 @@ export function PermissionsSection({ matrix }: PermissionsSectionProps) {
                       })}
                     </tr>
                   ))}
-                </>
+                </React.Fragment>
               ))}
             </tbody>
           </table>


harmony-frontend/src/components/settings/PermissionsSection.tsx
```diff
@@ -149,8 +151,8 @@ export function PermissionsSection({ matrix }: PermissionsSectionProps) {
             </thead>
             <tbody>
               {ACTION_GROUPS.map(group => (
-                <>
-                  <tr key={`${group.label}-header`} className='border-t border-white/10 bg-[#25262a]'>
+                <React.Fragment key={group.label}>
```
AvanishKulkarni, line 154: `group.label` ('Server', 'Channels', 'Messages', 'Settings') is hardcoded, unique, and stable — a solid key choice. The original ``${group.label}-header`` suffix on the removed `<tr>` was also redundant once the key moved to the Fragment wrapper, so dropping it is correct.

AvanishKulkarni (approved):
## PR Review Checklist

- [x] **Logic over Style** — The fix correctly moves the `key` prop from the shorthand fragment (which cannot accept props) to `<React.Fragment key={...}>`. Removing the now-redundant key from the inner `<tr>` is also correct.
- [x] **Security First** — No security concerns; pure rendering fix.
- [x] **Architectural Alignment** — N/A for this UI-only component change.
- [x] **Issue Completion** — Issue #556 describes exactly this symptom: `<>...</>` with no key inside `ACTION_GROUPS.map()`. The PR eliminates the warning at the correct level.
- [x] **No Nitpicking** — Review focuses only on correctness.
- [x] **Prevent CI Failures** — `import React from 'react'` is required to reference `React.Fragment` explicitly; will not break the build.

---

## Review

Correct, minimal fix. The root cause was that shorthand fragments (`<>`) cannot accept props, so React never saw a key at the list-iteration level — it only inspects the outermost returned element per map callback. Promoting the key to `<React.Fragment key={group.label}>` is the standard remedy for this pattern.

`group.label` ('Server', 'Channels', 'Messages', 'Settings') is hardcoded, unique, and stable — a good key value. The original ``${group.label}-header`` suffix on the removed `<tr>` key was redundant once the key moved up to the Fragment wrapper.

PR #562, opened 2026-04-29
fix: prevent page not found when generating invite links (#547)

### LINKED ISSUE #547: fix: prevent page not found errors when navigating to invite links
## Problem

Users encounter a 404 / page-not-found error after generating an invite link and navigating to it. The invite link is created successfully but the route either does not exist or is not handled correctly, leaving the user on a dead-end error page.

## Expected Behavior

Navigating to a generated invite link should route the user to the correct page (e.g., a join/accept-invite screen) without errors.

## Steps to Reproduce

1. Generate an invite link for a server/channel.
2. Copy and open the invite link in a browser.
3. Observe a 404 / page-not-found error.

## Acceptance Criteria

- [ ] All valid invite link URLs resolve to the correct page without a 404.
- [ ] Invalid or expired invite links display a meaningful error message rather than a generic 404.
- [ ] Frontend route(s) for invite links are properly defined and handle the invite token parameter.
- [ ] Backend invite-link endpoint validates the token and returns an appropriate error response for expired/invalid tokens.
- [ ] Tests cover the happy path (valid invite) and error path (invalid/expired invite).

## Notes

- Investigate whether the issue is a missing frontend route, a missing/broken API endpoint, or a combination of both.
- Ensure the fix works for both direct navigation (copy-paste URL) and redirect flows after login.

AvanishKulkarni:
## Summary

- `requireServerSettingsAccess` was calling `notFound()` even in `'throw'` mode (used by all server actions like `generateInviteAction`). In Next.js App Router, `notFound()` inside a server action propagates as a page-level 404 response that bypasses the client `try/catch` — causing the settings page to show 404 instead of an inline error. Fixed by throwing a plain `Error('Server not found')` in `'throw'` mode.
- The login page's "Register" link did not forward `returnUrl`, so new users following an invite link who clicked "Register" were dropped at `/channels` after signup. Fixed by propagating `returnUrl` to `/auth/signup?returnUrl=…`.
- 2 new test cases added to `server-settings-access.test.ts`; all 386 frontend tests pass.

## Test plan

- [ ] Navigate to server settings → Invites → click "Generate Invite Link" — no 404, invite appears inline
- [ ] Open invite link while logged out → redirected to login → click "Register" → after signup, lands back on `/invite/[code]`
- [ ] Open invite link while logged out → log in → lands back on `/invite/[code]`
- [ ] `npm test` in `harmony-frontend/` passes (386 tests)

Closes #547

### DIFF
--- a/harmony-frontend/src/__tests__/server-settings-access.test.ts
+++ b/harmony-frontend/src/__tests__/server-settings-access.test.ts
@@ -147,9 +147,19 @@ describe('requireServerSettingsAccess', () => {
     expect(mockRedirect).not.toHaveBeenCalled();
   });
 
-  it('delegates missing servers to notFound', async () => {
+  it('delegates missing servers to notFound in redirect mode', async () => {
     mockGetServerAuthenticated.mockResolvedValue(null);
 
     await expect(requireServerSettingsAccess('missing-server')).rejects.toThrow('NOT_FOUND');
+    expect(mockNotFound).toHaveBeenCalled();
+  });
+
+  it('throws a plain error for missing servers in throw mode (no notFound)', async () => {
+    mockGetServerAuthenticated.mockResolvedValue(null);
+
+    await expect(requireServerSettingsAccess('missing-server', 'throw')).rejects.toThrow(
+      'Server not found',
+    );
+    expect(mockNotFound).not.toHaveBeenCalled();
   });
 });
--- a/harmony-frontend/src/app/auth/login/page.tsx
+++ b/harmony-frontend/src/app/auth/login/page.tsx
@@ -14,6 +14,7 @@ function LoginForm() {
   const { login } = useAuth();
   const router = useRouter();
   const searchParams = useSearchParams();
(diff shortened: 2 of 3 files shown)


declanblanc (approved):
- [x] 1. Prioritize logic over style
- [x] 2. Security first
- [x] 3. Architectural alignment
- [x] 4. Issue completion
- [x] 5. No nitpicking
- [x] 6. Avoid repetition
- [ ] 8. Prevent CI failures — not executed locally here

**Approve**

### Findings (severity)

- **None blocking.** `requireServerSettingsAccess` avoiding `notFound()` in `'throw'` mode is the correct Next.js App Router behavior for server actions whose callers need inline errors instead of a page-level 404.

### Minor / follow-ups (non-blocking)

- **Low — Register link `returnUrl` handling**
  - Restricting propagation to same-origin-style paths (`/` and not `//`) is appropriate for reducing open-redirect risk while preserving invite flows.

### Tests

- Updated `server-settings-access.test.ts` correctly distinguishes redirect vs throw behavior for missing servers.

Solid fix for #547.

PR #566, opened 2026-04-29
fix(voice): clear stale participant entry after failed voice join

### LINKED ISSUE #492: bug: Cannot properly disconnect from voice channel on failed connection
## Summary
When joining a voice channel fails (microphone denied, network error), cannot cleanly disconnect. Voice UI shows a stuck state.

## Expected Behavior
- Failed connection attempts can be retried
- User can click Leave to return to normal state
- Backend voice state is cleaned up on failure
- No stale participant entries in channel

AvanishKulkarni:
## Summary

Fixes #492 — when a voice channel join fails after the backend `voice.join` tRPC call succeeds (e.g. microphone access denied, Twilio network error), the sidebar was left showing a stale participant entry for the local user.

**Root cause:** `localParticipantIdentityRef.current` was only assigned from `room.localParticipant.identity` _after_ a successful `TwilioVideo.connect()`. In the failure path, Twilio never connected so the ref stayed `null`, causing the cleanup in `leaveChannel()` to skip the `channelParticipants` filter — leaving a ghost entry in the sidebar.

**Fix:** Add `currentUserId` prop to `VoiceProvider` and set `localParticipantIdentityRef.current = currentUserId` immediately after token validation (before `TwilioVideo.connect`). The Twilio token already uses `userId` as its identity (see `voice.service.ts:152`), so the value is authoritative even before the room connects.

## Changes

- **`VoiceContext.tsx`** — add `currentUserId?: string` to `VoiceProviderProps`; set `localParticipantIdentityRef.current` from it after `voice.join` validation, before the `TwilioVideo.connect` call
- **`HarmonyShell.tsx`** — pass `currentUserId={authUser?.id}` to `VoiceProvider`

## Test plan

- [ ] All 397 frontend tests pass (`npm test` in `harmony-frontend`)
- [ ] Backend voice service tests pass (`voice.service.test.ts`)
- Manual: deny microphone permission in browser → click a voice channel → confirm error toast appears, sidebar shows no stale participant entry, join button re-enables for retry

### DIFF
--- a/harmony-frontend/src/components/layout/HarmonyShell.tsx
+++ b/harmony-frontend/src/components/layout/HarmonyShell.tsx
@@ -436,7 +436,7 @@ export function HarmonyShell({
   }, [isChannelLocked]);
 
   return (
-    <VoiceProvider serverId={currentServer.id} voiceChannelIds={voiceChannelIds}>
+    <VoiceProvider serverId={currentServer.id} voiceChannelIds={voiceChannelIds} currentUserId={authUser?.id}>
       <div className='flex h-screen overflow-hidden bg-[#202225] font-sans'>
         {/* Skip-to-content: visually hidden, appears on keyboard focus (WCAG 2.4.1) */}
         <a
--- a/harmony-frontend/src/contexts/VoiceContext.tsx
+++ b/harmony-frontend/src/contexts/VoiceContext.tsx
@@ -101,9 +101,15 @@ interface VoiceProviderProps {
   serverId: string;
   /** IDs of all voice channels in the current server. */
   voiceChannelIds: string[];
+  /**
+   * The authenticated user's ID. Used to clean up channelParticipants when
+   * a join fails before a Twilio room is established — at that point
+   * room.localParticipant.identity is not yet available.
+   */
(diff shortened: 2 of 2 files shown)


declanblanc (approved):
- [x] 1. Prioritize logic over style
- [x] 2. Security first
- [x] 3. Architectural alignment
- [x] 4. Issue completion
- [x] 5. No nitpicking
- [x] 6. Avoid repetition
- [ ] 8. Prevent CI failures — not executed locally here

**Approve**

### Findings (severity)

- **Low — No automated regression test for the ghost-participant path**
  - The fix is logically sound: `localParticipantIdentityRef` is seeded from `currentUserId` before `TwilioVideo.connect`, matching stated token identity, so `leaveChannel()` can remove the local row when connect fails.
  - Follow-up (non-blocking): a focused unit/integration test that mocks `voice.join` success + Twilio connect rejection would lock this in.

### Notes / assumptions

- Assumed Twilio token identity always equals app `userId` (as referenced). Post-connect still overwrites the ref from `room.localParticipant.identity`, which is appropriate.

Nice fix for the failure-path cleanup gap.

PR #571, opened 2026-04-29
fix: real-time profile updates (closes #491)

### LINKED ISSUE #491: bug: User display names and profile info do not update in real-time
## Summary
When a user updates their profile, other connected users don't see the change until they refresh.

## Expected Behavior
- Display name changes appear immediately in all UI
- Avatar changes appear immediately in all UI
- Status changes appear immediately in user indicators

## Steps to Reproduce
1. User A changes display name in settings
2. User B observes - display name doesn't update in sidebar/messages
3. Refresh page - change is now visible

AvanishKulkarni:
## Summary

Fixes #491 — display name, avatar, and `publicProfile` changes now propagate to all connected members in real time without requiring a page refresh.

- **Backend**: Added `USER_PROFILE_UPDATED` Redis pub/sub event channel and payload type. `user.service.ts` emits the event per server membership whenever `displayName`, `avatarUrl`, or `publicProfile` changes (status-only updates are not affected). The SSE router subscribes, hydrates from the DB to apply the `publicProfile` privacy gate (same pattern as `member:joined`), and emits a `member:profileUpdated` SSE event.
- **Frontend**: `useServerEvents` gains `onMemberProfileUpdated` option. `HarmonyShell` wires it to patch both `localMembers` (sidebar) and `localMessages` (message history + reply previews) so all in-flight UI updates immediately — no refresh needed.

## Test plan

- [ ] User A updates display name in settings → User B's sidebar and message history update immediately
- [ ] User A clears avatar → User B sees the fallback avatar immediately
- [ ] User A sets `publicProfile=false` → User B sees "Anonymous" / no avatar immediately
- [ ] Status-only update does NOT emit a `member:profileUpdated` event
- [ ] All backend user service tests pass (`npm test --testPathPattern=user.service`)
- [ ] All frontend tests pass (`cd harmony-frontend && npm test -- --watchAll=false`)

### DIFF
--- a/harmony-backend/src/events/eventTypes.ts
+++ b/harmony-backend/src/events/eventTypes.ts
@@ -21,6 +21,7 @@ export const EventChannels = {
   CHANNEL_DELETED: 'harmony:CHANNEL_DELETED',
   SERVER_UPDATED: 'harmony:SERVER_UPDATED',
   USER_STATUS_CHANGED: 'harmony:USER_STATUS_CHANGED',
+  USER_PROFILE_UPDATED: 'harmony:USER_PROFILE_UPDATED',
   REACTION_ADDED: 'harmony:REACTION_ADDED',
   REACTION_REMOVED: 'harmony:REACTION_REMOVED',
   USER_MENTIONED: 'harmony:USER_MENTIONED',
@@ -135,6 +136,12 @@ export interface UserStatusChangedPayload {
   status: 'ONLINE' | 'IDLE' | 'DND' | 'OFFLINE';
 }
 
+export interface UserProfileUpdatedPayload {
+  userId: string;
+  serverId: string;
+  timestamp: string;
+}
+
 export interface ReactionAddedPayload {
   messageId: string;
   channelId: string;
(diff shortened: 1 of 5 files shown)


declanblanc (changes requested):
- [x] 1. Prioritize logic over style — Focused on correctness of realtime propagation.
- [x] 2. Security first — SSE hydration respects `publicProfile`; no new exposure noted.
- [x] 3. Architectural alignment — Redis fan-out + SSE hydrate matches existing member patterns.
- [x] 4. Issue completion — Partially; see blocking finding on reply previews.
- [x] 5. No nitpicking
- [x] 6. Avoid repetition
- [ ] 8. Prevent CI failures — Not run locally.

**Request changes**

### Findings (severity)

- **High — Reply previews can stay stale when the updated user is only the parent author**
  - In `HarmonyShell`, `handleMemberProfileUpdated` updates `localMessages` only when `msg.author.id === id` and returns early otherwise.
  - For a reply authored by user B quoting user A, when **A** updates display name/avatar/public profile, `msg.author.id` is B, so the row is skipped and **`parentMessage.author` for A is never patched** — contradicting the stated goal (real-time reply previews) and issue #491 scope.
  - Fix direction: also match `msg.parentMessage?.author.id === id` and merge profile fields into `parentMessage.author` even when the main `msg.author` is a different user.

### Non-blocking

- **Low — Automated coverage:** No tests shown in the diff for the new event path or frontend patching; consider at least a unit test for the message-map reducer edge case above.

Backend event wiring and privacy hydration in `events.router.ts` look sound once the client handles all message shapes.

AvanishKulkarni:
> Re: comment 4201492484

Fixed in 40db18c. The early-return on msg.author.id !== id was removed; now we check both isMainAuthor and isParentAuthor before deciding what to patch, so parentMessage.author is correctly updated even when the updated user authored only the parent message.

declanblanc (approved):
The fix in 3c09b95 correctly addresses the high-severity finding. `handleMemberProfileUpdated` now checks both `isMainAuthor` and `isParentAuthor` and patches `parentMessage.author` even when the updated user only authored the parent message. All 425 frontend tests and 888 backend tests pass.

PR #593, opened 2026-04-30
fix: include attachments in public channel message API responses

### LINKED ISSUE #588: Images and attachments not visible in guest view
### Description

Images and other file attachments are not visible when viewing the app as a guest (unauthenticated user). Content that should be publicly accessible is not rendering for users who are not logged in.

### Steps to Reproduce

1. Open a public channel while not logged in (guest/unauthenticated view)
2. Observe that any messages containing image or file attachments do not display the media

### Expected Behavior

Images and file attachments should be visible to guest users in public channels, since the channel content is publicly accessible.

### Actual Behavior

Attachments are not displayed in guest view — images fail to load or are hidden entirely.

### Impact

- Guest users cannot see shared media in public channels
- Degrades the public-facing experience and SEO value of public channel pages

### Possible Area of Investigation

- Attachment/media URL generation or signing logic may be gating access behind authentication
- The frontend may be conditionally rendering attachment components only for authenticated sessions

declanblanc:
## Summary

- Guest users viewing public channels could not see images/file attachments because the two public REST endpoints (`GET /api/public/channels/:channelId/messages` and `GET /api/public/channels/:channelId/messages/:messageId`) omitted the `attachments` relation from the Prisma `select`
- Added `attachments: { select: { id, url, filename, contentType, sizeBytes } }` to both selects in `harmony-backend/src/routes/public.router.ts`
- The frontend `messageService.toFrontendMessage` already maps this shape correctly, so no frontend changes are needed

Closes #588

## Test plan

- [ ] Seed the local DB (`npm run db:seed:mock`), upload an image to a public channel as an authenticated user, then open the channel while logged out — image should render
- [ ] All 890 backend unit tests pass (`npm test` in `harmony-backend`)
- [ ] No frontend test changes required (frontend mapping logic was already correct)

### DIFF
--- a/harmony-backend/src/routes/public.router.ts
+++ b/harmony-backend/src/routes/public.router.ts
@@ -62,6 +62,9 @@ export function createPublicRouter(store?: Store) {
             createdAt: true,
             editedAt: true,
             author: { select: { id: true, username: true } },
+            attachments: {
+              select: { id: true, url: true, filename: true, contentType: true, sizeBytes: true },
+            },
           },
         });
 
@@ -111,6 +114,9 @@ export function createPublicRouter(store?: Store) {
             createdAt: true,
             editedAt: true,
             author: { select: { id: true, username: true } },
+            attachments: {
+              select: { id: true, url: true, filename: true, contentType: true, sizeBytes: true },
+            },
           },
         });
 


acabrera04 (approved):
- [x] **Prioritize Logic over Style** - The fix addresses the root logic gap: both guest-facing public message queries now include `attachments`, matching how the frontend maps message payloads.
- [x] **Security First** - Access control remains intact (`PRIVATE` channels are still blocked), and the added fields expose only attachment metadata needed for rendering.
- [x] **Architectural Alignment** - Change stays within the public route query/select layer and preserves existing service/adapter boundaries.
- [x] **Issue Completion** - The described issue (#588) is covered by updating both affected endpoints (`messages` list + single-message fetch).
- [x] **No Nitpicking** - Review focused on behavior, correctness, and risk; no style-only feedback.
- [x] **Avoid Repetition** - No existing review threads/comments were present to duplicate.
- [x] **Iterative Reviews** - This is a first-pass review on the current head, so no unresolved prior reviewer concerns to re-validate.
- [x] **Prevent CI Failures** - No syntax/import/runtime contract issues found in the diff, and current PR checks are passing.

Bundled review:

This change is scoped correctly and fixes the guest attachment visibility gap by returning attachment metadata from both public message endpoints. The selected attachment shape matches the frontend adapter expectations (`id`, `url`, `filename`, `contentType`, `sizeBytes`), and channel visibility guards remain in place. I did not find logic, security, architecture, or CI-break risks in the submitted diff.

PR #594, opened 2026-04-30
fix: re-anchor scroll to bottom when media loads in message list

declanblanc:
## Summary

- Adds a `ResizeObserver` on the message content wrapper that re-anchors `scrollTop` to the true bottom whenever content grows in height (images, inline video embeds, avatars — any source) and the user was already at the bottom
- Fixes a race condition where the component could mount with `messages = []`, mark itself as "mounted," then use **smooth scroll** when messages arrived; mid-animation scroll events would set `isNearBottomRef = false`, causing the `ResizeObserver` to skip re-anchoring when images loaded

## Root cause

When the API response is slow the component renders with an empty message list first. The previous code set `hasMountedRef = true` on that empty render, so when real messages arrived it fell through to the smooth-scroll path — a long animation from position 0 to the bottom. While animating, scroll events reported `distanceFromBottom > 100`, flipping `isNearBottomRef = false`. Images that loaded mid-animation were then ignored by the ResizeObserver.

## Test plan

- [ ] Open a channel with image/video attachments and verify the view stays pinned to the bottom as media loads
- [ ] Refresh the page many times and confirm the initial view is always at the bottom
- [ ] Verify that scrolling up and then receiving a new message does NOT auto-scroll (existing behavior preserved)
- [ ] Verify smooth scroll still works when a new message arrives while already at the bottom

### DIFF
--- a/harmony-frontend/src/components/channel/MessageList.tsx
+++ b/harmony-frontend/src/components/channel/MessageList.tsx
@@ -121,6 +121,10 @@ export function MessageList({
   }, []);
 
   useLayoutEffect(() => {
+    // Nothing to scroll to yet — don't mark as mounted so that the first
+    // real messages get the instant-scroll path instead of smooth scroll.
+    if (messages.length === 0) return;
+
     if (!hasMountedRef.current) {
       // Initial load: jump instantly so the user starts at the bottom
       hasMountedRef.current = true;
--- a/harmony-frontend/src/components/channel/MessageList.tsx
+++ b/harmony-frontend/src/components/channel/MessageList.tsx
@@ -102,6 +102,15 @@ export function MessageList({
     isNearBottomRef.current = distanceFromBottom <= 100;
   }, []);
 
+  // Called by MessageItem when an image or video finishes loading. If the user
+  // was at (or near) the bottom when the media started loading, the newly
+  // expanded content will have shifted the visible area up — re-anchor to bottom.
+  const handleMediaLoad = useCallback(() => {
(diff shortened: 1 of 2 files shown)


acabrera04 (approved):
- [x] **Prioritize Logic over Style**: Scroll anchoring logic addresses the reported race and preserves the near-bottom guard without style-only churn.
- [x] **Security First**: Change is client-side scroll behavior only; no auth, secret, input-validation, or trust-boundary regressions introduced.
- [x] **Architectural Alignment**: Frontend-only UI behavior change remains within component responsibilities; no conflict with backend architecture guidance.
- [x] **Issue Completion**: PR scope and implementation match the stated bug (media-height growth + initial empty-mount race) and preserve expected non-autoscroll behavior when user is not near bottom.
- [x] **No Nitpicking**: Review focused on logic/behavior/CI risk only; no stylistic or naming feedback.
- [x] **Avoid Repetition**: No prior review threads/comments exist on this PR, so no duplicate feedback was posted.
- [x] **Iterative Reviews**: First review on this PR from me; no previously raised concerns to re-check or resolve.
- [x] **Prevent CI Failures**: Diff is syntactically sound (imports/refs/effect cleanup), and required checks are passing.

Bundled review:

I reviewed the full patch in `MessageList.tsx` and validated the race-condition fix path: when the first render has no messages, initial mount no longer flips to smooth-scroll on first data arrival, and `ResizeObserver` re-anchors only when the user is still near bottom. This preserves expected behavior for users who have scrolled up while fixing media-load drift at the bottom. No blockers found.

PR #591, opened 2026-04-30
fix: derive unreadCount from notifications state + mark-as-read tests (#574)

declanblanc:
## Summary

- Removes the separate `unreadCount` state from `NotificationBell` and derives it via `useMemo` from the `notifications` array, eliminating any possible drift between the two
- Adds 31 automated tests (13 frontend + 18 backend) covering both single-read and bulk-read flows per the acceptance criteria

## Root Cause

`unreadCount` was maintained as an independent `useState` variable updated in sync with the `notifications` array in four separate call sites (`loadNotifications`, `markAsRead`, `markAllAsRead`, and the SSE optimistic path). Any one of these updating the array without also updating the counter (or vice-versa) would produce a stale badge. Using `useMemo` over the array makes the badge a pure function of the data — impossible to drift.

## Test plan

- [x] `harmony-frontend/src/__tests__/NotificationBell.test.tsx` — 13 tests: single-read mutation args, UI unread indicator removal, badge decrement, bulk-read mutation, badge reset, silent failure leaves state unchanged
- [x] `harmony-backend/tests/notification.service.test.ts` — 8 tests: `markAsRead` / `markAllAsRead` Prisma `updateMany` where-clause correctness, correct batch payload return, count 0 on no-op
- [x] `harmony-backend/tests/notification.router.test.ts` — 10 tests: authenticated requests succeed, unauthenticated rejected 401, invalid UUID rejected 400, correct tRPC result envelope
- [x] Full test suites pass — 908 backend tests, 445 frontend tests, zero regressions

Closes #574

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #602, opened 2026-05-01
fix: send no body instead of null for no-input tRPC mutations

### LINKED ISSUE #574: Notifications are not marked read via bulk or single actions
## Summary
Notifications remain unread even after using **Mark all as read** or clicking an individual notification.

## Problem
The read state is not being persisted/applied correctly in at least two flows:
- Bulk action: **Mark all as read**
- Single action: clicking an individual notification

## Steps to Reproduce
1. Have one or more unread notifications.
2. Click **Mark all as read**.
3. Observe unread indicators/count.
4. Reload the page and re-check notification state.
5. Repeat with a single unread notification by clicking it directly.

## Expected Behavior
- **Mark all as read** marks all current notifications as read immediately and persists after refresh.
- Clicking an individual notification marks only that notification as read immediately and persists after refresh.
- Unread badge/count updates accurately in both cases.

## Actual Behavior
Unread items continue to appear as unread after the actions above.

## Acceptance Criteria
- Bulk and single-read actions both update read state in UI and backend persistence.
- Unread badge/count decrements correctly for single-read and resets correctly for bulk-read.
- No regression to notification navigation/click behavior.
- Add automated tests for both flows (bulk read + single read).

declanblanc:
## Summary

Closes #574 (root cause — the badge-counter fix in #591 was incomplete).

- `trpcMutation` was passing `input ?? null` to Axios. When called with no input argument (i.e. `notification.markAllAsRead`), Axios serialized this to the JSON literal `null`, which the tRPC v11 HTTP adapter rejects with **HTTP 500** before even running auth middleware.
- The 500 was silently caught by the `catch {}` block in `markAllAsRead`, so no state update happened and the button appeared to do nothing.
- Removing the `?? null` fallback lets Axios omit the request body entirely when `input` is `undefined`, which tRPC handles correctly for void/no-input procedures.

## Root-cause evidence

```
# null body → 500 (broken)
curl -X POST .../trpc/notification.markAllAsRead -d 'null' -H 'Authorization: Bearer invalid'
→ HTTP 500 {"error":"Internal server error"}

# no body → 401 (correct — auth middleware runs)
curl -X POST .../trpc/notification.markAllAsRead -H 'Authorization: Bearer invalid'
→ HTTP 401 {"error":{"message":"UNAUTHORIZED",...}}

# {} body → 401 (also correct)
curl -X POST .../trpc/notification.markAllAsRead -d '{}' -H 'Authorization: Bearer invalid'
→ HTTP 401 {"error":{"message":"UNAUTHORIZED",...}}
```

## Why #591 didn't fully fix it

PR #591 fixed the `unreadCount` badge-counter drift (useMemo) and added tests, but the tests mock `trpcMutation` so they never exercise the real wire format. The actual HTTP call to `markAllAsRead` has been silently returning 500 since the endpoint was introduced.

## Test plan

- [x] All 457 frontend tests pass
- [x] `markAsRead` still returns the expected args (unchanged path)
- [x] All other `trpcMutation` callers pass input objects — they are unaffected

### DIFF
--- a/harmony-frontend/src/lib/api-client.ts
+++ b/harmony-frontend/src/lib/api-client.ts
@@ -234,7 +234,7 @@ class ApiClient {
 
   /** Call a tRPC mutation procedure (POST). Returns the unwrapped data. */
   async trpcMutation<T>(procedure: string, input?: unknown): Promise<T> {
-    const res = await this.client.post<TrpcResponse<T>>(`/trpc/${procedure}`, input ?? null);
+    const res = await this.client.post<TrpcResponse<T>>(`/trpc/${procedure}`, input);
     return res.data.result.data;
   }
 }


AvanishKulkarni (approved):
## Guideline Checklist

- [x] **Logic over Style** — Root cause is correctly identified. Removing `?? null` means `undefined` input flows through as-is, causing Axios to omit the request body entirely, which is what tRPC v11 expects for void/no-input procedures.
- [x] **Security First** — No new surface introduced. The change narrows rather than expands what is sent over the wire.
- [x] **Architectural Alignment** — The fix makes `trpcMutation` consistent with the `trpcQuery` sibling, which already guards `input !== undefined` before encoding the query string (api-client.ts:228–230).
- [x] **Issue Completion** — Closes #574. Test coverage for the notification flows was added in #591; this PR addresses the remaining root cause (the 500 on the wire). Acceptance criteria are now fully satisfied across both PRs.
- [x] **No Nitpicking** — N/A.
- [x] **Avoid Repetition** — No prior review comments on this PR.
- [x] **Iterative Reviews** — First review.
- [x] **Prevent CI Failures** — 457 frontend tests pass; verified that the only no-argument caller is `notification.markAllAsRead` (NotificationBell.tsx:180). All other callers pass explicit input objects and are unaffected.

---

**Approved.** The one-line change is correct, minimal, and well-evidenced. The curl output in the PR description is particularly useful as root-cause evidence — no concerns.

PR #605, opened 2026-05-02
feat: audio input/output device selection and inline voice settings

### LINKED ISSUE #603: feat: add audio input/output device selection in user settings

declanblanc:
Closes #603

## Summary

- Adds a **Voice & Audio** section to the user settings page with input/output device dropdowns, a live mic level meter, and a speaker test tone
- Adds an **inline voice settings popover** to the status bar so users can change devices without navigating away (which disconnects them from voice chat)
- Wires selected devices into `VoiceContext` — the stored input device is passed to Twilio's `audio` constraint on join, and `setSinkId` is called on remote audio elements for output routing

## Changes

| File | Change |
|------|--------|
| `src/hooks/useAudioDevices.ts` | New hook — enumerates `audioinput`/`audiooutput` devices, listens for `devicechange` hotplug events, persists selections to `localStorage` |
| `src/components/settings/AudioDeviceTests.tsx` | New shared file — `MicLevelMeter` and `SpeakerTest` components used by both the settings page and the status bar popover |
| `src/components/settings/AudioSettingsSection.tsx` | New settings section — full device selector UI with permission prompting |
| `src/components/settings/UserSettingsPage.tsx` | Adds **Voice & Audio** to the settings sidebar |
| `src/contexts/VoiceContext.tsx` | Reads stored device prefs on join; applies `setSinkId` to remote audio elements |
| `src/components/channel/UserStatusBar.tsx` | Adds sliders icon button + `VoiceSettingsPopover` — floats above the status bar, dismisses on click-outside, no navigation required |

## Test plan

- [ ] Open User Settings → Voice & Audio; verify input/output dropdowns populate after granting mic permission
- [ ] Select a non-default input device and click **Test Mic** — level meter should respond to speech
- [ ] Select a non-default output device and click **Test Speaker** — tone should play through that device
- [ ] Join a voice channel; verify the sliders icon appears in the status bar
- [ ] Click the sliders icon while in voice — popover opens above the status bar without disconnecting
- [ ] Change device selections in the popover; leave and rejoin — new device should be used
- [ ] Click outside the popover — it dismisses
- [ ] Verify 457 frontend tests still pass (`npm test` in `harmony-frontend/`)


harmony-frontend/src/contexts/VoiceContext.tsx
```diff
@@ -290,9 +315,13 @@ export function VoiceProvider({ children, serverId, voiceChannelIds, currentUser
 
         // Dynamic import keeps the Twilio SDK out of SSR.
         const TwilioVideo = await import('twilio-video');
+        const inputDeviceId = getStoredAudioInputDeviceId();
         const room = await TwilioVideo.connect(token, {
           name: channelId,
-          audio: true,
+          audio:
+            inputDeviceId && inputDeviceId !== 'default'
```
AvanishKulkarni, line 322: **Bug — use `ideal` instead of `exact`**

Passing `{ deviceId: { exact: inputDeviceId } }` will throw `OverconstrainedError` and abort the voice join when the stored device is no longer available (e.g. a USB headset unplugged between sessions). Issue #603 AC #6 requires a graceful fallback to the system default.

Change to:
```ts
audio:
  inputDeviceId && inputDeviceId !== 'default'
    ? { deviceId: { ideal: inputDeviceId } }
    : true,
```
`ideal` selects the preferred device when available but silently falls back to default rather than rejecting.

harmony-frontend/src/lib/api-client.ts
```diff
@@ -234,7 +234,7 @@ class ApiClient {
```
AvanishKulkarni, line 234: **Undocumented behaviour change — verify intentional**

Changing `input ?? null` to bare `input` means calls where `input` is `undefined` will now omit the POST body entirely instead of sending `null`. tRPC mutations typically expect a JSON body (even `null`), and the original `?? null` guard was likely there for that reason. This is unrelated to the audio feature — please revert or add a regression test confirming existing no-arg mutations still work correctly.

harmony-frontend/src/components/settings/AudioDeviceTests.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+      if (deviceId && deviceId !== 'default' && 'setSinkId' in audio) {
+        await (audio as HTMLAudioElement & { setSinkId: (id: string) => Promise<void> }).setSinkId(
+          deviceId,
+        );
+      }
+
+      oscillator.start();
+      oscillator.stop(ctx.currentTime + 1.0);
+      await audio.play();
```
AvanishKulkarni, line 152: **Minor resource leak — `AudioContext` not closed in error path**

`ctx` is created at the top of the `try` block but is only closed inside `oscillator.onended`. If `audio.play()` throws here, `ctx` is never closed. Consider:
```ts
} catch {
  void ctx?.close();
  setError('Could not play test tone.');
  setPlaying(false);
}
```
Not a blocker, but worth fixing while the code is new.

AvanishKulkarni (changes requested):

- [x] **Prioritize Logic over Style** — Code is well-structured and DRY; the hook / shared-components split cleanly separates concerns.
- [x] **Security First** — No secrets or unvalidated external input paths introduced.
- [x] **Architectural Alignment** — Follows existing frontend patterns (hooks, context, settings page structure).
- [ ] **Issue Completion** — AC #6 (graceful fallback when saved device is no longer available) is not met; see inline comment on `VoiceContext.tsx`.
- [x] **No Nitpicking** — Only functional/correctness issues flagged below.
- [ ] **Prevent CI Failures** — Undocumented change in `api-client.ts` may break existing tests; see inline comment.

---

## Review

Good overall design. The `useAudioDevices` hook as a shared source of truth, the extracted `AudioDeviceTests` subcomponents, and the popover pattern for in-call device changes are all solid. Two issues need attention before merge:

### Blocking — Twilio `exact` constraint breaks AC #6

When the stored input device is no longer attached, passing `{ deviceId: { exact: id } }` causes Twilio's `connect()` call to throw an `OverconstrainedError`, making voice join fail entirely rather than falling back gracefully to the system default. Issue #603 AC #6 explicitly requires a graceful fallback. The fix is to use `ideal` instead of `exact`.

### Non-blocking — Undocumented behaviour change in `api-client.ts`

The diff swaps `input ?? null` to bare `input`. When `input` is `undefined`, axios will omit the request body rather than send `null`. tRPC expects a JSON body even for no-argument mutations; the original `?? null` was likely there to guarantee that. This change is unrelated to the audio feature and should either be reverted or documented, and covered by an existing or new test.

declanblanc:
Closing in favor of #604 — all audio device feature work (Test Speaker button, inline voice settings popover in status bar, `useAudioDevices` hook, `AudioDeviceTests` component, and refactored `AudioSettingsSection`) has been merged into the `feature/add-audio-inputoutput-device-selection-in-user-settings` branch.

PR #606, opened 2026-05-02
feat: add load-older-messages pagination to channel message list

declanblanc:
## Summary

- Users were limited to the 50 most recent messages when viewing a channel — there was no way to scroll back through older history
- Wires the backend's existing cursor-based pagination (`message.getMessages` with `cursor`) to the frontend for the main channel message list
- Adds a \"Load older messages\" button at the top of the message list; each click fetches the next 50 older messages and prepends them, preserving the user's current scroll position so the viewport does not jump

## Changes

| File | What changed |
|---|---|
| `messageService.ts` | `getMessages` now accepts `cursor?` and returns `nextCursor` |
| `getOlderMessages.ts` (new) | Server action wrapping `getMessages` with a cursor, matching the `getThreadMessagesAction` pattern |
| `ChannelPageContent.tsx` | SSR captures `hasMore` and `nextCursor` from the initial fetch and forwards them to `HarmonyShell` |
| `HarmonyShell.tsx` | Holds `hasMoreOlder`/`olderCursor`/`isLoadingOlder` state; `handleLoadOlderMessages` prepends results and advances cursor; state resets on channel switch |
| `MessageList.tsx` | Shows a \"Load older messages\" button (or loading indicator) at the top; saves `scrollHeight` before prepending and restores it in `useLayoutEffect` to anchor the viewport |

## Test plan

- [ ] Open a channel with more than 50 messages — a \"Load older messages\" button should appear at the top
- [ ] Click the button — older messages should appear above the current ones without the viewport jumping
- [ ] Click again if more history exists — continues loading back in time
- [ ] When all history is loaded — button disappears
- [ ] Switching channels — button/state resets correctly for the new channel
- [ ] Channels with ≤ 50 messages — no button shown
- [ ] All 457 frontend tests pass

### DIFF
--- a/harmony-frontend/src/app/actions/getOlderMessages.ts
+++ b/harmony-frontend/src/app/actions/getOlderMessages.ts
@@ -1,0 +1,22 @@
+'use server';
+
+import { getMessages } from '@/services/messageService';
+import type { Message } from '@/types';
+
+export type GetOlderMessagesResult =
+  | { ok: true; messages: Message[]; nextCursor: string | undefined; hasMore: boolean }
+  | { ok: false; error: string };
+
+export async function getOlderMessagesAction(
+  channelId: string,
+  serverId: string,
+  cursor: string,
+): Promise<GetOlderMessagesResult> {
+  try {
+    const result = await getMessages(channelId, 1, { serverId, cursor });
+    return { ok: true, ...result };
+  } catch (err) {
+    const message = err instanceof Error ? err.message : 'Failed to load older messages.';
+    return { ok: false, error: message };
+  }
+}
--- a/harmony-frontend/src/app/actions/getOlderMessages.ts
+++ b/harmony-frontend/src/app/actions/getOlderMessages.ts
@@ -13,7 +13,7 @@ export async function getOlderMessagesAction(
   cursor: string,
 ): Promise<GetOlderMessagesResult> {
   try {
-    const result = await getMessages(channelId, 1, { serverId, cursor });
+    const result = await getMessages(channelId, undefined, { serverId, cursor });
     return { ok: true, ...result };
   } catch (err) {
     const message = err instanceof Error ? err.message : 'Failed to load older messages.';
--- a/harmony-frontend/src/components/channel/ChannelPageContent.tsx
+++ b/harmony-frontend/src/components/channel/ChannelPageContent.tsx
@@ -63,9 +63,10 @@ export async function ChannelPageContent({
     currentMember?.role === 'admin' ||
     currentMember?.role === 'owner';
   const isLockedPrivateChannel = channel.visibility === ChannelVisibility.PRIVATE && !isServerAdmin;
-  const sortedMessages = isLockedPrivateChannel
-    ? []
-    : [...(await getMessages(channel.id, 1, { serverId: server.id })).messages].reverse();
+  const initialMessagesResult = isLockedPrivateChannel
+    ? { messages: [], hasMore: false, nextCursor: undefined }
+    : await getMessages(channel.id, 1, { serverId: server.id });
+  const sortedMessages = [...initialMessagesResult.messages].reverse();
 
   return (
     <HarmonyShell
@@ -75,6 +76,8 @@ export async function ChannelPageContent({
       allChannels={allChannels}
       currentChannel={channel}
       messages={sortedMessages}
+      initialHasMore={initialMessagesResult.hasMore}
+      initialNextCursor={initialMessagesResult.nextCursor}
       members={members}
       basePath={isGuestView ? '/c' : '/channels'}
       lockedMessagePane={
--- a/harmony-frontend/src/components/channel/MessageList.tsx
+++ b/harmony-frontend/src/components/channel/MessageList.tsx
@@ -108,12 +108,22 @@ export function MessageList({
   // to prevent the viewport from jumping when older messages are added at the top.
   const savedScrollHeightRef = useRef(0);
 
+  const triggerLoadOlder = useCallback(() => {
+    const el = scrollContainerRef.current;
+    if (el) savedScrollHeightRef.current = el.scrollHeight;
+    onLoadOlderMessages?.();
+  }, [onLoadOlderMessages]);
+
   const handleScroll = useCallback(() => {
     const el = scrollContainerRef.current;
     if (!el) return;
     const distanceFromBottom = el.scrollHeight - el.scrollTop - el.clientHeight;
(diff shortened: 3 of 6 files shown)


AvanishKulkarni (changes requested):
## PR Review Checklist

- [x] **Logic over Style** — Core pagination logic is sound; cursor-based prepend + scroll anchor is the correct pattern
- [ ] **Security First** — No security vulnerabilities introduced
- [x] **Architectural Alignment** — Follows the existing `getThreadMessagesAction` pattern for server actions
- [ ] **Issue Completion** — PR appears complete against the described feature; no linked issue to verify acceptance criteria against
- [x] **No Nitpicking** — Focused on correctness and logic only
- [x] **Avoid Repetition** — First review on this PR
- [x] **Iterative Reviews** — N/A (first review)
- [ ] **Prevent CI Failures** — One race condition identified that can cause observable bugs

---

## Review

The overall approach is well-structured: the scroll-anchor technique (`savedScrollHeight` → `useLayoutEffect` restore) is the right tool, and the state reset on channel switch is handled correctly. One meaningful bug and a secondary consequence of it need to be addressed before merge.

### Critical: Race condition in scroll-based auto-trigger

**`harmony-frontend/src/components/channel/MessageList.tsx`, around the `handleScroll` callback:**

```ts
if (el.scrollTop <= 200 && hasMoreOlder && !isLoadingOlder) {
  triggerLoadOlder();
}
```

`handleScroll` fires on every scroll event (many times per second). Between the moment `triggerLoadOlder()` is called and when React re-renders `MessageList` with the updated `isLoadingOlder = true` prop from `HarmonyShell`, a stale closure sees `isLoadingOlder` still as `false`. This means multiple rapid scroll events can each pass the guard and call `onLoadOlderMessages()` in the same render cycle, causing **duplicate concurrent fetches**.

In `HarmonyShell.handleLoadOlderMessages`, the guard `if (!olderCursor || isLoadingOlder) return` suffers the same stale-closure problem — both calls see `isLoadingOlder = false` before the first `setIsLoadingOlder(true)` has propagated.

**Consequence:** The second call also enters `setIsLoadingOlder(true)` and dispatches a second `getOlderMessagesAction`. Both eventually resolve and `setLocalMessages` is called twice with potentially overlapping pages, producing duplicate messages in the list.

**Fix:** Add a `useRef` loading guard in `HarmonyShell` that is set synchronously (before any `await`) and checked on entry:

```ts
// In HarmonyShell
const isLoadingOlderRef = useRef(false);

const handleLoadOlderMessages = useCallback(async () => {
  if (!olderCursor || isLoadingOlderRef.current) return;
  isLoadingOlderRef.current = true;
  setIsLoadingOlder(true);
  // ... rest of fetch
  isLoadingOlderRef.current = false;
  setIsLoadingOlder(false);
}, [...]);
```

The ref acts as a synchronous mutex; the state variable still drives the UI indicator.

---

### Secondary consequence: `savedScrollHeightRef` overwrite

**`harmony-frontend/src/components/channel/MessageList.tsx`, `triggerLoadOlder`:**

```ts
const triggerLoadOlder = useCallback(() => {
  const el = scrollContainerRef.current;
  if (el) savedScrollHeightRef.current = el.scrollHeight;
  onLoadOlderMessages?.();
}, [onLoadOlderMessages]);
```

Because `triggerLoadOlder` can be called multiple times before the layout effect runs (due to the race above), `savedScrollHeightRef.current` gets overwritten on each call. When the layout effect fires after the first batch of messages arrives, it uses the *last* saved height instead of the one captured just before the first prepend, producing an incorrect scroll position.

Once the ref-based guard above is in place this becomes a non-issue, but it is worth noting as the root cause of the scroll-jump symptom that the race would produce.

---

### Minor: `page = 1` in server action is misleading but harmless

**`harmony-frontend/src/app/actions/getOlderMessages.ts:15`:**

```ts
const result = await getMessages(channelId, 1, { serverId, cursor });
```

The `page` argument is only consumed by the unauthenticated REST code path inside `getMessages`. Since `serverId` is always provided here, the tRPC path is always taken and `page` is ignored. It's not a bug, but passing `1` implies paging semantics that don't apply — consider passing `undefined` or documenting the intent to make this clearer for future readers.

declanblanc:
> Re: comment 4215185881

Fixed in c026eff1.

**Race condition (critical):** Added `isLoadingOlderRef = useRef(false)` as a synchronous mutex in `HarmonyShell`. It's set to `true` before the `await` and cleared after, so any concurrent scroll events that enter `handleLoadOlderMessages` before React re-renders will hit the ref guard and bail immediately. The state variable `isLoadingOlder` still drives the UI loading indicator.

**Scroll height overwrite (secondary):** Resolved as a direct consequence of the ref fix — `triggerLoadOlder` can only execute the fetch path once per load cycle now.

**Misleading `page = 1` arg:** Changed to `undefined` in `getOlderMessagesAction` so the call signature matches actual semantics (cursor-based, not page-based).

AvanishKulkarni (approved):
## PR Review Checklist

- [x] **Logic over Style** — Core pagination logic is sound; cursor-based prepend + scroll anchor is correct
- [x] **Security First** — No vulnerabilities introduced
- [x] **Architectural Alignment** — Follows existing `getThreadMessagesAction` pattern
- [x] **Issue Completion** — All three issues from the previous review have been addressed
- [x] **No Nitpicking** — Focused on correctness only
- [x] **Avoid Repetition** — Previous comments addressed below
- [x] **Iterative Reviews** — Checking prior feedback below
- [x] **Prevent CI Failures** — No logic issues remain

---

## Follow-up on Previous Review

All three items from the prior `CHANGES_REQUESTED` review are resolved:

**Race condition (critical) — ✅ Resolved**
`isLoadingOlderRef` is now used as a synchronous mutex in `handleLoadOlderMessages`. It is set to `true` before the `await` and back to `false` after, so any rapid scroll events that call `handleLoadOlderMessages` before React re-renders will hit the ref guard and return immediately — duplicate fetches are no longer possible. The ref is also correctly reset on channel switch alongside the state variable.

**`savedScrollHeightRef` overwrite (secondary) — ✅ Resolved**
With the synchronous mutex in place, `triggerLoadOlder` can only be active once at a time. The saved scroll height is captured exactly once per prepend cycle and is not overwritten before the layout effect runs.

**`page = 1` in server action (minor) — ✅ Resolved**
`getOlderMessages.ts` now passes `undefined` for `page`, removing the misleading page-semantics implication.

Approving.

declanblanc:
> Re: comment 4215245882

Thank you for the thorough follow-up review!

PR #607, opened 2026-05-02
feat: add delete message for authors and moderators

declanblanc:
## Summary

- Adds a **Delete Message** button to the message hover action bar (⋯ More menu) for message authors and MODERATOR+ users
- Wires the existing backend `message.deleteMessage` tRPC endpoint (which was already fully implemented with soft-delete, permission checks, and SSE event publishing) to a new frontend server action
- Deletion requires a confirmation dialog and is optimistically removed from local state; the SSE `message:deleted` event provides redundant cleanup

## Changes

- **`src/app/actions/deleteMessage.ts`** — new server action wrapping the tRPC mutation, returning a typed result distinguishing 403 from other errors
- **`MessageItem.tsx`** — `ActionBar` exposes a red "Delete Message" entry in the More dropdown; visible to the message author (`isOwnMessage`) or users with `canDeleteAny` (MODERATOR+); includes loading state and inline error feedback
- **`MessageList.tsx`** — threads `canDeleteAny` and `onDelete` props down to `MessageItem`
- **`HarmonyShell.tsx`** — computes `canDeleteAny` (reuses `canPin`'s MODERATOR+ check), adds `handleDeleteMessage` for optimistic local state removal, passes both to `MessageList`

## Test plan

- [ ] Log in as a regular member, hover a message you sent — confirm "Delete Message" appears in the ⋯ menu
- [ ] Click Delete Message → confirm dialog appears → confirm → message disappears immediately
- [ ] Log in as a regular member, hover another user's message — confirm "Delete Message" does **not** appear
- [ ] Log in as a MODERATOR+ user, hover any message — confirm "Delete Message" appears regardless of author
- [ ] Cancel the confirm dialog — confirm message is **not** deleted
- [ ] Verify that other open sessions/tabs also see the message disappear via the SSE `message:deleted` event


harmony-frontend/src/lib/api-client.ts
```diff
@@ -234,7 +234,7 @@ class ApiClient {
 
   /** Call a tRPC mutation procedure (POST). Returns the unwrapped data. */
   async trpcMutation<T>(procedure: string, input?: unknown): Promise<T> {
```
AvanishKulkarni, line 236: This line was changed from `input ?? null` to `input`, but `NotificationBell.tsx:185` calls `trpcMutation('notification.markAllAsRead')` with no argument — so `input` is `undefined` here. Previously axios sent a `null` JSON body; now it sends no body at all. If the backend procedure's input schema rejects a missing body, this breaks `markAllAsRead` at runtime. Revert to `input ?? null` and keep this change out of the PR.

harmony-frontend/src/components/layout/HarmonyShell.tsx
```diff
@@ -611,8 +615,10 @@ export function HarmonyShell({
                     messages={localMessages}
                     serverId={currentServer.id}
```
AvanishKulkarni: `canDeleteAny={canPin}` silently couples two independent permissions. If `canPin` is later tightened to ADMIN-only, delete-any access would change with it unintentionally. Extract a shared constant instead: `const canModerate = memberRole === 'MODERATOR' || memberRole === 'ADMIN';` and pass it to both props explicitly.

AvanishKulkarni (changes requested):
## PR Review Checklist

- [x] **1. Prioritize Logic over Style** — Logic is sound; optimistic removal + SSE redundant cleanup is correct. `filter` on a missing ID is a no-op, so dual-path deletion (optimistic + SSE) is safe.
- [x] **2. Security First** — Permission checks delegate to the backend. The server action correctly distinguishes 403 from other errors; no secrets or unvalidated inputs are exposed.
- [x] **3. Architectural Alignment** — Server action pattern matches `pinMessage.ts` / `editMessage.ts` conventions. Prop threading follows existing `canPin`/`onPinToggle` patterns.
- [x] **4. Issue Completion** — PR description covers the stated scope (delete for authors and MODERATOR+, confirmation, optimistic removal, SSE cleanup).
- [x] **5. No Nitpicking** — Only critical issues flagged below.
- [x] **6. Avoid Repetition** — No prior reviews exist on this PR.
- [x] **7. Iterative Reviews** — First review; N/A.
- [ ] **8. Prevent CI Failures** — The `api-client.ts` change (see inline) alters behavior for existing no-input mutation call sites and could cause a runtime regression.

---

## Review

The core feature — delete button, confirmation, optimistic local state removal, SSE redundancy — is implemented cleanly and follows the patterns established by pin/edit. Two issues need addressing before this is mergeable:

**1. Unintended behavior change in `api-client.ts` (blocking)**

The change from `input ?? null` to bare `input` is out of scope and silently alters behavior for every no-input `trpcMutation` call in the app. `NotificationBell.tsx:185` calls `apiClient.trpcMutation('notification.markAllAsRead')` with no input. Previously this sent a `null` JSON body; after this change axios omits the body entirely. If the backend procedure validates or expects a body (even an empty one), this breaks `markAllAsRead` at runtime. The `?? null` guard was likely deliberate. Revert this line to `input ?? null` and keep the api-client change out of this PR.

**2. Silent coupling of `canDeleteAny` to `canPin` (`HarmonyShell.tsx:618`)**

```tsx
canDeleteAny={canPin}
```

Both permissions happen to map to MODERATOR+ today, but they are semantically independent. If `canPin` is ever scoped differently (e.g., locked to ADMIN), delete-any permissions will silently change with it. Introduce a shared local constant:

```tsx
const canModerate = memberRole === 'MODERATOR' || memberRole === 'ADMIN';
// …
canPin={canModerate}
canDeleteAny={canModerate}
```

This makes the intent explicit and eliminates the implicit dependency.

harmony-frontend/src/lib/api-client.ts
acabrera04: Fixed in 07bbde42. Restored the no-input tRPC mutation behavior so omitted input is sent as null again, preserving existing callers like notification.markAllAsRead.

harmony-frontend/src/components/layout/HarmonyShell.tsx
acabrera04: Fixed in 07bbde42. Added an explicit canModerate permission and now pass canDeleteAny from that moderation permission instead of coupling it to canPin.

FardeenI (dismissed):
> The two previously-blocking items look addressed (client `trpcMutation` restoring `input ?? null`, and decoupling `canDeleteAny` from `canPin` via `canModerate`). CI is also green.
>
> **Must-fix: deleting a pinned message can leave its content visible in the pinned panel until refresh**  
> This undermines the expectation of deletion (and can effectively “leak” deleted content in the UI until the user closes/reopens or something else triggers a refresh).
>
> **Inline notes**
>
> - `harmony-frontend/src/components/layout/HarmonyShell.tsx:~323-329` / `:~239-243`  
>   Deletion removes from `localMessages`, but does **not** trigger a pinned-messages refresh:
>   ```ts
>   setLocalMessages(prev => prev.filter(m => m.id !== messageId));
>   ```
>   Same in the SSE path:
>   ```ts
>   setLocalMessages(prev => prev.filter(m => m.id !== messageId));
>   ```
>
> - `harmony-frontend/src/components/channel/PinnedMessagesPanel.tsx:87-105`  
>   The pinned panel only refetches when `refreshKey` changes (or panel is reopened / channel changes):
>   ```ts
>   useEffect(() => { ... }, [isOpen, channelId, serverId, refreshKey]);
>   ```
>
> **Suggested fix (conceptual)**: when a message is deleted in the active channel (both the local “onDelete” path and the `message:deleted` SSE path), bump `pinsRefreshKey` (or otherwise invalidate/refetch pinned messages) so an already-open pinned panel cannot continue rendering now-deleted message content.

acabrera04 (commented):
Fixed in 8672b5d6. Both delete paths now bump the pinned-messages refresh key: the local delete callback and the message:deleted SSE handler. That forces an already-open pinned panel to refetch so deleted pinned content is removed without waiting for a panel reopen or unrelated refresh.

harmony-frontend/src/components/layout/HarmonyShell.tsx
```diff
@@ -263,6 +264,11 @@ export function HarmonyShell({
     setPinsRefreshKey(prev => prev + 1);
   }, []);
 
+  const handleDeleteMessage = useCallback((messageId: string) => {
+    setLocalMessages(prev => prev.filter(m => m.id !== messageId));
```
AvanishKulkarni, line 272: **Must fix: deleting a pinned message leaks its content in the pinned panel until refresh.**

`handleDeleteMessage` removes the message from `localMessages` but never bumps `pinsRefreshKey`. If the pinned panel is open when a user deletes a pinned message, the panel continues rendering the deleted content until the user manually closes/reopens it.

Suggested fix:
```ts
const handleDeleteMessage = useCallback((messageId: string) => {
  setLocalMessages(prev => prev.filter(m => m.id !== messageId));
  setPinsRefreshKey(prev => prev + 1); // evict from open pinned panel
}, []);
```

harmony-frontend/src/components/layout/HarmonyShell.tsx
```diff
@@ -324,6 +330,7 @@ export function HarmonyShell({
     (messageId: string, channelId: string) => {
       if (channelId !== currentChannel.id) return;
       setLocalMessages(prev => prev.filter(m => m.id !== messageId));
```
AvanishKulkarni, line 336: Same gap as the `handleDeleteMessage` path above: the SSE `message:deleted` handler also only calls `setLocalMessages` without bumping `pinsRefreshKey`. Apply the same `setPinsRefreshKey(prev => prev + 1)` call here so that remote deletions (other users / other tabs) also evict pinned content from the open panel.

harmony-frontend/src/components/message/MessageItem.tsx
```diff
@@ -479,13 +492,49 @@ function ActionBar({
     }
   }, [isPinned, messageId, onPinToggle, serverId]);
 
+  const handleDelete = useCallback(async () => {
+    if (!serverId) return;
+    if (!window.confirm('Delete this message? This cannot be undone.')) return;
```
AvanishKulkarni, line 497: **Suggestion (non-blocking):** `window.confirm` is a native browser dialog that can be suppressed by pop-up blockers and doesn't match the app's design language. An inline toggle in the dropdown (e.g. "Are you sure? **[Confirm]** [Cancel]") would be more consistent. Fine to address in a follow-up.

harmony-frontend/src/components/message/MessageItem.tsx
```diff
@@ -479,13 +492,49 @@ function ActionBar({
...
+          : 'Failed to delete message. Please try again.';
+        setDeleteErrorMsg(msg);
+        setDeleteState('error');
+        if (deleteErrorTimerRef.current) clearTimeout(deleteErrorTimerRef.current);
+        deleteErrorTimerRef.current = setTimeout(() => {
+          setDeleteState('idle');
+          setDeleteErrorMsg('');
+        }, 3000);
+      }
+    } catch {
```
AvanishKulkarni, line 516: **Suggestion (non-blocking):** The `catch` block is identical to the `!result.ok` error branch. Extract a small helper to avoid duplication:
```ts
const showDeleteError = (msg: string) => {
  setDeleteErrorMsg(msg);
  setDeleteState('error');
  if (deleteErrorTimerRef.current) clearTimeout(deleteErrorTimerRef.current);
  deleteErrorTimerRef.current = setTimeout(() => {
    setDeleteState('idle');
    setDeleteErrorMsg('');
  }, 3000);
};
```
Fine to address in a follow-up.

AvanishKulkarni (changes requested):
The two previously-blocking items are confirmed fixed (`api-client.ts` `input ?? null` restored, `canModerate` extracted and passed as `canDeleteAny`). The core feature is clean and follows the pin/edit patterns correctly. One new must-fix below.

harmony-frontend/src/components/layout/HarmonyShell.tsx
acabrera04: Fixed in 12fa198b. The local delete callback now bumps pinsRefreshKey after removing the message from localMessages, so an open pinned panel refetches and stops showing deleted pinned content.

harmony-frontend/src/components/layout/HarmonyShell.tsx
acabrera04: Fixed in 12fa198b. The message:deleted SSE handler now also bumps pinsRefreshKey after removing the message, so remote deletes and other-tab deletes evict deleted pinned content from an open panel.

harmony-frontend/src/components/message/MessageItem.tsx
acabrera04: Deferred follow-up. This was marked non-blocking, so I kept this PR focused on the merge conflict and required pinned-panel invalidation fixes rather than changing the delete confirmation interaction.

harmony-frontend/src/components/message/MessageItem.tsx
acabrera04: Deferred follow-up. This was explicitly non-blocking; I left the existing delete-error handling shape unchanged to avoid broadening this PR while resolving the required behavior and merge conflicts.

AvanishKulkarni (approved):
Both blocking items are resolved:

- `handleDeleteMessage` now calls `setPinsRefreshKey(prev => prev + 1)` — an open pinned panel correctly evicts deleted messages on the local path.
- The SSE `message:deleted` handler also bumps `pinsRefreshKey` — remote deletions are covered too.

The two non-blocking suggestions (`window.confirm` UX and the duplicate error-handling in the `catch` block) can be followed up separately. Feature is correct and ready to merge.

PR #608, opened 2026-05-02
feat: delete text/voice channels (admin/owner)

declanblanc:
## Summary

- Adds a **settings gear icon** to both text and voice channel sidebar rows (admins/owners only, hover-revealed), giving all channel types a direct path to their settings page
- Adds a **Delete Channel** section to the channel settings page for both text and voice channels
- Enforces a guard preventing deletion of the last text channel in a server (backend + frontend)
- Fixes a tRPC void-return bug that caused a spurious error after successful deletion

### Backend (`harmony-backend`)

- **`src/repositories/channel.repository.ts`** — new `countByServerIdAndType` method used to count text channels before deletion
- **`src/services/channel.service.ts`** — `deleteChannel` now:
  - Rejects with `BAD_REQUEST` if the channel is a TEXT channel and it's the last one in the server
  - Returns the deleted channel record instead of `void`, so tRPC serializes a proper `result.data` field (fixes "response missing result.data" error on success)
- **`tests/channel.service.events.test.ts`** — updated Prisma mock to include `count`, defaulting to 2 so existing event-publishing tests continue to pass

### Frontend (`harmony-frontend`)

- **`src/components/channel/ChannelSidebar.tsx`** — both text and voice channel rows now show a hover-revealed settings gear icon (admin/owner only) linking to `/settings/[serverSlug]/[channelSlug]`
- **`src/app/settings/[serverSlug]/[channelSlug]/page.tsx`** — fetches the full channel list at page load and derives `isLastTextChannel` to pass down
- **`src/app/settings/[serverSlug]/[channelSlug]/actions.ts`** — new `deleteChannelAction` server action: resolves channel/server from URL slugs (defense-in-depth), calls the service, invalidates caches, then redirects to the server home
- **`src/components/settings/ChannelSettingsPage.tsx`** — new **Delete Channel** sidebar section (styled in red) with:
  - A confirmation input requiring the user to type the exact channel name before the button activates
  - A blocked state (no confirmation form, explanatory message) when the channel is the only text channel — voice channels are always deletable

## Reviewer notes

- Voice channels previously had no route to their settings page because they render as join-buttons rather than navigation links. The gear icon resolves this without changing the join interaction. Text channels get the same treatment for consistency
- The backend guard is the authoritative enforcement point for the last-text-channel rule; the frontend disabled state is a UX convenience only
- The `trpcMutate` helper throws when `result.data` is `undefined` — returning the deleted record from the service is consistent with how `deleteServer` is already handled elsewhere in the codebase
- The two pre-existing test failures (`metaTagUpdate.integration`, `cache.middleware`) are unrelated to this change and were failing on `main` before this branch

### DIFF
--- a/harmony-frontend/src/components/settings/ChannelSettingsPage.tsx
+++ b/harmony-frontend/src/components/settings/ChannelSettingsPage.tsx
@@ -12,6 +12,7 @@ import { cn, getUserErrorMessage } from '@/lib/utils';
 import {
   saveChannelSettings,
   fetchAuditLog,
+  deleteChannelAction,
 } from '@/app/settings/[serverSlug]/[channelSlug]/actions';
 import { VisibilityToggle } from '@/components/channel/VisibilityToggle';
 import { SeoPreviewSection } from '@/components/settings/SeoPreviewSection';
@@ -108,14 +109,15 @@ function ChannelNotificationsSection({ channel, serverId }: { channel: Channel; 
   );
 }
 
-type Section = 'overview' | 'permissions' | 'visibility' | 'seo' | 'notifications';
+type Section = 'overview' | 'permissions' | 'visibility' | 'seo' | 'notifications' | 'danger';
 
-const SECTIONS: { id: Section; label: string }[] = [
+const SECTIONS: { id: Section; label: string; danger?: boolean }[] = [
   { id: 'overview', label: 'Overview' },
   { id: 'permissions', label: 'Permissions' },
   { id: 'visibility', label: 'Visibility' },
   { id: 'seo', label: 'SEO Preview' },
   { id: 'notifications', label: 'Notifications' },
+  { id: 'danger', label: 'Delete Channel', danger: true },
 ];
 
 // ─── Overview section ─────────────────────────────────────────────────────────
@@ -608,7 +610,71 @@ function VisibilitySection({
   );
 }
 
-// ─── Loading spinner ──────────────────────────────────────────────────────────
+// ─── Danger zone section ──────────────────────────────────────────────────────
+
+function DangerZoneSection({ channel, serverSlug }: { channel: Channel; serverSlug: string }) {
+  const [confirmText, setConfirmText] = useState('');
+  const [deleting, setDeleting] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const nameMatches = confirmText === channel.name;
+
+  async function handleDelete() {
+    if (!nameMatches || deleting) return;
+    setDeleting(true);
+    setError(null);
+    try {
+      await deleteChannelAction(serverSlug, channel.slug);
+    } catch (err) {
+      setError(getUserErrorMessage(err, 'Failed to delete channel.'));
+      setDeleting(false);
+    }
+  }
+
+  return (
+    <div className='max-w-lg space-y-6'>
+      <h2 className='text-xl font-semibold text-white'>Delete Channel</h2>
+      <div className='rounded border border-red-500/40 bg-red-950/20 p-5 space-y-4'>
+        <p className='text-sm text-gray-300'>
+          Deleting <span className='font-semibold text-white'>#{channel.name}</span> is permanent
+          and cannot be undone. All messages and settings for this channel will be lost.
+        </p>
+        <div>
+          <label
+            htmlFor='confirm-channel-name'
+            className='mb-1.5 block text-xs font-semibold uppercase tracking-wide text-gray-400'
+          >
+            Type the channel name to confirm
+          </label>
+          <input
+            id='confirm-channel-name'
+            type='text'
+            value={confirmText}
+            onChange={e => setConfirmText(e.target.value)}
+            placeholder={channel.name}
+            disabled={deleting}
+            className={cn(
+              'w-full rounded px-3 py-2 text-sm text-white placeholder-gray-600 outline-none',
+              'focus:ring-2 focus:ring-red-500 disabled:opacity-50',
+              BG.input,
+            )}
+          />
+        </div>
+        {error && <p className='text-xs text-red-400'>{error}</p>}
+        <button
+          type='button'
+          onClick={handleDelete}
+          disabled={!nameMatches || deleting}
+          className='rounded px-4 py-2 text-sm font-medium text-white bg-red-600 hover:bg-red-700 disabled:opacity-40 disabled:cursor-not-allowed transition-colors'
+        >
+          {deleting ? 'Deleting…' : 'Delete Channel'}
+        </button>
+      </div>
+    </div>
+  );
+}
+
+// ─── Props ────────────────────────────────────────────────────────────────────
 
 export interface ChannelSettingsPageProps {
   channel: Channel;
@@ -673,7 +739,7 @@ export function ChannelSettingsPage({
 
         {/* Nav items */}
         <nav aria-label='Settings sections'>
-          {SECTIONS.map(({ id, label }) => (
+          {SECTIONS.map(({ id, label, danger }) => (
             <button
               key={id}
               type='button'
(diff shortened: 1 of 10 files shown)


AvanishKulkarni (changes requested):
## Review

**Guideline Checklist**

- [x] Prioritize Logic over Style
- [x] Security First — slug-based ID resolution in the server action is correct defense-in-depth
- [x] Architectural Alignment — service/repository layers used correctly
- [x] Issue Completion — all stated goals delivered
- [x] No Nitpicking
- [x] Avoid Repetition (first review)
- [x] Iterative Reviews (N/A)
- [ ] Prevent CI Failures — two issues flagged below

---

### Overview

Solid implementation overall. The backend guard + frontend disabled-state layering is the right pattern, and returning the deleted record from `deleteChannel` to fix the tRPC void-return bug is the correct long-term fix (consistent with `deleteServer`). Three issues require changes before merge.

---

### Issue 1 — Voice channel active background lost (visual regression)

**File:** `harmony-frontend/src/components/channel/ChannelSidebar.tsx`

The refactor wraps voice channels in an outer `div` but moves the active-state styling (only `text-white`) to the inner `button`. The outer `div` is missing `BG_ACTIVE` when `connectedChannelId === channel.id`. Before this PR the entire row got `bg-[#3d4148]` when connected; after it doesn't.

**Fix:** Apply `BG_ACTIVE` conditionally on the outer `div`:

```tsx
<div
  className={cn(
    'group flex items-center gap-1.5 rounded px-2 py-1 text-sm transition-colors',
    connectedChannelId === channel.id
      ? cn(BG_ACTIVE, 'hover:bg-[#3d4148]')
      : 'hover:bg-[#393c43]',
  )}
>
```

---

### Issue 2 — Missing `/c/` cache revalidation

**File:** `harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts`

`deleteChannelAction` only invalidates `/channels/${serverSlug}` and `/settings/${serverSlug}`. All comparable actions (`deleteServer`, `updateVisibility`) also call `revalidatePath(`/c/${serverSlug}`, 'layout')`. Without it, the public-channel route tree (`/c/:serverSlug/:channelSlug`) can serve stale Next.js cached data for the deleted channel.

**Fix:** Add the missing revalidation:

```ts
revalidatePath(`/channels/${serverSlug}`, 'layout');
revalidatePath(`/c/${serverSlug}`, 'layout');   // ← missing
revalidatePath(`/settings/${serverSlug}`, 'layout');
```

---

### Issue 3 — Removed null-body safety net in `ApiClient.trpcMutation`

**File:** `harmony-frontend/src/lib/api-client.ts`

The `?? null` that was removed guaranteed Axios always sends a JSON body, even for no-input mutations. With `undefined` passed directly, Axios omits the body entirely. The standalone `trpcMutate` in `trpc-client.ts` correctly uses `JSON.stringify(input ?? {})` to handle this. The `ApiClient` path should be consistent.

**Fix:** Restore the guard, or serialize consistently:

```ts
const res = await this.client.post<TrpcResponse<T>>(`/trpc/${procedure}`, input ?? null);
```

acabrera04:
> Re: comment 4215513001

Fixed in 6d6ac638. Applied the three requested fixes: restored the active background on connected voice-channel rows, added /c/{serverSlug} layout revalidation after channel deletion, and restored the explicit null body for no-input tRPC mutations.

FardeenI (dismissed):
> **Must-fix: Admins can delete channels (backend), but the new UI entrypoints are gated to owner/system-admin only**
>
> Inline notes:
>
> - `harmony-frontend/src/components/channel/ChannelSidebar.tsx:199-206` defines:
>   ```ts
>   const isAdmin =
>     isAuthenticated && (currentUser.isSystemAdmin || currentUser.id === server.ownerId);
>   const hasServerSettingsAccess =
>     isAuthenticated &&
>     (currentUser.isSystemAdmin ||
>       currentUser.id === server.ownerId ||
>       currentUser.role === 'owner' ||
>       currentUser.role === 'admin');
>   ```
>   But the new channel settings gear + “Add channel” are gated by `isAdmin`:
>   - Add text channel: `ChannelSidebar.tsx:338-340`
>   - Text channel settings gear: `ChannelSidebar.tsx:381-387`
>   - Add voice channel: `ChannelSidebar.tsx:418-420`
>   - Voice channel settings gear: `ChannelSidebar.tsx:453-475`
>
> - This conflicts with the actual permission model:
>   - Backend explicitly grants admins `channel:delete` (and create/update) in `harmony-backend/src/services/permission.service.ts:57-68`.
>   - Frontend settings access also allows membership role `admin` in `harmony-frontend/src/app/settings/[serverSlug]/settings-access.ts:42-44` (via `hasServerAdminAccess` at `:18`).
>
> **Impact:** A server “admin” (non-owner) can have permission to delete channels, and can access settings pages if they know the URL, but they won’t see the new sidebar gear icon (nor “add channel”)—so the PR’s stated “admins/owners” UX is not actually delivered for admins.
>
> **Suggested fix:** Use the same access predicate for these controls as the rest of settings/admin UX (e.g., gate the gear icon and `onAdd` on `hasServerSettingsAccess` or an equivalent “canManageChannels” check that includes membership role `admin`).

acabrera04:
> Re: comment 4217279890

Fixed in 59c7609d. Channel add buttons and channel settings links now use the same server settings/admin access predicate that includes server admins, and the sidebar test covers admin vs member visibility for those entrypoints.

acabrera04:
> Re: comment 4215513001

Already addressed. The three requested fixes landed across 6d6ac638 and the current head: active voice-channel row background is restored, /c/{serverSlug} layout revalidation is included after deletion, and no-input tRPC mutations send an explicit null body. CI was also fixed in 84534938 by updating the backend deleteChannel return-value expectations.

harmony-frontend/src/components/settings/ChannelSettingsPage.tsx
```diff
@@ -775,6 +879,13 @@ export function ChannelSettingsPage({
           {effectiveSection === 'notifications' && (
             <ChannelNotificationsSection channel={channel} serverId={channel.serverId} />
           )}
+          {activeSection === 'danger' && (
```
AvanishKulkarni: Minor style inconsistency: all other section guards here use `effectiveSection`, but this one uses `activeSection` directly. In the current code these are always equivalent for `'danger'` (since `effectiveSection` only remaps `'seo' → 'overview'`), so this is functionally harmless. Worth aligning with the pattern in a follow-up to avoid a silent regression if the `effectiveSection` derivation ever expands.

AvanishKulkarni (approved):
## Review

**Guideline Checklist**

- [x] Prioritize Logic over Style — logic is clean; one minor style inconsistency noted inline
- [x] Security First — slug-based ID resolution in the server action is correct defense-in-depth; backend tRPC procedure enforces auth
- [x] Architectural Alignment — service/repository layers used correctly, consistent with `deleteServer` pattern
- [x] Issue Completion — delete flow, last-text-channel guard, and settings gear icons all delivered
- [x] No Nitpicking — only the style inconsistency is noted
- [x] Avoid Repetition — checked both prior reviews before writing
- [x] Iterative Reviews — all prior change requests verified resolved (see below)
- [x] Prevent CI Failures — no issues found

---

### Prior Review Resolutions

**AvanishKulkarni — all three issues resolved ✓**
- Issue 1 (voice channel active BG lost): Fixed — outer `div` applies `BG_ACTIVE` conditionally on `connectedChannelId === channel.id`.
- Issue 2 (missing `/c/` revalidation): Fixed — `revalidatePath(`/c/${serverSlug}`, 'layout')` is present in `deleteChannelAction`.
- Issue 3 (`trpcMutation ?? null`): `?? null` is not removed in this PR; main branch retains it at `api-client.ts:239`. Non-issue.

**FardeenI — `isAdmin` mismatch resolved ✓**
- `isAdmin` variable is removed; gear icons for both text and voice channels are now gated by `hasServerSettingsAccess`, matching the backend permission model. Add-channel buttons are also updated. Fully addressed.

harmony-frontend/src/components/settings/ChannelSettingsPage.tsx
acabrera04: Fixed in c70f4a62. The danger section now uses effectiveSection like the surrounding section guards.

PR #610, opened 2026-05-03
[codex] Fix SEO preview fallback for public channels

### LINKED ISSUE #569: Investigate SEO Preview failure on PUBLIC_INDEXABLE channels and prevent user-facing server errors
## Summary
SEO Preview is failing on a public indexed channel and currently surfaces a production server-component error message in the UI. We should investigate root cause and ensure users see a safe fallback instead of internal error text.

## Environment/Context
- Server visibility: public
- Channel visibility: `PUBLIC_INDEXABLE`
- Location: Channel Settings -> SEO Preview

## Problem
In a valid `PUBLIC_INDEXABLE` context, SEO Preview should render normally. Instead, it throws and exposes a generic production server-component error message to end users, which is poor UX and can reduce trust.

## Expected Behavior
- SEO Preview works for `PUBLIC_INDEXABLE` channels.
- If backend/metadata generation fails, users see a controlled, non-technical fallback message with retry guidance.
- Internal error details are logged server-side for debugging, not displayed to users.

## Investigation Goals
1. Reproduce in local/dev/prod-like environment for `PUBLIC_INDEXABLE` channel.
2. Identify failing boundary (data fetch, auth check, metadata generation, server action, or render path).
3. Confirm whether this regressed recently and which commit/path introduced it.
4. Add targeted telemetry/logging (non-PII) around the failing path to speed diagnosis.

## Acceptance Criteria
1. Root cause identified and documented in issue comments or linked PR.
2. SEO Preview renders successfully for healthy `PUBLIC_INDEXABLE` channels.
3. Failures show a user-safe fallback UI (no raw server-component error text).
4. Error details remain internal (logs/monitoring only) and do not leak to end users.
5. Regression coverage added for both success and graceful-failure paths.

## Repro Note
Observed in production UI at Channel Settings -> SEO Preview on `#general` where the page shows the red server-component error message instead of metadata preview content.

acabrera04:
## Summary

Fixes the Channel Settings SEO Preview failure for public indexed channels that do not yet have persisted generated meta tags.

## Root Cause

The admin SEO Preview endpoint returned 404 when a valid public indexed channel had no generated meta tags row. The settings server action treated that as a hard failure, and the client could surface production server-component error text instead of a controlled preview or fallback state.

## Changes

- Return an ephemeral fallback preview from the admin SEO Preview endpoint when the persisted meta tag row is missing.
- Log a non-PII backend warning for the fallback path.
- Log frontend server-action failures with sanitized metadata and rethrow a generic retry message.
- Render a fixed user-safe retry message for initial SEO Preview load failures.
- Add backend and frontend regressions for the fallback preview and safe-error UI.

## Verification

- harmony-backend: npm test -- admin.metaTag.router.test.ts --runInBand
- harmony-frontend: npm test -- SeoPreviewSection.test.tsx --runInBand
- harmony-backend: npx tsc --noEmit
- harmony-backend: npm run lint
- harmony-frontend: npm run lint

Note: harmony-frontend npx tsc --noEmit is currently blocked by pre-existing unrelated casts in src/__tests__/gifsRoute.test.ts.

Fixes #569

### DIFF
--- a/harmony-backend/src/routes/admin.metaTag.router.ts
+++ b/harmony-backend/src/routes/admin.metaTag.router.ts
@@ -143,7 +143,12 @@ adminMetaTagRouter.get(
       const record = await metaTagRepository.findByChannelId(channelId);
 
       if (!record) {
-        res.status(404).json({ error: 'Meta tags not found for this channel' });
+        logger.warn(
+          { channelId, operation: 'adminSeoPreviewFallback' },
+          'Admin SEO preview missing persisted meta tags; generating fallback preview',
+        );
+        const fallbackPreview = await metaTagService.getFallbackMetaTagsForPreview(channelId);
+        res.json(fallbackPreview);
         return;
       }
 
--- a/harmony-backend/src/routes/admin.metaTag.router.ts
+++ b/harmony-backend/src/routes/admin.metaTag.router.ts
@@ -21,11 +21,13 @@ import { prisma } from '../db/prisma';
 import { redis } from '../db/redis';
 import { createLogger } from '../lib/logger';
 import { auditLogService } from '../services/auditLog.service';
+import { cacheService, sanitizeKeySegment } from '../services/cache.service';
 import type { MetaTagPreview } from '../services/metaTag/types';
 
 const logger = createLogger({ component: 'admin-meta-tag-router' });
 
 const BASE_URL = process.env.BASE_URL ?? 'https://harmony.chat';
+const FALLBACK_PREVIEW_TTL_SECONDS = 60;
 
 // ─── Validation schemas ───────────────────────────────────────────────────────
 
@@ -41,6 +43,10 @@ const IDEMPOTENCY_TTL_SECONDS = 60;
 
 function idempotencyKey(channelId: string, key: string): string {
   return `meta-tag:idempotency:${channelId}:${key}`;
+}
+
+function fallbackPreviewCacheKey(channelId: string): string {
+  return `meta-tag:admin-preview-fallback:${sanitizeKeySegment(channelId)}`;
 }
 
 // ─── Admin authorization middleware ──────────────────────────────────────────
@@ -123,6 +129,30 @@ function buildPreview(
   };
 }
 
+async function getCachedFallbackPreview(channelId: string): Promise<MetaTagPreview> {
+  const cacheKey = fallbackPreviewCacheKey(channelId);
+
+  try {
+    const entry = await cacheService.get<MetaTagPreview>(cacheKey);
+    if (entry && !cacheService.isStale(entry, FALLBACK_PREVIEW_TTL_SECONDS)) {
+      return entry.data;
+    }
+  } catch (err) {
+    logger.warn(
+      { err, channelId, cacheKey },
+      'Failed to read admin SEO fallback preview cache; generating directly',
+    );
+  }
+
+  const fallbackPreview = await metaTagService.getFallbackMetaTagsForPreview(channelId);
+  cacheService
+    .set(cacheKey, fallbackPreview, { ttl: FALLBACK_PREVIEW_TTL_SECONDS })
+    .catch((err) =>
+      logger.warn({ err, channelId, cacheKey }, 'Failed to cache admin SEO fallback preview'),
+    );
+  return fallbackPreview;
+}
+
 // ─── Router ───────────────────────────────────────────────────────────────────
 
 export const adminMetaTagRouter = Router();
@@ -147,7 +177,7 @@ adminMetaTagRouter.get(
           { channelId, operation: 'adminSeoPreviewFallback' },
           'Admin SEO preview missing persisted meta tags; generating fallback preview',
         );
-        const fallbackPreview = await metaTagService.getFallbackMetaTagsForPreview(channelId);
+        const fallbackPreview = await getCachedFallbackPreview(channelId);
         res.json(fallbackPreview);
         return;
       }
--- a/harmony-backend/src/routes/admin.metaTag.router.ts
+++ b/harmony-backend/src/routes/admin.metaTag.router.ts
@@ -21,13 +21,12 @@ import { prisma } from '../db/prisma';
 import { redis } from '../db/redis';
 import { createLogger } from '../lib/logger';
 import { auditLogService } from '../services/auditLog.service';
-import { cacheService, sanitizeKeySegment } from '../services/cache.service';
+import { cacheService, CacheKeys, CacheTTL } from '../services/cache.service';
 import type { MetaTagPreview } from '../services/metaTag/types';
 
 const logger = createLogger({ component: 'admin-meta-tag-router' });
 
 const BASE_URL = process.env.BASE_URL ?? 'https://harmony.chat';
-const FALLBACK_PREVIEW_TTL_SECONDS = 60;
 
 // ─── Validation schemas ───────────────────────────────────────────────────────
 
@@ -43,10 +42,6 @@ const IDEMPOTENCY_TTL_SECONDS = 60;
 
 function idempotencyKey(channelId: string, key: string): string {
   return `meta-tag:idempotency:${channelId}:${key}`;
-}
-
-function fallbackPreviewCacheKey(channelId: string): string {
-  return `meta-tag:admin-preview-fallback:${sanitizeKeySegment(channelId)}`;
 }
 
 // ─── Admin authorization middleware ──────────────────────────────────────────
@@ -130,11 +125,11 @@ function buildPreview(
 }
 
 async function getCachedFallbackPreview(channelId: string): Promise<MetaTagPreview> {
-  const cacheKey = fallbackPreviewCacheKey(channelId);
+  const cacheKey = CacheKeys.metaChannelAdminPreviewFallback(channelId);
 
   try {
     const entry = await cacheService.get<MetaTagPreview>(cacheKey);
-    if (entry && !cacheService.isStale(entry, FALLBACK_PREVIEW_TTL_SECONDS)) {
+    if (entry && !cacheService.isStale(entry, CacheTTL.metaChannelAdminPreviewFallback)) {
       return entry.data;
     }
   } catch (err) {
@@ -146,7 +141,7 @@ async function getCachedFallbackPreview(channelId: string): Promise<MetaTagPrevi
 
   const fallbackPreview = await metaTagService.getFallbackMetaTagsForPreview(channelId);
   cacheService
-    .set(cacheKey, fallbackPreview, { ttl: FALLBACK_PREVIEW_TTL_SECONDS })
+    .set(cacheKey, fallbackPreview, { ttl: CacheTTL.metaChannelAdminPreviewFallback })
     .catch((err) =>
       logger.warn({ err, channelId, cacheKey }, 'Failed to cache admin SEO fallback preview'),
     );
--- a/harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
+++ b/harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
@@ -19,6 +19,7 @@ import {
 } from '@/services/channelService';
 import { getServer, getServerMembersWithRole } from '@/services/serverService';
 import { createFrontendLogger } from '@/lib/frontend-logger';
+import { SEO_PREVIEW_LOAD_ERROR } from '@/lib/seoConstants';
 import type { ServerMemberInfo } from '@/types';
 import type { Channel } from '@/types';
 import { ChannelVisibility } from '@/types';
@@ -121,7 +122,7 @@ export async function fetchSeoPreview(
       operation: 'fetchSeoPreview',
       route: `/settings/${serverSlug}/${channelSlug}`,
     });
-    throw new Error('SEO preview is temporarily unavailable. Please try again in a few minutes.');
+    throw new Error(SEO_PREVIEW_LOAD_ERROR);
   }
 }
 
--- a/harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
+++ b/harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
@@ -18,6 +18,7 @@ import {
   type ChannelMemberEntry,
 } from '@/services/channelService';
 import { getServer, getServerMembersWithRole } from '@/services/serverService';
+import { createFrontendLogger } from '@/lib/frontend-logger';
 import type { ServerMemberInfo } from '@/types';
 import type { Channel } from '@/types';
 import { ChannelVisibility } from '@/types';
@@ -31,6 +32,11 @@ import {
   type MetaTagJobStatus,
   type MetaTagPreview,
 } from '@/services/metaTagAdminService';
+
+const seoPreviewLogger = createFrontendLogger({
+  component: 'seo-preview-actions',
+  feature: 'seo-preview',
+});
 
 export async function saveChannelSettings(
   serverSlug: string,
@@ -107,7 +113,16 @@ export async function fetchSeoPreview(
   channelSlug: string,
 ): Promise<MetaTagPreview> {
   const channel = await resolveChannelForSeo(serverSlug, channelSlug);
-  return getMetaTagPreview(channel.id);
+  try {
+    return await getMetaTagPreview(channel.id);
+  } catch (err) {
+    seoPreviewLogger.error('SEO preview fetch failed', {
+      error: err,
+      operation: 'fetchSeoPreview',
+      route: `/settings/${serverSlug}/${channelSlug}`,
+    });
+    throw new Error('SEO preview is temporarily unavailable. Please try again in a few minutes.');
+  }
 }
 
 export async function saveSeoOverrides(
--- a/harmony-frontend/src/components/settings/SeoPreviewSection.tsx
+++ b/harmony-frontend/src/components/settings/SeoPreviewSection.tsx
@@ -13,6 +13,8 @@ import type { MetaTagJobStatus, MetaTagPreview } from '@/services/metaTagAdminSe
 const TITLE_MAX = 70;
 const DESCRIPTION_MAX = 200;
 const OG_IMAGE_MAX = 500;
+const SEO_PREVIEW_LOAD_ERROR =
+  'SEO preview is temporarily unavailable. Please try again in a few minutes.';
 
 function buildValidationError(value: string, max: number, label: string): string | null {
   if (value.length <= max) return null;
@@ -86,9 +88,9 @@ export function SeoPreviewSection({
         setCustomTitle(nextPreview.customTitle ?? '');
         setCustomDescription(nextPreview.customDescription ?? '');
         setCustomOgImage(nextPreview.customOgImage ?? '');
-      } catch (err) {
+      } catch {
         if (cancelled) return;
-        setError(getUserErrorMessage(err, 'Failed to load SEO preview.'));
+        setError(SEO_PREVIEW_LOAD_ERROR);
       } finally {
         if (!cancelled) setLoading(false);
       }
--- a/harmony-frontend/src/components/settings/SeoPreviewSection.tsx
+++ b/harmony-frontend/src/components/settings/SeoPreviewSection.tsx
@@ -2,6 +2,7 @@
 
 import { useEffect, useRef, useState } from 'react';
 import { getUserErrorMessage, cn } from '@/lib/utils';
+import { SEO_PREVIEW_LOAD_ERROR } from '@/lib/seoConstants';
 import {
   fetchSeoPreview,
   fetchSeoRegenerationStatus,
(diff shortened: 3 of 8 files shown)


harmony-backend/src/routes/admin.metaTag.router.ts
```diff
@@ -143,7 +143,12 @@ adminMetaTagRouter.get(
       const record = await metaTagRepository.findByChannelId(channelId);
 
       if (!record) {
-        res.status(404).json({ error: 'Meta tags not found for this channel' });
+        logger.warn(
+          { channelId, operation: 'adminSeoPreviewFallback' },
+          'Admin SEO preview missing persisted meta tags; generating fallback preview',
+        );
+        const fallbackPreview = await metaTagService.getFallbackMetaTagsForPreview(channelId);
+        res.json(fallbackPreview);
```
AvanishKulkarni, line 176: **Performance concern (follow-up, not a blocker):** `getFallbackMetaTagsForPreview` calls `generateMetaTagsFromContext`, which is presumably an LLM call. This means every admin panel load for a channel without persisted tags triggers a generation call. If an admin opens or refreshes the preview panel before tags have been generated, each load hits the LLM.

Consider caching the fallback result in Redis with a short TTL (e.g. 60 s) keyed on `channelId`, similar to how other meta tag results are cached elsewhere in the service. That would bound cost without changing the visible behaviour.

harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
```diff
@@ -107,7 +113,16 @@ export async function fetchSeoPreview(
...
 ): Promise<MetaTagPreview> {
   const channel = await resolveChannelForSeo(serverSlug, channelSlug);
-  return getMetaTagPreview(channel.id);
+  try {
+    return await getMetaTagPreview(channel.id);
+  } catch (err) {
+    seoPreviewLogger.error('SEO preview fetch failed', {
+      error: err,
+      operation: 'fetchSeoPreview',
+      route: `/settings/${serverSlug}/${channelSlug}`,
```
Comment (AvanishKulkarni) [inline harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts:123]: Good — including the settings route in the log context (`route: \`/settings/${serverSlug}/${channelSlug}\``) makes it straightforward to correlate server-action failures with a specific channel in production logs without leaking user data.

harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
```diff
@@ -107,7 +113,16 @@ export async function fetchSeoPreview(
...
+    });
+    throw new Error('SEO preview is temporarily unavailable. Please try again in a few minutes.');
```
Comment (AvanishKulkarni) [inline harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts]: Correct approach. The raw `err` is logged server-side (above) so the information is not lost, and the rethrown message is safe to surface in the browser. One minor note: the message string here and `SEO_PREVIEW_LOAD_ERROR` in the component are identical — if either drifts the UI will show a mismatched message. Consider exporting the constant from a shared location (e.g. `@/lib/seoConstants`) so both sides stay in sync.

AvanishKulkarni (commented):
## Review

### Overview

Fixes a real user-facing regression cleanly: the admin SEO Preview panel hard-failed with a 404 for valid public channels that had no persisted meta tags. The three-layer fix (backend fallback, server-action error sanitization, component constant) is well-scoped and all three layers have targeted regression tests.

---

### Code Quality

**Backend (`admin.metaTag.router.ts`)** — minimal and correct. Reuses the existing `getFallbackMetaTagsForPreview` path (already in use in `public.router.ts`) rather than introducing new generation logic. Logger placement before the call is appropriate.

**Frontend action (`actions.ts`)** — the try/catch wrapper is the right place to sanitize errors for Next.js Server Actions. Logging the raw error server-side while rethrowing a generic message client-side is the correct pattern.

**Component (`SeoPreviewSection.tsx`)** — dropping `getUserErrorMessage(err, ...)` in favour of the constant is correct since the error is now sanitized upstream. Bare `catch {}` avoids an unused-variable lint warning.

**Tests** — both backend and frontend tests cover exactly the failure paths introduced. The frontend test asserting that the internal Next.js error string does **not** appear in the rendered UI is a nice guard.

---

### Issues

**Pre-existing inconsistency (not introduced here, but now more visible):** The non-fallback success path still uses the local `buildPreview(record)` function which sets `searchPreview.url` to `${BASE_URL}/channels/${record.channelId}` (using the raw `channelId` instead of the canonical slug-based URL). The new fallback path returns the correct `/c/serverSlug/channelSlug` format. Worth a follow-up issue so both paths are consistent.

---

### Verdict

**Approvable** with the LLM caching note (see inline) as a recommended follow-up. The fix is correct, well-tested, and minimal.

harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
Comment (acabrera04) [reply harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts:123]: Acknowledged. No code change needed for this one; the route context remains in the server-side log metadata and does not expose raw backend errors to the browser.

harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts
Comment (acabrera04) [reply harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts]: Fixed in 3c0cde4. Moved the safe SEO preview load message into a shared frontend constant and updated both the server action and component/test to use it.

harmony-backend/src/routes/admin.metaTag.router.ts
acabrera04: Fixed in 3c0cde4. The admin fallback preview now uses a 60-second cache keyed by channelId, with cache read/write failures falling back to direct generation so the preview remains available.

acabrera04:
> Re: comment 4216797974

Thanks. Addressed the inline caching follow-up in 3c0cde4 with a short-lived admin fallback preview cache. I left the noted persisted-preview URL inconsistency out of this PR because it is pre-existing and called out here as a follow-up, not part of this fix.

FardeenI (approved):

1. [x] **Prioritize Logic over Style:** Change addresses a real control-flow failure (404 → unhandled UI error) by returning a valid preview shape for the “no persisted tags yet” case.
2. [x] **Security First:** No secrets exposed; frontend logging is metadata-sanitized (only safe error fields + safe path). No user-facing internal error strings.
3. [x] **Architectural Alignment:** Backend reuses existing `metaTagService.getFallbackMetaTagsForPreview` + shared `cacheService` rather than introducing a parallel generation path.
4. [x] **Issue Completion:** Matches #569 ACs (root cause documented in PR body; safe fallback UI; internal logging; regression coverage added).
5. [x] **No Nitpicking:** Review focused only on behavior, failure modes, and CI risk.
6. [x] **Avoid Repetition:** Prior review already called out the pre-existing preview URL inconsistency; not repeating that here.
7. [x] **Iterative Reviews:** N/A (first review from me); no prior comments from me to re-check.
8. [x] **Prevent CI Failures:** Ran targeted tests locally: backend `admin.metaTag.router.test.ts` + frontend `SeoPreviewSection.test.tsx` both pass.

---

## Bundled Review Comment

The PR fixes the “PUBLIC_INDEXABLE channel with no persisted generated meta-tags row” case end-to-end: backend now returns an ephemeral fallback preview (instead of 404), the server action sanitizes failures before they hit the client, and the component renders a user-safe retry message.

Inline notes:

- **Backend fallback + caching**
  - `harmony-backend/src/routes/admin.metaTag.router.ts:175-183` — missing persisted record now returns a generated fallback preview (previously 404), which resolves the settings panel hard-fail scenario.
  - `harmony-backend/src/routes/admin.metaTag.router.ts:132-154` — brief Redis-backed caching of the fallback preview reduces repeated regeneration work for rapid refreshes.

- **Frontend error sanitization + safe UI**
  - `harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/actions.ts:112-127` — server action logs sanitized metadata and rethrows a generic retry error string, preventing Next server-component error text from surfacing.
  - `harmony-frontend/src/components/settings/SeoPreviewSection.tsx:72-96` — initial load failure now always renders the fixed safe retry message.

- **Regression coverage**
  - `harmony-backend/tests/admin.metaTag.router.test.ts:205-229` — validates fallback preview shape + caching behavior.
  - `harmony-frontend/src/__tests__/SeoPreviewSection.test.tsx:70-81` — asserts the internal server-component error string does not appear in the UI.

Local verification:
- `harmony-backend`: `npm test -- admin.metaTag.router.test.ts --runInBand` ✅
- `harmony-frontend`: `npm test -- SeoPreviewSection.test.tsx --runInBand` ✅

(No `gh pr checks` reported on the branch from GitHub, so CI status isn’t visible from the PR UI.)

harmony-backend/src/routes/admin.metaTag.router.ts
```diff
@@ -21,11 +21,13 @@ import { prisma } from '../db/prisma';
 import { redis } from '../db/redis';
 import { createLogger } from '../lib/logger';
 import { auditLogService } from '../services/auditLog.service';
+import { cacheService, sanitizeKeySegment } from '../services/cache.service';
 import type { MetaTagPreview } from '../services/metaTag/types';
 
 const logger = createLogger({ component: 'admin-meta-tag-router' });
 
 const BASE_URL = process.env.BASE_URL ?? 'https://harmony.chat';
+const FALLBACK_PREVIEW_TTL_SECONDS = 60;
```
AvanishKulkarni: **TTL too short for the cost of this operation.** `getFallbackMetaTagsForPreview` calls `generateMetaTagsFromContext`, the same generation pipeline used for full regeneration jobs. A 60-second TTL means a busy admin reloading the SEO preview panel can re-trigger generation once per minute. Consider raising this to 5–10 minutes (300–600 s), or tying it to whatever TTL constant the regeneration path already uses, to avoid repeatedly paying the generation cost from the admin UI.

harmony-backend/src/routes/admin.metaTag.router.ts
```diff
@@ -43,6 +45,10 @@ function idempotencyKey(channelId: string, key: string): string {
   return `meta-tag:idempotency:${channelId}:${key}`;
 }
 
+function fallbackPreviewCacheKey(channelId: string): string {
+  return `meta-tag:admin-preview-fallback:${sanitizeKeySegment(channelId)}`;
```
AvanishKulkarni: **Cache key convention diverges from the rest of `cacheService`.** All existing key helpers in `cache.service.ts` follow `<domain>:channel:<id>` (e.g. `channel:<id>:visibility`, `meta:channel:<id>`). This key inverts the pattern to `meta-tag:...-fallback:<id>` and lives in the router rather than alongside the other helpers. No collision risk, but it will be confusing when scanning cache keys. Either move it to `cacheService.keys` or add a short comment here explaining why it is intentionally router-local.

harmony-backend/src/routes/admin.metaTag.router.ts
```diff
@@ -123,6 +129,30 @@ function buildPreview(
...
+    logger.warn(
+      { err, channelId, cacheKey },
+      'Failed to read admin SEO fallback preview cache; generating directly',
+    );
+  }
+
+  const fallbackPreview = await metaTagService.getFallbackMetaTagsForPreview(channelId);
+  cacheService
+    .set(cacheKey, fallbackPreview, { ttl: FALLBACK_PREVIEW_TTL_SECONDS })
+    .catch((err) =>
```
AvanishKulkarni, line 145: The fire-and-forget pattern here is acceptable (a cache miss on write failure is fine). Worth being aware: if Redis is unavailable, both the `get` path (line ~136) and the `set` path here will each emit a `warn` on every request, which can produce two warnings per request with no observable operational spike. If that matters for your alerting story, consider a single per-channel backoff flag or a rate-limited log call. Not a blocker.

harmony-frontend/src/components/settings/SeoPreviewSection.tsx
```diff
@@ -86,9 +87,9 @@ export function SeoPreviewSection({
         setCustomTitle(nextPreview.customTitle ?? '');
         setCustomDescription(nextPreview.customDescription ?? '');
         setCustomOgImage(nextPreview.customOgImage ?? '');
-      } catch (err) {
+      } catch {
```
AvanishKulkarni, line 90: Silently discarding the error is correct right now because `fetchSeoPreview` already re-throws a sanitized `Error(SEO_PREVIEW_LOAD_ERROR)` — nothing useful remains. Just note that if `fetchSeoPreview` is ever changed to throw a structured error with extra context (e.g. retry-after, error code), that information will be silently swallowed here. A `console.warn(err)` gated on `process.env.NODE_ENV !== "production"` would future-proof this without leaking to users.

AvanishKulkarni (commented):
## Review Checklist

- [x] **Logic over Style** — Logic is sound; no DRY violations.
- [x] **Security First** — Error messages are sanitized before reaching the client; PII is not logged.
- [x] **Architectural Alignment** — Uses existing `cacheService`, `metaTagService`, and `createFrontendLogger` patterns consistent with the rest of the codebase.
- [x] **Issue Completion** — All 5 acceptance criteria of #569 are addressed: root cause fixed, success path works, safe fallback UI, no error leakage, regression tests added.
- [x] **No Nitpicking** — Review focused on logic/architecture/correctness only.
- [x] **Prevent CI Failures** — Tests updated to match new behavior; imports are present; no syntax issues.

---

## Overview

This PR correctly addresses the root cause: the admin SEO preview endpoint returned `404` when a valid `PUBLIC_INDEXABLE` channel had no persisted `generated_meta_tags` row, causing the settings UI to surface raw server-component error text. The fix adds an ephemeral fallback preview on the backend and sanitizes the error surface on the frontend. The approach is well-scoped. Inline comments below cover the actionable issues.

harmony-backend/src/routes/admin.metaTag.router.ts
AvanishKulkarni: **[Blocking]** 60 seconds is too short given that `getFallbackMetaTagsForPreview` calls `generateMetaTagsFromContext` — the same generation pipeline used for full regeneration jobs. An admin reloading the SEO preview panel can re-trigger the full generation call every minute. Raise this to at least 300 s (5 min) or align it with whatever TTL constant the regeneration path uses, so repeated admin page loads do not repeatedly pay the generation cost.

harmony-backend/src/routes/admin.metaTag.router.ts
AvanishKulkarni: **[Non-blocking]** Cache key pattern diverges from the established `cacheService.keys` convention — all other helpers follow `<domain>:channel:<id>` (e.g. `channel:<id>:visibility`, `meta:channel:<id>`) and live in `cache.service.ts`. This key inverts the structure and is router-local. No collision risk, but it will be confusing when scanning Redis keys. Either move it to `cacheService.keys` or add a brief comment here explaining why it is intentionally kept router-local.

harmony-backend/src/routes/admin.metaTag.router.ts
AvanishKulkarni, line 145: **[Non-blocking]** Fire-and-forget is acceptable here (a cache miss on write failure is fine). Worth noting: if Redis is down, both the `get` path (~line 136) and this `set` path will each emit a `warn` per request — two warnings per request with no observable spike. If your alerting story cares about distinguishing "Redis is flapping" from "normal fallback path", a single structured warn at the entry point (before the `get`) would be cleaner.

harmony-frontend/src/components/settings/SeoPreviewSection.tsx
AvanishKulkarni, line 90: **[Non-blocking]** Silently discarding the error is correct right now because `fetchSeoPreview` already re-throws a sanitized `Error(SEO_PREVIEW_LOAD_ERROR)`. Just note: if the server action is ever changed to throw a structured error with retry-after or an error code, that context will be silently swallowed here. A `process.env.NODE_ENV !== "production" && console.warn(err)` guard would future-proof this without leaking to users.

AvanishKulkarni (changes requested):
## Review Checklist

- [x] **Logic over Style** — Logic is sound; no DRY violations.
- [x] **Security First** — Error messages are sanitized before reaching the client; PII is not logged.
- [x] **Architectural Alignment** — Uses existing `cacheService`, `metaTagService`, and `createFrontendLogger` patterns consistent with the rest of the codebase.
- [x] **Issue Completion** — All 5 acceptance criteria of #569 are addressed: root cause fixed, success path works, safe fallback UI, no error leakage, regression tests added.
- [x] **No Nitpicking** — Review focused on logic/architecture/correctness only.
- [x] **Prevent CI Failures** — Tests updated to match new behavior; imports are present; no syntax issues.

---

## Overview

This PR correctly addresses the root cause: the admin SEO preview endpoint returned `404` when a valid `PUBLIC_INDEXABLE` channel had no persisted `generated_meta_tags` row, causing the settings UI to surface raw server-component error text. The fix adds an ephemeral fallback preview on the backend and sanitizes the error surface on the frontend. The approach is well-scoped.

One blocking change needed (TTL); the rest are observations. See inline comments.

harmony-backend/src/routes/admin.metaTag.router.ts
acabrera04: Fixed in 66bda10. Raised the admin SEO fallback preview cache TTL from 60 seconds to the shared 300-second TTL.

harmony-backend/src/routes/admin.metaTag.router.ts
acabrera04: Fixed in 66bda10. Moved the admin fallback preview key into CacheKeys using the meta:channel:<id>:admin-preview-fallback pattern and added cache-service coverage.

acabrera04: Fixed in 66bda10. Moved the admin fallback preview key into CacheKeys using the meta:channel:<id>:admin-preview-fallback pattern and added cache-service coverage.

harmony-backend/src/routes/admin.metaTag.router.ts
acabrera04: Fixed in 66bda10. Raised the admin SEO fallback preview cache TTL from 60 seconds to the shared 300-second TTL.

harmony-backend/src/routes/admin.metaTag.router.ts
acabrera04: Noted. Leaving the fire-and-forget cache write as-is because cache failures are non-blocking for preview responses; the existing structured warnings still preserve visibility if Redis is unavailable.

acabrera04: Noted. Leaving the fire-and-forget cache write as-is because cache failures are non-blocking for preview responses; the existing structured warnings still preserve visibility if Redis is unavailable.

harmony-frontend/src/components/settings/SeoPreviewSection.tsx
acabrera04: No code change for now. fetchSeoPreview intentionally collapses failures to the sanitized SEO_PREVIEW_LOAD_ERROR before this catch, and adding dev-only console logging would not change production behavior. We can revisit if structured server action errors are introduced.

acabrera04: No code change for now. fetchSeoPreview intentionally collapses failures to the sanitized SEO_PREVIEW_LOAD_ERROR before this catch, and adding dev-only console logging would not change production behavior. We can revisit if structured server action errors are introduced.

acabrera04:
> Re: comment 4217293016

Fixed in 66bda10. Addressed the blocking TTL issue by using a 300-second shared TTL, and addressed the cache-key convention note by moving the fallback preview key into CacheKeys. The remaining inline notes were non-blocking observations and are replied to separately.

acabrera04:
> Re: comment 4217289961

Acknowledged. The follow-up commit 66bda10 addresses the later TTL and cache-key review comments; no additional changes were needed for this review summary.

acabrera04:
> Re: comment 4217291359

Fixed in 66bda10. Addressed the TTL issue by using a 300-second shared TTL, and addressed the cache-key convention note by moving the fallback preview key into CacheKeys. The remaining inline notes were non-blocking observations and are replied to separately.

acabrera04:
> Re: comment 4216797974

Acknowledged. The follow-up commit 66bda10 addresses the later TTL and cache-key comments. The pre-existing success-path URL inconsistency remains outside this PR fix and can be handled as a follow-up.

AvanishKulkarni (approved):
## Review Checklist

- [x] **Logic over Style** — Logic is sound; no DRY violations.
- [x] **Security First** — Error messages are sanitized before reaching the client; PII is not logged.
- [x] **Architectural Alignment** — `CacheKeys` and `CacheTTL` entries now live in `cache.service.ts` alongside the other helpers, consistent with the established convention.
- [x] **Issue Completion** — All 5 acceptance criteria of #569 are satisfied.
- [x] **No Nitpicking** — Review focused on logic/architecture/correctness only.
- [x] **Avoid Repetition** — Checking against prior review comments only.
- [x] **Iterative Reviews** — All prior concerns addressed; resolving them below.
- [x] **Prevent CI Failures** — New `cache.service.test.ts` entries cover the key format and TTL value; no broken imports or syntax issues.

---

## Previous Comments — Resolution

**[Blocking — Resolved]** `FALLBACK_PREVIEW_TTL_SECONDS = 60` → The hard-coded constant is gone. TTL is now `CacheTTL.metaChannelAdminPreviewFallback = 300` (5 minutes), which is appropriate for the cost of the generation call.

**[Non-blocking — Resolved]** Cache key convention → `CacheKeys.metaChannelAdminPreviewFallback` now lives in `cache.service.ts` following the `meta:channel:<id>:...` pattern, and is covered by tests in `cache.service.test.ts`.

**[Non-blocking — Acknowledged]** Fire-and-forget double warn and silent `catch {}` were not changed, which is fine — both were explicitly non-blocking and the current behavior is defensible.

:approve:

PR #613, opened 2026-05-03
Fix password reset required flow

### LINKED ISSUE #514: No password reset flow when account flagged with password reset requirement
## Problem
When a user attempts to sign in with an account that has `passwordHash` in an invalid format (indicating the password needs to be reset), the backend throws an error with the message: `This account must reset its password before signing in.` However, there is no UI flow or endpoint for users to actually reset their password.

## Current Behavior
- Backend rejects login with error message: `This account must reset its password before signing in.`
- User sees this error but has no way to resolve it
- No password reset endpoint exists
- User is stuck unable to sign in

## What Needs to Be Implemented
1. Create a password reset endpoint (likely a tRPC procedure)
2. Create UI flow to handle this specific error state
3. Allow users to reset their password when they encounter this error
4. Consider how admin/staff might flag an account for password reset

## Related Code
- Backend error thrown: `harmony-backend/src/services/auth.service.ts:278`
- Triggers when `passwordHash` is invalid/corrupted during login attempt

## Acceptance Criteria
- [ ] Password reset endpoint exists and works
- [ ] Frontend displays appropriate UI when receiving this error
- [ ] User can complete password reset flow without being signed in
- [ ] User can then sign in with new password

acabrera04:
## Summary

- Adds an unauthenticated password-reset-required endpoint for accounts whose stored password verifier record is invalid.
- Adds login-page recovery UI when the backend returns `This account must reset its password before signing in.`
- Adds backend route/service tests and frontend service/page tests for the reset-required flow.

Closes #514

## Verification

- `npm test -- auth.service.test.ts auth.test.ts` in `harmony-backend` passed.
- `npm test -- authService.test.ts login-password-reset-required.test.tsx` in `harmony-frontend` passed.
- `npm run build` in `harmony-backend` passed.
- `npm run lint` in `harmony-backend` passed with existing warnings.
- `npm run lint` in `harmony-frontend` passed with existing warnings.
- `npm test` in `harmony-frontend` passed: 45 suites, 468 tests.
- `npm run build` in `harmony-frontend` passed.

## Notes

- Full backend Jest requires a configured local `DATABASE_URL` and Redis auth; without those, unrelated database/cache suites fail before exercising this issue.
- Raw `npx tsc --noEmit` in `harmony-frontend` is blocked by existing `src/__tests__/gifsRoute.test.ts` cast errors unrelated to this change; `next build` TypeScript passed.


harmony-backend/src/services/auth.service.ts
```diff
@@ -166,7 +176,11 @@ export const authService = {
   },
 
   async getLoginPasswordSalt(email: string): Promise<string> {
```
AvanishKulkarni, line 178: **Account enumeration via response differentiation**

The `resetRequiredPassword` service currently emits two *distinct* responses for the unauthenticated endpoint:

| Scenario | TRPCError code | HTTP status | Body |
|---|---|---|---|
| Email not found | `UNAUTHORIZED` | 401 | `"Invalid credentials"` |
| Email exists, valid hash | `BAD_REQUEST` | 400 | `"Password reset is not required for this account."` |

An attacker can probe any email against `POST /api/auth/password-reset-required` and distinguish between "this email isn't registered" (401) and "this email is registered with a valid password" (400). The timing-safe `bcrypt.compare` in the not-found path correctly avoids a timing oracle, but the different HTTP status codes and messages still leak registration status.

**Fix:** Return the same error (e.g., UNAUTHORIZED / 401 / `"Invalid credentials"`) for both cases — identical status, identical message body — so there is no information difference regardless of whether the email exists or whether its hash is valid.

```ts
// harmony-backend/src/services/auth.service.ts — resetRequiredPassword
if (!user || decodePasswordVerifierRecord(user.passwordHash)) {
  if (!user) await bcrypt.compare(passwordVerifier, TIMING_DUMMY_HASH);
  throw new TRPCError({ code: 'UNAUTHORIZED', message: 'Invalid credentials' });
}
```

The timing dummy is still needed on the not-found branch to preserve the existing timing protection.

AvanishKulkarni (dismissed):
## Review Checklist

- [x] **Logic over Style** — The `createPasswordVerifierRecord` helper is a clean DRY refactor of the existing inline pattern. Flow is sound end-to-end.
- [ ] **Security First** — Account enumeration vulnerability in `resetRequiredPassword`: distinct HTTP 400 vs. 401 responses reveal whether an email is registered. See inline comment.
- [x] **Architectural Alignment** — REST auth route pattern followed consistently; no tRPC/REST mixing concerns.
- [x] **Issue Completion** — All four acceptance criteria from #514 are met: endpoint exists, frontend UI displays the flow, reset works unauthenticated, and login succeeds afterward.
- [x] **No Nitpicking** — No style-only feedback given.
- [x] **Prevent CI Failures** — The `Prisma` → `PrismaClientKnownRequestError` import change is correct; `user.update` mock additions are complete; tests are verified passing per PR description.

---

## Review

The feature is well-scoped and the implementation is solid. The SRP-style `createPasswordVerifierRecord` helper, the timing-safe not-found path (`bcrypt.compare` against a dummy hash), and the frontend state machine that transitions from login to reset mode are all good work. Test coverage is thorough — unit tests for the service, integration-style route tests, and UI component tests.

**One security issue requires a fix before merge.** The `POST /api/auth/password-reset-required` endpoint leaks whether an email address is registered through its response differentiation:

- Email not found → HTTP **401** + `"Invalid credentials"`
- Email found, valid hash → HTTP **400** + `"Password reset is not required for this account."`

Any caller can enumerate registered email addresses by probing the endpoint, making the 400/401 split an account enumeration oracle. The fix is to unify both into the same UNAUTHORIZED response (see inline comment on `auth.service.ts` for the one-liner change). The test in `auth.service.test.ts` that currently asserts `code: 'BAD_REQUEST'` for the valid-hash case will need to be updated to match the new unified error.

harmony-backend/src/services/auth.service.ts
acabrera04: Fixed in ef0a23d4. resetRequiredPassword now returns the same UNAUTHORIZED / Invalid credentials response when the email is missing and when the account already has a valid verifier, while preserving the timing dummy compare for the missing-user path. Updated the service and route tests to assert the unified response.

AvanishKulkarni (dismissed):
## Review Checklist (Iteration 2)

- [x] **Logic over Style** — No new logic issues.
- [ ] **Security First** — Previous status-code enumeration oracle is fixed. A timing side-channel remains in the valid-hash branch (see inline comment below).
- [x] **Architectural Alignment** — Consistent.
- [x] **Issue Completion** — All acceptance criteria from #514 still satisfied.
- [x] **No Nitpicking** — No style-only feedback.
- [x] **Avoid Repetition** — Not repeating the previous status-code concern (now resolved).
- [x] **Iterative Reviews** — Previous CHANGES_REQUESTED was dismissed; the status-code fix is confirmed.
- [x] **Prevent CI Failures** — Test assertions correctly updated to UNAUTHORIZED/401.

---

## Review

The status-code enumeration oracle from the first review has been correctly resolved — both not-found and valid-hash paths now return identical UNAUTHORIZED/401/"Invalid credentials" responses. That change is good.

**One remaining security issue:** a timing side-channel survives in `resetRequiredPassword`.

In `harmony-backend/src/services/auth.service.ts`, the two error branches have very different execution times:

- **Email not found** → runs `bcrypt.compare(passwordVerifier, TIMING_DUMMY_HASH)` before throwing (~100 ms+)
- **Email found, valid hash** → runs only `decodePasswordVerifierRecord` before throwing (microseconds)

An attacker can probe any email against `POST /api/auth/password-reset-required` and distinguish "this email is registered" (fast 401) from "this email isn't registered" (slow 401) purely from response latency. The fix is a one-liner: add `await bcrypt.compare(passwordVerifier, TIMING_DUMMY_HASH)` before the second throw, mirroring what the not-found branch already does.

acabrera04:
Fixed in 8833cfa7. resetRequiredPassword now runs the same dummy bcrypt compare before rejecting accounts that already have a valid verifier record, matching the missing-user rejection path so the unified 401 response does not retain a timing oracle. I also updated auth.service.test.ts to assert that the reset-not-required branch performs the bcrypt compare before rejecting.

AvanishKulkarni (approved):
## Review Checklist (Iteration 3)

- [x] **Logic over Style** — Clean, well-structured implementation throughout.
- [x] **Security First** — Both enumeration issues fully resolved: status codes are unified, and `bcrypt.compare` with the timing dummy is now called in all UNAUTHORIZED-throwing branches. The new `compareSpy` test verifies the timing protection is present in the valid-hash case.
- [x] **Architectural Alignment** — Consistent with existing REST auth route patterns.
- [x] **Issue Completion** — All four acceptance criteria from #514 satisfied.
- [x] **No Nitpicking** — No style-only feedback.
- [x] **Avoid Repetition** — No previously-raised concerns repeated.
- [x] **Iterative Reviews** — Both prior CHANGES_REQUESTED reviews addressed correctly.
- [x] **Prevent CI Failures** — Tests updated to match new behavior; spy correctly restored after the timing test.

---

Both iteration feedback items are resolved. The timing dummy is now applied uniformly to every branch that throws UNAUTHORIZED in `resetRequiredPassword`, and the test coverage correctly exercises and asserts that. Approved.

PR #614, opened 2026-05-03
Wire frontend message threading

### LINKED ISSUE #503: feat: Wire up message.createReply tRPC to frontend
## Summary
Backend has message.createReply ready, but frontend doesn't expose reply UI.

## Acceptance Criteria
- [ ] Reply UI component created
- [ ] Reply submit calls message.createReply
- [ ] Request includes parentMessageId, content, serverId, channelId
- [ ] Response updates thread list

acabrera04:
## Summary

- Wire created reply responses into channel state so parent reply counts update for composer and SSE replies.
- Propagate the latest reply into open thread views and merge replies without duplicates across tRPC responses and realtime events.
- Add regression coverage for issue #503 threading state updates.

Closes #503

## Verification

- `npm test -- issue-503-threading-frontend.test.tsx issue-563-reply-button-focus.test.tsx --runInBand` (frontend) - passed
- `npm run lint` (frontend) - passed with existing warnings only
- `npm exec tsc -- --noEmit` (backend) - passed
- `npm run lint` (backend) - passed with existing warnings only
- `npm test -- message.replies.test.ts --runInBand` (backend) - blocked by missing `DATABASE_URL` in this isolated worktree
- `npm exec tsc -- --noEmit --project tsconfig.json --pretty false` (frontend) - blocked by existing `src/__tests__/gifsRoute.test.ts` type assertions unrelated to this change


harmony-frontend/src/components/layout/HarmonyShell.tsx
```diff
@@ -278,25 +281,36 @@ export function HarmonyShell({
...
+  const trackCreatedMessage = useCallback((msg: Message) => {
     // Dedup: the SSE event for the sender's own message can arrive before the tRPC
     // response (Redis pub/sub on the same backend + established SSE connection beats
     // the HTTP round-trip). Without this check, the message would be added twice.
-    setLocalMessages(prev => (prev.some(m => m.id === msg.id) ? prev : [...prev, msg]));
-    setReplyingTo(null);
+    setLocalMessages(prev => mergeCreatedMessageIntoChannelMessages(prev, msg));
+    if (msg.parentMessageId) {
+      setLatestReplyByParentId(prev =>
+        prev[msg.parentMessageId!]?.id === msg.id ? prev : { ...prev, [msg.parentMessageId!]: msg },
```
AvanishKulkarni: The `!` assertion works because `msg` is captured in the closure, but TypeScript cannot narrow through the async callback boundary, hence the manual override. A local variable avoids the assertion entirely:

```ts
const parentId = msg.parentMessageId;
if (parentId) {
  setLatestReplyByParentId(prev =>
    prev[parentId]?.id === msg.id ? prev : { ...prev, [parentId]: msg },
  );
}
```

harmony-frontend/src/components/message/MessageItem.tsx
```diff
@@ -676,6 +679,12 @@ export function MessageItem({
   const [isThreadOpen, setIsThreadOpen] = useState(false);
   const [localReplyCount, setLocalReplyCount] = useState(message.replyCount ?? 0);
   const [profileAnchorRect, setProfileAnchorRect] = useState<DOMRect | null>(null);
+  const [prevReplyCount, setPrevReplyCount] = useState(message.replyCount ?? 0);
+  if (prevReplyCount !== (message.replyCount ?? 0)) {
```
AvanishKulkarni, line 764: This render-phase derived-state pattern is correct (equivalent to `getDerivedStateFromProps`), but the interaction with `onReplyCountChange` in `ThreadView` is subtle: when the composer sends a reply, the delta increment from `onReplyCountChange(1)` gets applied first, then this sync fires on the next render and *replaces* `localReplyCount` with the updated prop value. Net result is correct, but a one-liner comment explaining the handoff would prevent someone from removing one of the two paths thinking it is redundant.

harmony-frontend/src/components/message/ThreadView.tsx
```diff
@@ -17,6 +17,11 @@ import type { Message } from '@/types';
 
 // ─── ReplyItem ────────────────────────────────────────────────────────────────
 
+function appendUniqueReplies(base: Message[], incoming: Message[]) {
```
AvanishKulkarni: Consider moving `appendUniqueReplies` to `lib/message-threading.ts` alongside `mergeCreatedMessageIntoChannelMessages`. Both are pure dedup utilities on `Message[]` — co-locating them would make `appendUniqueReplies` directly unit-testable without needing to render `ThreadView`.

AvanishKulkarni (commented):
## Overview

This PR correctly fixes the threading state update bugs from #503: reply counts now increment exactly once whether the reply comes from the composer or SSE, `ThreadView` merges incoming replies without duplicates, and `localReplyCount` in `MessageItem` stays in sync with the underlying prop. The implementation is focused and minimal — good scope discipline.

A few targeted suggestions below.

### Double-count interaction (non-obvious but not a bug)

When the thread composer sends a reply, `ThreadView.handleReplyCreated` fires `onReplyCountChange(1)` which bumps `localReplyCount` by delta. Shortly after, `trackCreatedMessage` → `mergeCreatedMessageIntoChannelMessages` updates `message.replyCount` on the prop. The render-phase `prevReplyCount` sync then **replaces** `localReplyCount` with the prop value, so the delta increment is effectively overwritten. Net result is correct, but the two-path handoff is fragile — a short comment would prevent a future contributor from removing one path thinking the other makes it redundant.

### `appendUniqueReplies` placement

The function is a pure dedup utility on `Message[]`, parallel in role to `mergeCreatedMessageIntoChannelMessages`. Both belong in `lib/message-threading.ts` where they can be co-tested. Leaving it private to `ThreadView.tsx` means it is currently untested in isolation.

### Tests

The three new test cases cover the main scenarios well. Gaps worth noting:
- No test for the `onReplyCountChange` + render-phase sync interaction.
- `appendUniqueReplies` is untested in isolation; moving it to the shared lib would fix this automatically.

harmony-frontend/src/components/layout/HarmonyShell.tsx
acabrera04: Fixed in 90490f8d. Replaced the non-null assertion with a local `parentId` so the callback no longer needs the manual TypeScript override.

harmony-frontend/src/components/message/MessageItem.tsx
acabrera04: Fixed in 90490f8d. Added a short comment explaining that the optimistic reply-count delta is replaced by the authoritative parent count once channel state catches up.

harmony-frontend/src/components/message/ThreadView.tsx
acabrera04: Fixed in 90490f8d. Moved `appendUniqueReplies` into `lib/message-threading.ts` and added direct unit coverage for duplicate filtering.

harmony-frontend/src/components/message/ThreadView.tsx
```diff
@@ -13,6 +13,7 @@ import { formatMessageTimestamp } from '@/lib/utils';
 import { getThreadMessagesAction } from '@/app/actions/getThreadMessages';
```
AvanishKulkarni, line 13: **Race condition — double-count when SSE arrives before `createReplyAction` resolves.**

`handleReplyCreated` calls `onReplyCountChange?.(1)` to bump `localReplyCount` in `MessageItem`. But `trackCreatedMessage` in `HarmonyShell` now *also* bumps the parent count via `mergeCreatedMessageIntoChannelMessages` on every SSE event. The `prevReplyCount` sync in `MessageItem` is designed to correct an optimistic bump once the authoritative value arrives — but it only triggers when `message.replyCount` *changes*. If the SSE fires first (which is plausible since Redis pub/sub on the same host often beats the HTTP response), the sync runs and locks in `localReplyCount = 1`, then `onReplyCountChange(1)` adds another 1 → display shows 2, stuck.

Remove the `onReplyCountChange?.(1)` call here. The count is now owned centrally by `mergeCreatedMessageIntoChannelMessages`; `ThreadView` no longer needs to push its own delta.

AvanishKulkarni (changes requested):

- [x] **1. Prioritize Logic over Style** — Logic is clean; utility functions are well-factored. One logic correctness issue found (see below).
- [x] **2. Security First** — No new surface area; utilities are pure functions with no I/O.
- [x] **3. Architectural Alignment** — Follows existing patterns (render-phase derived state, SSE handler delegation).
- [x] **4. Issue Completion** — All six acceptance criteria for #503 are satisfied: reply UI exists, `createReply`/`getThreadMessages` are wired, thread list updates, pagination dedup is correct, required fields are present.
- [ ] **5. No Nitpicking / 6. Avoid Repetition / 7. Iterative Reviews** — The three previous concerns (non-null assertion, derived-state comment, move `appendUniqueReplies`) were all addressed in commit `90490f8d`. One new correctness issue found that was not raised before.
- [x] **8. Prevent CI Failures** — No syntax errors, missing imports, or obvious test breakage.

---

## Review

The overall approach is sound — centralising dedup logic in `message-threading.ts` and sharing a single `trackCreatedMessage` path for both the composer and SSE is the right design. The previous concerns have been resolved. One new correctness bug is present.

### `harmony-frontend/src/components/message/ThreadView.tsx` – double-count race when SSE beats the tRPC response

`ThreadView.handleReplyCreated` (line 240) still calls `onReplyCountChange?.(1)`, which adds 1 to `localReplyCount` in `MessageItem`. At the same time, `trackCreatedMessage` in `HarmonyShell` now bumps the parent`s `replyCount` in `localMessages` via `mergeCreatedMessageIntoChannelMessages`, so the `prevReplyCount` sync in `MessageItem` will eventually set `localReplyCount` to the authoritative value.

This works correctly only when the tRPC HTTP response arrives *before* the SSE event. The race goes wrong in the other order:

```
1. Backend creates reply → emits Redis pub/sub event
2. SSE arrives: message.replyCount becomes 1
3. MessageItem re-renders: prevReplyCount(0) ≠ 1 → sync sets localReplyCount = 1
4. createReplyAction() resolves → onReplyCountChange(1) → localReplyCount = 2  ← WRONG
5. message.replyCount stays 1, prevReplyCount is already 1 → no further sync
```

The SSE-first path is realistic: Redis pub/sub on the same backend instance often delivers before the HTTP round-trip completes.

**Fix:** Remove the `onReplyCountChange?.(1)` call from `ThreadView.handleReplyCreated`. The parent count is now managed authoritatively through `mergeCreatedMessageIntoChannelMessages` triggered by both the SSE path and the `handleMessageSent` path. The `onReplyCountChange` prop itself can stay for other callers, but `ThreadView` should no longer push a delta that races with the centralised count update.

harmony-frontend/src/components/message/ThreadView.tsx
acabrera04: Fixed in b126f378. Removed ThreadView's local onReplyCountChange delta path so parent reply counts are owned by the centralized merge/SSE state, avoiding the SSE-before-createReplyAction double-count race.

acabrera04:
> Re: comment 4217370939

Fixed in b126f378. Removed the remaining ThreadView reply-count delta so the parent count is updated only through the centralized merge/SSE path; this closes the SSE-first double-count race described in the review.

acabrera04:
> Re: comment 4216798758

Addressed. The earlier suggestions in this review were handled in 90490f8d by moving appendUniqueReplies into the shared message-threading helper and documenting/syncing the derived reply-count state. The remaining fragile double-count handoff is now removed in b126f378, so ThreadView no longer pushes a local count delta at all.

harmony-frontend/src/components/message/ThreadView.tsx
AvanishKulkarni, line 13: **Minor: `handleReplySent` raw-appends without dedup, creating a potential duplicate in the rare SSE-first race.**

`handleReplySent` does `setReplies(prev => [...prev, reply])`. With `incomingReply` now wired, the same reply can arrive via two paths when the SSE fires before `createReplyAction` resolves:

1. SSE → `incomingReply` → `appendUniqueReplies` appends reply
2. `createReplyAction` resolves → `handleReplySent` → raw append → duplicate

`appendUniqueReplies` already handles this. Consider:

```ts
const handleReplySent = useCallback((reply: Message) => {
  setReplies(prev => appendUniqueReplies(prev, [reply]));
  setIsComposerOpen(false);
}, []);
```

Not blocking since the race is rare in practice, but it is the same family of issue as the one just fixed.

AvanishKulkarni (approved):

- [x] **1. Prioritize Logic over Style** — Logic is sound; utilities are clean and well-named.
- [x] **2. Security First** — No new attack surface.
- [x] **3. Architectural Alignment** — Follows existing patterns consistently.
- [x] **4. Issue Completion** — All six #503 acceptance criteria satisfied.
- [x] **5. No Nitpicking** — Only substantive concerns raised.
- [x] **6. Avoid Repetition** — No previously-raised comments duplicated.
- [x] **7. Iterative Reviews** — My prior concern (`onReplyCountChange` race) was fully resolved: the prop was removed from `ThreadViewProps`, the raw delta call in `handleReplySent` was dropped, and `incomingReply` is the single canonical path for external replies. Previous inline comments can be resolved.
- [x] **8. Prevent CI Failures** — No syntax errors, missing imports, or test breakage.

---

## Review

The requested change was fully addressed. `onReplyCountChange` is gone from the interface and call-site; the parent reply count is now owned exclusively by `mergeCreatedMessageIntoChannelMessages` through the `trackCreatedMessage` path. The race condition I flagged is closed.

One minor follow-up is worth fixing while the code is in motion.

harmony-frontend/src/components/message/ThreadView.tsx
acabrera04: Fixed in f9e3e4fb. ThreadView.handleReplySent now uses appendUniqueReplies(prev, [reply]) so an SSE-first incomingReply and the later createReplyAction response cannot add the same reply twice.

acabrera04:
> Re: comment 4217423011

Fixed in f9e3e4fb. Applied the suggested appendUniqueReplies path in handleReplySent, so the thread list is deduped whether the SSE incomingReply or createReplyAction response arrives first.

PR #615, opened 2026-05-03
[codex] Fix cloud integration public fixture failures

### LINKED ISSUE #592: Cloud Integration Tests Failure Tracker
Tracks failures from `.github/workflows/run-cloud-integration-tests.yml`.

acabrera04:
## Summary

- Fix cloud fixture discovery so JSON-LD assertions only target `PUBLIC_INDEXABLE` channels.
- Stop public message responses from selecting or emitting attachment `sizeBytes` BigInts.
- Add regressions for both the cloud fixture selection and JSON-safe public message serialization.

## Root Cause

Issue #592 was opened by the cloud integration failure tracker after GPC-2 selected a guest-accessible channel that was not guaranteed to be `PUBLIC_INDEXABLE`, so the page correctly omitted JSON-LD. While verifying the same cloud subset, GPC-5 also exposed that public message pages with attachments can return HTTP 500 because Express cannot JSON-serialize Prisma BigInt attachment sizes.

Closes #592.

## Checks

- `npm test -- --runTestsByPath env.test.ts` in `tests/integration`
- `npm test -- --runTestsByPath tests/public.router.test.ts` in `harmony-backend`
- `npm run lint` in `harmony-backend` (passes with two existing warnings in unrelated tests)
- Live cloud subset against the issue URLs: JSON-LD assertions pass with this fixture fix; deployed GPC-5 remains red until this backend fix is deployed.

### DIFF
--- a/harmony-backend/src/routes/public.router.ts
+++ b/harmony-backend/src/routes/public.router.ts
@@ -11,6 +11,36 @@ import { inviteService } from '../services/invite.service';
 
 const logger = createLogger({ component: 'public-router' });
 
+const PUBLIC_ATTACHMENT_SELECT = {
+  id: true,
+  url: true,
+  filename: true,
+  contentType: true,
+} as const;
+
+type PublicMessageRecord = {
+  attachments?: Array<{
+    id: string;
+    url: string;
+    filename: string;
+    contentType: string;
+  }>;
+};
+
+function toJsonSafePublicMessage<T extends PublicMessageRecord>(message: T) {
+  const { attachments = [], ...rest } = message;
+
+  return {
+    ...rest,
+    attachments: attachments.map((attachment) => ({
+      id: attachment.id,
+      url: attachment.url,
+      filename: attachment.filename,
+      contentType: attachment.contentType,
+    })),
+  };
+}
+
 /**
  * Factory so createApp() can inject a rate-limit store (e.g. a mock in tests
  * or a RedisStore in production) without requiring a real Redis connection in
@@ -63,13 +93,13 @@ export function createPublicRouter(store?: Store) {
             editedAt: true,
             author: { select: { id: true, username: true } },
             attachments: {
-              select: { id: true, url: true, filename: true, contentType: true, sizeBytes: true },
+              select: PUBLIC_ATTACHMENT_SELECT,
             },
           },
         });
 
         res.set('Cache-Control', `public, max-age=${CacheTTL.channelMessages}`);
-        res.json({ messages, page, pageSize });
+        res.json({ messages: messages.map(toJsonSafePublicMessage), page, pageSize });
       } catch (err) {
         logger.error({ err, channelId: req.params.channelId }, 'Public messages route failed');
         if (!res.headersSent) {
@@ -115,7 +145,7 @@ export function createPublicRouter(store?: Store) {
             editedAt: true,
             author: { select: { id: true, username: true } },
             attachments: {
-              select: { id: true, url: true, filename: true, contentType: true, sizeBytes: true },
+              select: PUBLIC_ATTACHMENT_SELECT,
             },
           },
         });
@@ -126,7 +156,7 @@ export function createPublicRouter(store?: Store) {
         }
 
         res.set('Cache-Control', `public, max-age=${CacheTTL.channelMessages}`);
-        res.json(message);
+        res.json(toJsonSafePublicMessage(message));
       } catch (err) {
         logger.error(
           { err, channelId: req.params.channelId, messageId: req.params.messageId },
@@ -247,7 +277,9 @@ export function createPublicRouter(store?: Store) {
         const channels = await prisma.channel.findMany({
           where: {
             serverId: server.id,
-            visibility: { in: [ChannelVisibility.PUBLIC_INDEXABLE, ChannelVisibility.PUBLIC_NO_INDEX] },
+            visibility: {
+              in: [ChannelVisibility.PUBLIC_INDEXABLE, ChannelVisibility.PUBLIC_NO_INDEX],
+            },
           },
           orderBy: { position: 'asc' },
           select: { id: true, name: true, slug: true, type: true, topic: true },
--- a/tests/integration/env.ts
+++ b/tests/integration/env.ts
@@ -122,6 +122,7 @@ async function resolveIndexableChannelsForServer(
 ): Promise<string[]> {
   const publicChannels: string[] = [];
 
+  // Sequential fetch: stop as soon as we have 3 PUBLIC_INDEXABLE channels.
   for (const candidate of channelCandidates) {
     if (!candidate.slug) continue;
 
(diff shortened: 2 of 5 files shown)


harmony-backend/src/routes/public.router.ts
```diff
@@ -11,6 +11,36 @@ import { inviteService } from '../services/invite.service';
...
+  }>;
+};
+
+function toJsonSafePublicMessage<T extends PublicMessageRecord>(message: T) {
+  const { attachments = [], ...rest } = message;
+
+  return {
+    ...rest,
+    attachments: attachments.map((attachment) => ({
+      id: attachment.id,
```
AvanishKulkarni, line 36: Nit: the explicit field-by-field projection inside `attachments.map(...)` is redundant — `PUBLIC_ATTACHMENT_SELECT` already excludes `sizeBytes`, so the Prisma result already only contains these four fields. Could simplify to `{ ...rest, attachments }` and be equally safe. Keeping it explicit is fine as a defensive guard, just worth noting the duplication.

tests/integration/env.ts
```diff
@@ -116,6 +116,33 @@ type DiscoveredServerFixture = {
...
 let cloudFixturePromise: Promise<CloudFixture> | null = null;
 
+async function resolveIndexableChannelsForServer(
+  serverSlug: string,
+  channelCandidates: Array<{ slug?: string }>,
+): Promise<string[]> {
+  const publicChannels: string[] = [];
+
+  for (const candidate of channelCandidates) {
+    if (!candidate.slug) continue;
```
AvanishKulkarni, line 127: The sequential `await fetch` inside the loop is intentional — the early `break` at 3 channels avoids unnecessary API calls. Worth a short comment here (e.g., `// sequential fetch: stop as soon as we have 3 PUBLIC_INDEXABLE channels`) so future readers do not reflexively parallelize this with Promise.all and lose the early-exit behaviour.

AvanishKulkarni (commented):
## PR #615 Review — Fix cloud integration public fixture failures

**Overview:** Fixes two independent bugs caught by cloud integration tests: (1) fixture discovery selecting non-`PUBLIC_INDEXABLE` channels as JSON-LD test targets, and (2) Express crashing with a BigInt serialization error when public message endpoints returned attachment `sizeBytes`. All CI checks pass. ✅

---

### Code Quality

**`harmony-backend/src/routes/public.router.ts`**

- `PUBLIC_ATTACHMENT_SELECT` constant is a clean DRY extraction — eliminates the repetition between the paginated and single-message endpoints.
- `toJsonSafePublicMessage` is a well-named, narrowly scoped helper. The `PublicMessageRecord` type constraint is correct and safe.
- Root cause fix (dropping `sizeBytes` from the Prisma select) is the right approach over a serialization workaround (e.g., a `JSON.stringify` replacer). Fixing at the data layer is correct.

**`tests/integration/env.ts`**

- `resolveIndexableChannelsForServer` correctly moves the visibility check from "trust the channel list" to "fetch each channel and verify." This is the right fix.
- Error message improvement (`'no PUBLIC_INDEXABLE public server/channel pair is available'`) is a good touch for debuggability.

---

### Test Coverage

- Both regression tests for the BigInt fix correctly verify that `sizeBytes` is excluded and the four expected fields are present.
- `MESSAGE_WITH_ATTACHMENT` fixture correctly uses `BigInt(1234)` to simulate the real Prisma type.
- The new `env.test.ts` test (`'skips PUBLIC_NO_INDEX channels when choosing JSON-LD cloud fixture targets'`) covers the fixture skipping logic well.
- Updated `fetchMock` call counts in existing tests are correct — each now accounts for the extra per-channel visibility probes.

---

### Minor Notes

See inline comments below. Neither is a blocker.

**Recommendation: Approve — the fix is correct, minimal, and well-tested.**

harmony-backend/src/routes/public.router.ts
acabrera04: Won't fix: I kept the explicit projection as a defensive JSON-safety guard. I tried the simplification, but `harmony-backend/tests/public.router.test.ts` intentionally injects a BigInt `sizeBytes` into the mocked Prisma result to verify these public routes cannot serialize unsafe attachment fields; the simplified helper regressed those tests. The current explicit map preserves that safety net.

tests/integration/env.ts
acabrera04: Fixed in 13add39e. Added the sequential-fetch comment above the loop so future changes preserve the early-exit behavior after discovering three PUBLIC_INDEXABLE channels.

FardeenI (approved):
### PR Review Checklist (per `.github/PR_REVIEW.md`)
1. [x] **Prioritize Logic over Style** — Changes are narrowly targeted to two concrete failures (fixture selection + BigInt serialization) with regression coverage.
2. [x] **Security First** — Fix removes a real 500/DoS vector (Express JSON serialization on BigInt) on unauthenticated public endpoints; no new sensitive exposure spotted.
3. [x] **Architectural Alignment** — Uses existing public-router + integration env fixture discovery patterns; no new architectural divergence introduced.
4. [x] **Issue Completion** — PR explicitly addresses #592’s tracked cloud failures (indexable-channel selection + public message 500) and adds regressions.
5. [x] **No Nitpicking** — No style-only concerns raised.
6. [x] **Avoid Repetition** — Prior review comments were minor/nits and already addressed/acknowledged; not repeating them here.
7. [x] **Iterative Reviews** — New commit documents the sequential-fetch intent; prior feedback appears handled.
8. [x] **Prevent CI Failures** — No obvious compile/runtime breaks in the diff; tests added/updated to cover regressions. (Note: `gh pr checks` reports no checks on the branch—see review note.)

---

## Bundled Review (single comment)

This PR fixes two high-impact cloud/public regressions and backs them with focused tests.

Inline notes:

> `harmony-backend/src/routes/public.router.ts:~11-165`  
By removing `sizeBytes` from the Prisma `attachments` select and wrapping responses through `toJsonSafePublicMessage(...)`, the public message endpoints should no longer be able to emit BigInt values that crash `res.json()` with HTTP 500. This is the right layer to fix it (prevent unsafe fields from being selected/returned rather than trying to stringify BigInts).

> `tests/integration/env.ts:~116-547`  
Cloud fixture discovery now verifies channel `visibility === 'PUBLIC_INDEXABLE'` via `GET /api/public/servers/:serverSlug/channels/:channelSlug` before selecting JSON-LD targets, which matches the actual contract (the channel list endpoint intentionally doesn’t expose visibility). This should stop the fixture from picking `PUBLIC_NO_INDEX` channels and then failing JSON-LD assertions.

> `harmony-backend/tests/public.router.test.ts` + `tests/integration/env.test.ts`  
The regressions are well-aimed: mocked `BigInt(sizeBytes)` ensures we don’t reintroduce JSON-unsafe public responses, and the cloud env tests cover skipping `PUBLIC_NO_INDEX` during fixture selection.

CI note (process): `gh pr checks 615` reports “no checks reported” for this branch. Before merging, please ensure the expected workflows have run/passed (or that required checks will gate the merge).

PR #616, opened 2026-05-03
feat: add guest message search for public channels

### LINKED ISSUE #486: feat: Message search for guests in public channels
## Summary
Guests viewing public channels should be able to search messages using a search UI component.

## Expected Behavior
- Guests can search messages by keyword in public channels
- Search results displayed in a sidebar or modal
- Only messages in the current channel are searchable
- Search is case-insensitive

## Acceptance Criteria
- [ ] Full-text search API endpoint created (backend)
- [ ] Search UI component added to GuestChannelView
- [ ] Search results display with author, timestamp, and context
- [ ] Guest search is rate-limited appropriately

acabrera04:
## Summary
- add a guest-accessible public channel message search endpoint scoped to the current channel
- add guest search UI in public channel headers with author, timestamp, and context results
- add focused backend and frontend regression coverage

Closes #486

## Verification
- npm --prefix harmony-backend test -- public.router.test.ts --runInBand
- npm --prefix harmony-backend run lint
- npm --prefix harmony-frontend test -- publicMessageSearchService.test.ts GuestMessageSearch.test.tsx --runInBand
- npm --prefix harmony-frontend run lint
- npm --prefix harmony-frontend run build

## Notes
- Full backend test suite was attempted but local DB/Redis-dependent tests require DATABASE_URL and Redis auth; the issue-relevant backend public router suite passed.
- Full frontend tsc --noEmit still reports unrelated existing errors in src/__tests__/gifsRoute.test.ts; next build TypeScript passed.

### DIFF
--- a/harmony-backend/src/routes/public.router.ts
+++ b/harmony-backend/src/routes/public.router.ts
@@ -10,6 +10,35 @@ import { metaTagService } from '../services/metaTag/metaTagService';
 import { inviteService } from '../services/invite.service';
 
 const logger = createLogger({ component: 'public-router' });
+const DEFAULT_SEARCH_LIMIT = 20;
+const MAX_SEARCH_LIMIT = 50;
+const MAX_SEARCH_QUERY_LENGTH = 120;
+
+function parseSearchQuery(value: unknown): string {
+  const raw = Array.isArray(value) ? value[0] : value;
+  return typeof raw === 'string' ? raw.trim().slice(0, MAX_SEARCH_QUERY_LENGTH) : '';
+}
+
+function parseSearchLimit(value: unknown): number {
+  const raw = Array.isArray(value) ? value[0] : value;
+  const parsed = typeof raw === 'string' ? Number(raw) : Number(raw);
+  if (!Number.isFinite(parsed)) return DEFAULT_SEARCH_LIMIT;
+  return Math.min(MAX_SEARCH_LIMIT, Math.max(1, Math.floor(parsed)));
+}
+
+function buildSearchContext(content: string, query: string): string {
+  const normalizedContent = content.toLowerCase();
+  const normalizedQuery = query.toLowerCase();
+  const matchIndex = normalizedContent.indexOf(normalizedQuery);
+  if (matchIndex === -1) return content.length > 180 ? `${content.slice(0, 177)}...` : content;
+
+  const radius = 70;
+  const start = Math.max(0, matchIndex - radius);
+  const end = Math.min(content.length, matchIndex + query.length + radius);
+  const prefix = start > 0 ? '...' : '';
+  const suffix = end < content.length ? '...' : '';
+  return `${prefix}${content.slice(start, end)}${suffix}`;
+}
 
 /**
  * Factory so createApp() can inject a rate-limit store (e.g. a mock in tests
@@ -78,6 +107,65 @@ export function createPublicRouter(store?: Store) {
       }
     },
   );
+
+  /**
+   * GET /api/public/channels/:channelId/messages/search?q=<keyword>
+   * Searches non-deleted messages within a single guest-accessible public channel.
+   */
+  router.get('/channels/:channelId/messages/search', async (req: Request, res: Response) => {
+    try {
+      const { channelId } = req.params;
+      const query = parseSearchQuery(req.query.q);
+      const limit = parseSearchLimit(req.query.limit);
+
+      if (!query) {
+        res.status(400).json({ error: 'Search query is required' });
+        return;
+      }
+
+      const channel = await prisma.channel.findUnique({
+        where: { id: channelId },
+        select: { id: true, visibility: true },
+      });
+
+      if (!channel || channel.visibility === ChannelVisibility.PRIVATE) {
+        res.status(404).json({ error: 'Channel not found' });
+        return;
+      }
+
+      const messages = await prisma.message.findMany({
+        where: {
+          channelId,
+          isDeleted: false,
+          content: { contains: query, mode: 'insensitive' },
+        },
+        orderBy: { createdAt: 'desc' },
+        take: limit,
+        select: {
+          id: true,
+          content: true,
+          createdAt: true,
+          editedAt: true,
+          author: { select: { id: true, username: true } },
+        },
+      });
+
+      res.set('Cache-Control', 'no-store');
+      res.json({
+        results: messages.map((message) => ({
+          ...message,
+          context: buildSearchContext(message.content, query),
+        })),
+        query,
+        limit,
+      });
+    } catch (err) {
+      logger.error({ err, channelId: req.params.channelId }, 'Public message search route failed');
+      if (!res.headersSent) {
+        res.status(500).json({ error: 'Internal server error' });
+      }
+    }
+  });
 
   /**
    * GET /api/public/channels/:channelId/messages/:messageId
@@ -247,7 +335,9 @@ export function createPublicRouter(store?: Store) {
         const channels = await prisma.channel.findMany({
           where: {
             serverId: server.id,
-            visibility: { in: [ChannelVisibility.PUBLIC_INDEXABLE, ChannelVisibility.PUBLIC_NO_INDEX] },
+            visibility: {
+              in: [ChannelVisibility.PUBLIC_INDEXABLE, ChannelVisibility.PUBLIC_NO_INDEX],
+            },
           },
           orderBy: { position: 'asc' },
           select: { id: true, name: true, slug: true, type: true, topic: true },
--- a/harmony-backend/src/routes/public.router.ts
+++ b/harmony-backend/src/routes/public.router.ts
@@ -21,7 +21,7 @@ function parseSearchQuery(value: unknown): string {
 
 function parseSearchLimit(value: unknown): number {
   const raw = Array.isArray(value) ? value[0] : value;
-  const parsed = typeof raw === 'string' ? Number(raw) : Number(raw);
+  const parsed = Number(raw);
   if (!Number.isFinite(parsed)) return DEFAULT_SEARCH_LIMIT;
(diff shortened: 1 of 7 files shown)


harmony-backend/src/routes/public.router.ts
```diff
@@ -10,6 +10,35 @@ import { metaTagService } from '../services/metaTag/metaTagService';
...
 const logger = createLogger({ component: 'public-router' });
+const DEFAULT_SEARCH_LIMIT = 20;
+const MAX_SEARCH_LIMIT = 50;
+const MAX_SEARCH_QUERY_LENGTH = 120;
+
+function parseSearchQuery(value: unknown): string {
+  const raw = Array.isArray(value) ? value[0] : value;
+  return typeof raw === 'string' ? raw.trim().slice(0, MAX_SEARCH_QUERY_LENGTH) : '';
+}
+
```
AvanishKulkarni, line 21: **[Blocker — not a logic error, but missing feature]** The new search route is not covered by the rate-limit middleware. Other public routes in this router use the injected `store` to throttle unauthenticated traffic, but this handler is inserted outside that scope. Since this is an unauthenticated endpoint that hits the DB on every request, an unthrottled caller can trivially cause load spikes. Issue #486 explicitly lists _"Guest search is rate-limited appropriately"_ as an acceptance criterion — this needs to be wired up before the issue can be closed.

AvanishKulkarni (commented):
## Review Checklist

- [x] **Prioritize Logic over Style:** Core search logic (debounce, cancel-on-unmount, context windowing) is sound and non-redundant.
- [x] **Security First:** Input is sanitized at both the service layer (client-side trim) and backend (`parseSearchQuery` with length cap). Channel visibility is verified before any DB query.
- [ ] **Architectural Alignment:** The new route is added inline in `public.router.ts` without a dedicated service layer — minor deviation from the service-layer pattern used elsewhere.
- [ ] **Issue Completion:** One acceptance criterion is **not met**: _"Guest search is rate-limited appropriately"_ — the new endpoint is placed outside the rate-limited middleware scope. See inline comment.
- [x] **No Nitpicking:** Review is focused on logic and correctness only.
- [x] **Avoid Repetition:** First review on this PR.
- [x] **Iterative Reviews:** First review on this PR.
- [x] **Prevent CI Failures:** No syntax errors, imports are correct, tests cover the new paths.

---

The PR is well-structured and covers the happy path thoroughly. The **missing rate-limit** on the new search endpoint is the only blocker — it is an explicit acceptance criterion of #486 and an unauthenticated, DB-hitting endpoint with no throttle is a DoS risk. See inline comments for details on each finding.

harmony-backend/src/routes/public.router.ts
```diff
@@ -10,6 +10,35 @@ import { metaTagService } from '../services/metaTag/metaTagService';
...
+function parseSearchLimit(value: unknown): number {
```
AvanishKulkarni, line 22: **[Minor]** Both branches of the ternary produce the same result — `Number(raw)` either way. The type-guard is vestigial and slightly misleading:
```ts
// current
const parsed = typeof raw === string ? Number(raw) : Number(raw);

// fix — just coerce directly
const parsed = Number(raw);
```
If the intent was to short-circuit a numeric `raw` without coercion, the second branch should be `raw as number`.

harmony-backend/src/routes/public.router.ts
```diff
@@ -79,6 +108,65 @@ export function createPublicRouter(store?: Store) {
     },
```
AvanishKulkarni, line 108: **[Blocker — unmet acceptance criterion]** This route is not covered by the rate-limit middleware. Other public routes in this router use the injected `store` to throttle unauthenticated traffic, but this handler sits outside that scope. An unauthenticated caller can drive unbounded DB queries with no throttle.

Issue #486 explicitly lists _"Guest search is rate-limited appropriately"_ as an acceptance criterion. Wrap this handler with the same rate-limit middleware used for the other `/channels/:channelId/messages` routes (or apply the store-backed limiter at the router level before this handler), then add a test asserting 429 after the limit is exceeded.

harmony-frontend/src/components/channel/GuestMessageSearch.tsx
```diff
@@ -0,0 +1,203 @@
...
+        event.preventDefault();
+        if (isOpen) closeSearch();
+        else openSearch();
+      }
+      if (event.key === 'Escape') closeSearch();
+    }
+
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
+  }, [closeSearch, isOpen, openSearch]);
```
AvanishKulkarni, line 77: **[Minor — edge-case state glitch]** `setIsLoading(true)` is set here optimistically, but the actual fetch lives in a separate `useEffect` behind a 250 ms debounce. If the user types and then immediately clears the input, `handleQueryChange` resets `isLoading` to `false` — but the debounce timer for the previous non-empty value is still pending. When that timer fires and the `cancelled` flag is `true`, the callback returns early *without* calling `setIsLoading(false)`, so `isLoading` is already `false` via the clear-path reset and there is no double-set. In practice this works, but the correctness relies on React batching and the clear-path running after the optimistic set. A safer pattern is to derive loading state entirely inside the `useEffect` (set `true` at the top of the async callback, `false` after `setResults`), so the flag is always owned by the same code that drives it.

harmony-backend/src/routes/public.router.ts
```diff
@@ -79,6 +108,65 @@ export function createPublicRouter(store?: Store) {
...
+   * GET /api/public/channels/:channelId/messages/search?q=<keyword>
+   * Searches non-deleted messages within a single guest-accessible public channel.
+   */
+  router.get('/channels/:channelId/messages/search', async (req: Request, res: Response) => {
+    try {
+      const { channelId } = req.params;
+      const query = parseSearchQuery(req.query.q);
+      const limit = parseSearchLimit(req.query.limit);
+
+      if (!query) {
```
AvanishKulkarni, line 121: **[Observation — design intent worth documenting]** Only `PRIVATE` channels are gated here; `PUBLIC_NO_INDEX` channels are searchable. This is consistent with how the existing message-list endpoint treats `PUBLIC_NO_INDEX`, but operators who set a channel to `PUBLIC_NO_INDEX` to limit discoverability may not expect its content to be keyword-searchable by any unauthenticated guest. If the intent is intentional (access ≠ indexability), a short inline comment here would make that clear and prevent future reviewers from treating it as a bug.

harmony-backend/src/routes/public.router.ts
acabrera04: Fixed in c21a8c94. The public router already applies the injected store-backed limiter before this route, and I added a regression test proving guest message search returns 429 after the public limit is exhausted.

harmony-backend/src/routes/public.router.ts
acabrera04: Fixed in c21a8c94. Simplified the parser to use a direct Number(raw) coercion.

harmony-backend/src/routes/public.router.ts
acabrera04: Fixed in c21a8c94. Added a regression test asserting the guest message search endpoint returns 429 after the public rate limit is exhausted.

harmony-frontend/src/components/channel/GuestMessageSearch.tsx
acabrera04: Fixed in c21a8c94. Moved the optimistic loading transition out of handleQueryChange so loading is owned by the debounced search effect.

harmony-backend/src/routes/public.router.ts
acabrera04: Fixed in c21a8c94. Added an inline comment documenting that PUBLIC_NO_INDEX affects crawler indexing, while guest read and search access remains allowed.

AvanishKulkarni (approved):
## Review Checklist

- [x] **Prioritize Logic over Style:** Core search logic (debounce with cancellation, context windowing, input guards) is sound and non-redundant.
- [x] **Security First:** Input sanitized client-side (trim) and backend-side (`parseSearchQuery` with 120-char cap). Channel visibility verified before any DB query. `Cache-Control: no-store` set on the search response.
- [x] **Architectural Alignment:** The inline handler pattern is consistent with the rest of `public.router.ts`. Minor deviation from service-layer was flagged previously as non-blocking; no further concerns.
- [x] **Issue Completion:** All four acceptance criteria from #486 are now met — full-text search endpoint, UI component in `GuestChannelView`, results display author/timestamp/context, and rate limiting confirmed via regression test.
- [x] **No Nitpicking:** Review focused on correctness only.
- [x] **Avoid Repetition:** Previous comments have been addressed; no repeated findings.
- [x] **Iterative Review:** Previous blockers verified resolved (see below).
- [x] **Prevent CI Failures:** No syntax errors, imports correct, tests cover search path including 429.

---

## Previous Review Follow-up

All five inline comments from the prior review are resolved in commit `c21a8c94`:

- **Rate limit (blocker):** Confirmed — `router.use(createPublicRateLimiter(store))` at the top of `createPublicRouter` covers every route including the new search handler. Regression test proves 429 after 100 requests. ✅
- **Redundant ternary in `parseSearchLimit`:** Simplified to a direct `Number(raw)` coercion — no duplicate branches remain. ✅
- **Loading state glitch:** `setIsLoading(true)` moved inside the debounced `setTimeout` callback, so clearing the input before the timer fires can never leave a stuck loading indicator. ✅
- **`PUBLIC_NO_INDEX` design intent:** Inline comment added documenting that `PUBLIC_NO_INDEX` controls crawler indexing only, not guest read/search access. ✅
- **`PUBLIC_NO_INDEX` observation (design):** Documented as intended behavior. ✅

## Summary

This PR is clean and complete. The implementation is correct, well-tested, and all acceptance criteria of #486 are satisfied. Approving.

PR #617, opened 2026-05-03
fix: harden username validation

### LINKED ISSUE #441: [LOW] Username regex allows leading hyphens/underscores (impersonation risk)
## Description
The username validation regex `^[a-zA-Z0-9_-]+$` permits usernames like `--admin`, `_admin`, or `__system` that visually impersonate system or admin accounts in the UI. Combined with the dev-admin backdoor convention, this creates a social-engineering / phishing surface.

**Relevant file:** `harmony-backend/src/routes/auth.router.ts` line 31

## Fix
Require the username to start with an alphanumeric character:

```ts
/^[a-zA-Z0-9][a-zA-Z0-9_-]{2,31}$/
```

This ensures usernames are 3–32 characters and always begin with a letter or digit.

acabrera04:
## Summary
- Require registration usernames to start with a letter or number while preserving the existing 3-32 character length and allowed follow-up characters.
- Add route-level regression coverage for leading hyphen/underscore impersonation-style usernames.

Closes #441

## Verification
- `npm --prefix harmony-backend test -- auth.test.ts` (red before fix, green after)
- `npm --prefix harmony-backend run lint` (passes with existing warnings only)
- `npm --prefix harmony-frontend run lint` (passes with existing warnings only)
- `npm --prefix harmony-frontend test -- --runInBand`
- `npm --prefix harmony-frontend run build`

## Notes
- Issue #441 is assigned to `acabrera04`.
- Full backend suite was attempted, but the only visible long-running Jest process belonged to another worker's `/private/tmp/harmony-issue-480` worktree and was not interrupted.

### DIFF
--- a/harmony-backend/tests/auth.test.ts
+++ b/harmony-backend/tests/auth.test.ts
@@ -146,6 +146,29 @@ describe('POST /api/auth/register', () => {
     expect(res.body.error).toBe('Validation failed');
   });
 
+  it.each(['--admin', '_admin', '__system'])(
+    'rejects username %s because usernames must start with a letter or number',
+    async (username) => {
+      mockPrisma.user.findUnique.mockResolvedValue(null);
+      mockPrisma.user.create.mockResolvedValue({ ...mockUser, username });
+      mockPrisma.refreshToken.create.mockResolvedValue(mockRefreshToken);
+
+      const res = await request(app)
+        .post('/api/auth/register')
+        .set('Origin', 'http://localhost:3000')
+        .send({
+          email: `${username.replace(/[^a-zA-Z0-9]/g, '')}@example.com`,
+          username,
+          passwordSalt: PASSWORD_SALT,
+          passwordVerifier: derivePasswordVerifier('password123'),
+        });
+
+      expect(res.status).toBe(400);
+      expect(res.body.error).toBe('Validation failed');
+      expect(mockPrisma.user.create).not.toHaveBeenCalled();
+    },
+  );
+
   it('returns 409 when email is already in use', async () => {
     mockPrisma.user.findUnique.mockResolvedValue(mockUser);
 
--- a/harmony-backend/tests/auth.test.ts
+++ b/harmony-backend/tests/auth.test.ts
@@ -149,10 +149,6 @@ describe('POST /api/auth/register', () => {
   it.each(['--admin', '_admin', '__system'])(
     'rejects username %s because usernames must start with a letter or number',
     async (username) => {
-      mockPrisma.user.findUnique.mockResolvedValue(null);
-      mockPrisma.user.create.mockResolvedValue({ ...mockUser, username });
-      mockPrisma.refreshToken.create.mockResolvedValue(mockRefreshToken);
-
       const res = await request(app)
         .post('/api/auth/register')
         .set('Origin', 'http://localhost:3000')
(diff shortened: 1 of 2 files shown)


AvanishKulkarni (commented):
## Review

### Overview
A clean, targeted fix for issue #441. The regex change is correct and the test coverage directly addresses the impersonation vectors. Two things worth addressing before merge.

---

### Correctness
- Regex  is sound — first char anchored to alphanumeric, tail allows `_` and `-`. The `min(3)`/`max(32)` chain still applies independently.
- End anchor `$` is present, no partial-match bypass.

### Gaps / Risks

| Area | Finding |
|---|---|
| **Frontend sync** | No client-side username regex on the signup page — users only get an error after a round-trip. Low correctness risk but a UX gap. |
| **Mention regex** | `MessageItem.tsx:706` uses `/@([a-zA-Z0-9_-]{1,32})/g` which doesn't require a letter/number start. Minor inconsistency, low practical risk since no existing users have leading-hyphen names. |

### Verdict
Approve with minor suggestions (see inline comments). Core change is solid.

AvanishKulkarni (commented):
## Review

### Overview
A clean, targeted fix for issue #441. The regex change is correct and the test coverage directly addresses the impersonation vectors. Two things worth addressing before merge.

---

### Correctness
- Regex `^[a-zA-Z0-9][a-zA-Z0-9_-]*$` is sound — first char anchored to alphanumeric, tail allows `_` and `-`. The `min(3)`/`max(32)` chain still applies independently.
- End anchor `$` is present, no partial-match bypass.

### Gaps / Risks

| Area | Finding |
|---|---|
| **Frontend sync** | No client-side username regex on the signup page — users only get an error after a round-trip. Low correctness risk but a UX gap. |
| **Mention regex** | `MessageItem.tsx:706` uses `/@([a-zA-Z0-9_-]{1,32})/g` which does not require a letter/number start. Minor inconsistency, low practical risk since no existing users have leading-hyphen names. |

### Verdict
Approve with minor suggestions (see inline comments). Core change is solid.

harmony-backend/tests/auth.test.ts
```diff
@@ -146,6 +146,29 @@ describe('POST /api/auth/register', () => {
...
+      mockPrisma.user.create.mockResolvedValue({ ...mockUser, username });
+      mockPrisma.refreshToken.create.mockResolvedValue(mockRefreshToken);
+
+      const res = await request(app)
+        .post('/api/auth/register')
+        .set('Origin', 'http://localhost:3000')
+        .send({
+          email: `${username.replace(/[^a-zA-Z0-9]/g, '')}@example.com`,
+          username,
+          passwordSalt: PASSWORD_SALT,
```
AvanishKulkarni, line 158: The mock setup for `user.create` and `refreshToken.create` is unnecessary here — the request is rejected at the Zod validation layer and never reaches the DB. Removing these makes the test intent clearer: this is a pure input-rejection test, not a DB interaction test.

harmony-backend/tests/auth.test.ts
```diff
@@ -146,6 +146,29 @@ describe('POST /api/auth/register', () => {
...
+        });
+
+      expect(res.status).toBe(400);
+      expect(res.body.error).toBe('Validation failed');
+      expect(mockPrisma.user.create).not.toHaveBeenCalled();
+    },
+  );
+
   it('returns 409 when email is already in use', async () => {
     mockPrisma.user.findUnique.mockResolvedValue(mockUser);
```
AvanishKulkarni, line 199: Consider adding at least one positive case here — e.g. a username starting with a digit (`0validname`) — to confirm numeric-first usernames are accepted. The existing happy-path test (`alice`) only covers pure-alpha starts, so the numeric-start branch of the new regex is currently untested.

harmony-backend/tests/auth.test.ts
acabrera04: Fixed in 166fcfac. Removed the unnecessary user.create and refreshToken.create mocks from the invalid-username rejection test so the test stays focused on validation-layer rejection.

harmony-backend/tests/auth.test.ts
acabrera04: Fixed in 166fcfac. Added a positive registration test for a username that starts with a digit (0validname), including an assertion that the created username is passed through.

PR #618, opened 2026-05-03
Require proxy hop config in production

### LINKED ISSUE #440: [LOW] Require TRUST_PROXY_HOPS in production to prevent rate-limit bypass
## Description
If `TRUST_PROXY_HOPS` is not set in a production deployment (e.g., on Railway behind a reverse proxy), `req.ip` resolves to the proxy's IP instead of the client's. This means all rate-limit hits are bucketed under the same IP, effectively disabling per-client login throttling and allowing unlimited brute-force attempts.

**Relevant file:** `harmony-backend/src/app.ts` lines 99–114

## Fix
Add a startup guard that throws when `NODE_ENV === 'production'` and `TRUST_PROXY_HOPS` is unset or zero:

```ts
if (process.env.NODE_ENV === 'production' && !process.env.TRUST_PROXY_HOPS) {
  throw new Error('TRUST_PROXY_HOPS must be set in production');
}
```

acabrera04:
## Summary
- Require a positive TRUST_PROXY_HOPS value when the backend starts in production
- Add regression coverage for unset, zero, and valid production proxy hop config
- Keep production-mode tRPC error formatter tests isolated from Redis-backed rate limiter stores

## Verification
- npm --prefix harmony-backend test -- --runTestsByPath tests/app.rate-limit.test.ts tests/trpc.error-formatter.test.ts --runInBand
- npm --prefix harmony-backend run lint (passes with existing warnings in unrelated tests/channelMember.test.ts and tests/events.router.sse-server-updated.test.ts)

Full backend suite note: npm --prefix harmony-backend test -- --runInBand is blocked in this local environment by missing DATABASE_URL and Redis NOAUTH; the issue-related targeted backend tests pass.

Closes #440

### DIFF
--- a/harmony-backend/src/app.ts
+++ b/harmony-backend/src/app.ts
@@ -153,15 +153,14 @@ export function createApp(options: CreateAppOptions = {}) {
   // clients spoof XFF and poison rate-limit buckets. Set to `1` on Railway.
   const trustProxyHopsEnv = process.env.TRUST_PROXY_HOPS;
   if (trustProxyHopsEnv !== undefined && trustProxyHopsEnv.trim() === '') {
-    throw new Error(
-      `Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected a non-negative integer.`,
-    );
+    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected an integer.`);
   }
   const trustProxyHops = trustProxyHopsEnv === undefined ? 0 : Number(trustProxyHopsEnv);
   if (!Number.isInteger(trustProxyHops) || trustProxyHops < 0) {
-    throw new Error(
-      `Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected a non-negative integer.`,
-    );
+    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected an integer.`);
+  }
+  if (isProduction && trustProxyHops < 1) {
+    throw new Error('TRUST_PROXY_HOPS must be set to a positive integer in production.');
   }
   if (trustProxyHops > 0) {
     app.set('trust proxy', trustProxyHops);
--- a/harmony-backend/src/app.ts
+++ b/harmony-backend/src/app.ts
@@ -153,11 +153,11 @@ export function createApp(options: CreateAppOptions = {}) {
   // clients spoof XFF and poison rate-limit buckets. Set to `1` on Railway.
   const trustProxyHopsEnv = process.env.TRUST_PROXY_HOPS;
   if (trustProxyHopsEnv !== undefined && trustProxyHopsEnv.trim() === '') {
-    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected an integer.`);
+    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected a non-negative integer.`);
   }
   const trustProxyHops = trustProxyHopsEnv === undefined ? 0 : Number(trustProxyHopsEnv);
   if (!Number.isInteger(trustProxyHops) || trustProxyHops < 0) {
-    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected an integer.`);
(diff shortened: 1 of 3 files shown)


harmony-backend/src/app.ts
```diff
@@ -153,15 +153,14 @@ export function createApp(options: CreateAppOptions = {}) {
   // clients spoof XFF and poison rate-limit buckets. Set to `1` on Railway.
   const trustProxyHopsEnv = process.env.TRUST_PROXY_HOPS;
   if (trustProxyHopsEnv !== undefined && trustProxyHopsEnv.trim() === '') {
-    throw new Error(
-      `Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected a non-negative integer.`,
-    );
+    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected an integer.`);
   }
```
AvanishKulkarni, line 157: The error message was changed from `"Expected a non-negative integer"` to `"Expected an integer"`, but the `if (!Number.isInteger(trustProxyHops) || trustProxyHops < 0)` guard below still rejects negatives. This makes the message misleading — an operator who passes `-1` will see "Expected an integer" when the real constraint is "non-negative integer". Please revert both occurrences back to `"Expected a non-negative integer"`.

harmony-backend/src/app.ts
```diff
@@ -153,15 +153,14 @@ export function createApp(options: CreateAppOptions = {}) {
...
   const trustProxyHops = trustProxyHopsEnv === undefined ? 0 : Number(trustProxyHopsEnv);
   if (!Number.isInteger(trustProxyHops) || trustProxyHops < 0) {
-    throw new Error(
-      `Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected a non-negative integer.`,
-    );
+    throw new Error(`Invalid TRUST_PROXY_HOPS value "${trustProxyHopsEnv}". Expected an integer.`);
+  }
```
AvanishKulkarni, line 161: Same misleading message here. This branch fires when the value is a valid integer but negative (e.g. `TRUST_PROXY_HOPS=-1`). Saying `"Expected an integer"` is incorrect — the constraint is non-negative. Revert to `"Expected a non-negative integer"` to match the actual validation condition.

AvanishKulkarni (commented):
## Review Checklist

- [x] **Prioritize Logic over Style** — Logic is sound. The production guard correctly catches both `undefined` and `0`. The `afterEach` env-cleanup refactor is cleaner and equivalent.
- [x] **Security First** — Positive hardening change. Closes a real rate-limit bypass; no new vulnerabilities introduced.
- [x] **Architectural Alignment** — Scoped to startup validation and test isolation; no architectural patterns violated.
- [x] **Issue Completion** — Guard throws on both unset and zero `TRUST_PROXY_HOPS` in production. Tests cover all three required cases (unset, zero, valid), satisfying [#440](https://github.com/CS485-Harmony/Harmony/issues/440).
- [x] **No Nitpicking** — Only actionable logic issues flagged below.
- [x] **Prevent CI Failures** — Tests are self-contained, env state is properly restored, and `createNoopRateLimitStore` prevents production-mode tests from hitting Redis. No syntax or import issues.

---

## Overview

This PR adds a fail-fast startup guard to `app.ts`: if `NODE_ENV === 'production'` and `TRUST_PROXY_HOPS` is unset or `0`, the app throws immediately rather than silently bucketing all rate-limit hits under the proxy IP. It also adds targeted regression tests and fixes a gap where `trpc.error-formatter.test.ts` would fail to connect to Redis when run in production mode.

The core logic is correct and the test coverage is solid. One real issue found: the error message for the existing non-negative integer validation was weakened in this PR (see inline comment).

**`createNoopRateLimitStore` in `trpc.error-formatter.test.ts`** — the `init` method is typed with a synchronous `void` return. Verify the installed version of `express-rate-limit`'s `Store` interface accepts `void` (not `Promise<void>`) for `init`. This is likely fine today but could silently break on a dependency bump. Making it `async init` costs nothing and is future-safe.

harmony-backend/src/app.ts
acabrera04: Fixed in 96d06b27. Restored the TRUST_PROXY_HOPS validation error to say "Expected a non-negative integer" so the message matches the negative-value guard.

harmony-backend/src/app.ts
acabrera04: Fixed in 96d06b27. Restored the negative TRUST_PROXY_HOPS validation error to say "Expected a non-negative integer" to match the actual constraint.

PR #620, opened 2026-05-03
Fix SSE membership revalidation

### LINKED ISSUE #480: [Security] SSE streams do not re-validate membership after connection
## Summary
SSE authorization is only checked when the stream is established. If a user is removed from a server after connecting, the stream can continue delivering events until disconnect.

## Evidence
- Initial membership checks only at connect:
  - `harmony-backend/src/routes/events.router.ts` (`/channel/:channelId` membership check)
  - `harmony-backend/src/routes/events.router.ts` (`/server/:serverId` membership check)
- Long-lived event handlers continue without periodic auth/membership revalidation.

## Risk
Users who lose access may continue receiving private updates (messages/member events) on an already-open SSE connection.

## Suggested remediation
1. Enforce revocation-aware SSE lifecycle:
   - periodic membership revalidation, or
   - subscription invalidation on membership/visibility change events, or
   - session versioning and forced close when version changes.
2. Close active SSE connections immediately when membership is revoked.

## Acceptance criteria
- Revoked users stop receiving SSE events without requiring manual reconnect.
- Automated test covers “connected user removed from server/channel” scenario.

acabrera04:
## Summary
- close server/channel SSE streams immediately when the connected user is removed from the server
- add periodic SSE authorization revalidation for server streams and private channel membership changes
- add regression coverage for connected-then-revoked SSE access

Closes #480

## Verification
- npm --prefix harmony-backend test -- events.router.test.ts events.router.server.test.ts --runInBand
- npm --prefix harmony-backend test -- events.router.member.test.ts events.router.visibility.test.ts --runInBand
- npm --prefix harmony-backend run lint
- npm --prefix harmony-frontend run lint
- npm --prefix harmony-frontend run build

Note: full backend Jest was attempted but hung without output in this isolated worktree and was stopped; focused SSE suites passed.

### DIFF
--- a/harmony-backend/src/routes/events.router.ts
+++ b/harmony-backend/src/routes/events.router.ts
@@ -147,6 +147,7 @@ type BufferedSseState = {
   ready: boolean;
   pendingFrames: string[];
   heartbeat: ReturnType<typeof setInterval> | null;
+  authorizationRevalidation: ReturnType<typeof setInterval> | null;
 };
 
 function formatEvent(eventType: string, data: unknown, id?: string): string {
@@ -160,6 +161,7 @@ function createBufferedSseState(): BufferedSseState {
     ready: false,
     pendingFrames: [],
     heartbeat: null,
+    authorizationRevalidation: null,
   };
 }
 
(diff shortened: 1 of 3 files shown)


harmony-backend/src/routes/events.router.ts
```diff
@@ -170,12 +172,99 @@ function cleanupSseConnection(state: BufferedSseState, subscriptions: EventSubsc
...
   }
 }
 
+function closeSseForRevokedAccess(
+  res: Response,
+  state: BufferedSseState,
+  subscriptions: EventSubscription[],
+): void {
+  if (state.closed) return;
+  cleanupSseConnection(state, subscriptions);
```
AvanishKulkarni, line 192: **[Blocking] Missing `state.closed = true` — re-entry is possible**

`state.closed` is never set here, so the `if (state.closed) return;` guard on line 190 doesn't actually latch. After `cleanupSseConnection` runs, a concurrent `MEMBER_LEFT` event or a revalidation tick firing within the same event-loop turn can re-enter this function and call `res.end()` a second time.

Fix: set the flag before cleanup so re-entry is blocked:
```ts
function closeSseForRevokedAccess(res, state, subscriptions) {
  if (state.closed) return;
  state.closed = true; // must come first, before any async-reachable side effects
  cleanupSseConnection(state, subscriptions);
  ...
}
```

harmony-backend/src/routes/events.router.ts
```diff
@@ -170,12 +172,99 @@ function cleanupSseConnection(state: BufferedSseState, subscriptions: EventSubsc
...
+  channelId: string,
+  serverId: string,
+): Promise<boolean> {
+  const membership = await prisma.serverMember.findFirst({
+    where: { userId, serverId },
+    select: { role: true },
+  });
+  if (!membership) return false;
+
+  const channel = await prisma.channel.findUnique({
```
AvanishKulkarni: **[Performance] Sequential DB round-trips on the revalidation hot path**

Trips 1 (serverMember) and 2 (channel) have no data dependency on each other — neither result feeds into the other's query. Running them sequentially adds avoidable latency on every 30-second revalidation tick. They can be parallelised:

```ts
const [membership, channel] = await Promise.all([
  prisma.serverMember.findFirst({ where: { userId, serverId }, select: { role: true } }),
  prisma.channel.findUnique({ where: { id: channelId }, select: { serverId: true, visibility: true } }),
]);
```

Additionally, at the revalidation call site `serverId` is already known from the outer closure, so the `channel.serverId !== serverId` guard in trip 2 is redundant on revalidation (the channel cannot move servers). You could skip trip 2 entirely on revalidation if you restructure `canAccessChannelStream` to accept an optional pre-fetched `visibility`.

harmony-backend/tests/events.router.server.test.ts
```diff
@@ -561,6 +555,82 @@ describe('GET /api/events/server/:serverId — authorisation', () => {
...
+      });
+      req.on('error', reject);
+      req.setTimeout(1_000, () => {
+        req.destroy();
+        reject(new Error('SSE stream did not close after membership revalidation'));
+      });
+    });
+
+    expect(prisma.serverMember.findFirst).toHaveBeenCalledTimes(2);
+    delete process.env.SSE_MEMBERSHIP_REVALIDATION_INTERVAL_MS;
```
AvanishKulkarni: **[Test reliability] `process.env` mutation not cleaned up on failure path**

The `delete process.env.SSE_MEMBERSHIP_REVALIDATION_INTERVAL_MS` here only runs when the promise resolves successfully. If the 1-second timeout fires and rejects, the 20 ms interval bleeds into subsequent tests in the suite.

Move the cleanup into `afterEach` (or wrap the whole test body in `try/finally`):
```ts
afterEach(() => {
  delete process.env.SSE_MEMBERSHIP_REVALIDATION_INTERVAL_MS;
});
```
The same issue exists in the corresponding test in `events.router.test.ts`.

AvanishKulkarni (changes requested):
## PR Review Checklist

- [x] **Prioritize Logic over Style** — Logic is sound: immediate revocation via event subscription + periodic fallback revalidation. No DRY violations.
- [x] **Security First** — Directly addresses the auth-only-at-connect gap from #480. No new vulnerabilities.
- [x] **Architectural Alignment** — New helper functions follow existing patterns in `events.router.ts`.
- [x] **Issue Completion** — Both acceptance criteria from #480 are satisfied: revoked users' streams close immediately, and automated tests cover the scenario.
- [x] **No Nitpicking** — Review limited to logic/correctness.
- [x] **Iterative Reviews** — N/A.
- [x] **Prevent CI Failures** — New mocks for `channelMember.findUnique` and `MEMBER_LEFT` channel added correctly to existing test setup. No obvious pipeline issues.

---

## Review

The overall approach is well-designed — the dual reactive+polling strategy (immediate `MEMBER_LEFT` event close + periodic revalidation fallback) is the right architecture for this problem. Three issues need to be addressed before merge.

See inline comments for specifics.

harmony-backend/src/routes/events.router.ts
acabrera04: Fixed in 472a5588. `closeSseForRevokedAccess` now latches `state.closed` before teardown, while cleanup remains idempotent and clears the subscription list after unsubscribing.

harmony-backend/src/routes/events.router.ts
acabrera04: Fixed in 472a5588. `canAccessChannelStream` now fetches the server membership and channel metadata with `Promise.all`, removing the sequential DB round-trip on the revalidation path.

harmony-backend/tests/events.router.server.test.ts
acabrera04: Fixed in 472a5588. Both SSE test files now clean up `SSE_MEMBERSHIP_REVALIDATION_INTERVAL_MS` in `afterEach`, so the shortened interval is removed even when a test fails.

PR #621, opened 2026-05-03
Limit public auth validation error details

### LINKED ISSUE #438: [LOW] Limit Zod validation error details on public auth endpoints
## Description
The public `/api/auth/register` and `/api/auth/login` endpoints return the full `ZodError['errors']` array on 400 responses, exposing internal field names, path structure, and validation rule details to unauthenticated clients.

**Relevant file:** `harmony-backend/src/routes/auth.router.ts` (multiple error response sites)

## Fix
On public (unauthenticated) endpoints, return only human-readable per-field messages rather than the full Zod issue tree. For example:

```ts
res.status(400).json({
  error: 'Validation failed',
  fields: zodError.errors.map(e => ({ field: e.path.join('.'), message: e.message }))
});
```

acabrera04:
## Summary

- Replace raw Zod issue-tree responses in the auth router with sanitized `{ field, message }` validation fields.
- Add register/login regression tests that assert public auth validation responses do not expose raw Zod `details`.

## Verification

- `npm --prefix harmony-backend test -- auth.test.ts`
- `npm --prefix harmony-backend run lint` (passes with pre-existing warnings only)
- `npm --prefix harmony-frontend run lint` (passes with pre-existing warnings only)
- `npm --prefix harmony-frontend test -- --runInBand`
- `npm --prefix harmony-frontend run build`

Note: full `npm --prefix harmony-backend test -- --runInBand` could not complete in this fresh worktree because unrelated DB/Redis-backed suites require local `DATABASE_URL` and Redis auth configuration.

Closes #438

### DIFF
--- a/harmony-backend/src/routes/auth.router.ts
+++ b/harmony-backend/src/routes/auth.router.ts
@@ -74,6 +74,7 @@ function publicValidationError(error: z.ZodError): {
   return {
     error: 'Validation failed',
     fields: error.errors.map((issue) => ({
+      // Empty field names represent root-level validation errors with no Zod path.
       field: issue.path.map(String).join('.'),
       message: issue.message,
     })),
--- a/harmony-backend/src/routes/auth.router.ts
+++ b/harmony-backend/src/routes/auth.router.ts
@@ -67,13 +67,26 @@ function trpcCodeToHttp(code: TRPCError['code']): number {
   }
 }
 
+function publicValidationError(error: z.ZodError): {
+  error: 'Validation failed';
+  fields: Array<{ field: string; message: string }>;
+} {
+  return {
+    error: 'Validation failed',
+    fields: error.errors.map((issue) => ({
+      field: issue.path.map(String).join('.'),
+      message: issue.message,
+    })),
+  };
+}
+
 function handleError(res: Response, err: unknown): void {
   if (err instanceof TRPCError) {
     res.status(trpcCodeToHttp(err.code)).json({ error: err.message });
     return;
   }
   if (err instanceof z.ZodError) {
-    res.status(400).json({ error: 'Validation failed', details: err.errors });
+    res.status(400).json(publicValidationError(err));
     return;
   }
   logger.error({ err }, 'Auth route failed');
@@ -97,7 +110,7 @@ authRouter.post('/register/challenge', (_req: Request, res: Response) => {
 authRouter.post('/register', async (req: Request, res: Response) => {
   const parsed = registerSchema.safeParse(req.body);
   if (!parsed.success) {
-    res.status(400).json({ error: 'Validation failed', details: parsed.error.errors });
+    res.status(400).json(publicValidationError(parsed.error));
     return;
   }
 
@@ -117,7 +130,7 @@ authRouter.post('/register', async (req: Request, res: Response) => {
 authRouter.post('/login/challenge', async (req: Request, res: Response) => {
   const parsed = challengeSchema.safeParse(req.body);
   if (!parsed.success) {
-    res.status(400).json({ error: 'Validation failed', details: parsed.error.errors });
+    res.status(400).json(publicValidationError(parsed.error));
     return;
   }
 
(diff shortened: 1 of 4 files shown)


harmony-backend/src/routes/auth.router.ts
```diff
@@ -67,13 +67,26 @@ function trpcCodeToHttp(code: TRPCError['code']): number {
...
+function publicValidationError(error: z.ZodError): {
+  error: 'Validation failed';
+  fields: Array<{ field: string; message: string }>;
+} {
+  return {
+    error: 'Validation failed',
+    fields: error.errors.map((issue) => ({
+      field: issue.path.map(String).join('.'),
+      message: issue.message,
+    })),
```
AvanishKulkarni, line 80: The helper correctly strips Zod internals. One edge case: when a ZodError fires at the root level (e.g. union/enum mismatch with no path), `issue.path` is `[]` and `join('.')` produces `field: ""`. Callers receiving an empty string should treat it as a root-level error — worth a brief note in the PR description or a code comment so the frontend knows to handle it.

AvanishKulkarni (changes requested):
## Review Checklist

- [x] **1. Logic / DRY** — `publicValidationError` is extracted into a single helper and applied consistently across all 5 affected call sites. No duplication.
- [x] **2. Security First** — Strips raw Zod internals (`code`, `validation`, `path` array, etc.) from public responses. Positive change.
- [x] **3. Architectural Alignment** — No architectural violations; change is scoped to the auth REST layer.
- [x] **4. Issue Completion** — Matches the fix shape specified in #438 exactly.
- [x] **5. No Nitpicking** — Focused on critical issues only.
- [ ] **8. Prevent CI Failures** — One breaking client-side regression identified (see inline comment).

---

**This PR achieves its security goal cleanly**, but it introduces a silent regression in the frontend's error display path that will ship if merged as-is.

The backend sanitization is correct. However, `harmony-frontend/src/lib/utils.ts` still reads `data.details` to extract validation messages. After this change the backend emits `fields` instead, so the branch will never match — the user will see the generic fallback `"Something went wrong. Please try again."` instead of the actual field-level message on register and login failures.

Additionally, when a Zod error fires at the top level (no path), `issue.path.map(String).join('.')` yields `field: ""`. This is consistent with the issue recommendation, but callers should treat an empty string as a root-level error.

**Verdict: Request changes.** The frontend `getUserErrorMessage` utility must be updated in the same PR to consume `fields` before this is safe to merge.

AvanishKulkarni:
**Frontend breakage — `getUserErrorMessage` reads `details`, not `fields`**

`harmony-frontend/src/lib/utils.ts:96`:
```ts
// current code — will silently stop working after this PR
if (Array.isArray(data.details) && data.details.length > 0) {
  const messages = data.details.map((d: { message?: string }) => d.message).filter(Boolean);
  if (messages.length > 0) return messages.join('. ');
}
```

After this PR the backend returns `fields` instead of `details`. The `data.details` branch never fires, then the `data.error !== 'Validation failed'` guard also blocks the fallback, so the user sees the generic `"Something went wrong. Please try again."` string instead of the actual field messages on register/login.

Suggested fix for `utils.ts`:
```ts
// Validation errors: { error: "Validation failed", fields: [{ field, message }] }
if (Array.isArray(data.fields) && data.fields.length > 0) {
  const messages = data.fields.map((f: { message?: string }) => f.message).filter(Boolean);
  if (messages.length > 0) return messages.join('. ');
}
```

This should be done in the same PR so the backend and frontend stay in sync.

harmony-backend/src/routes/auth.router.ts
acabrera04: Fixed in 32abada5. Added a code comment documenting that empty validation field names represent root-level Zod errors with no path. Also fixed the related frontend consumer in 5aabd86a so field messages, including root-level entries, are displayed correctly.

acabrera04:
Processed the top-level review note about `getUserErrorMessage` still reading `details`: fixed in 5aabd86a by reading the new `fields` validation response before the legacy `details` shape, and added a unit test covering both field-level and root-level validation messages. The inline root-level field documentation thread was also fixed in 32abada5 and resolved.

PR #622, opened 2026-05-03
Fix stale system admin cache

### LINKED ISSUE #435: [MEDIUM] Fix stale isSystemAdmin module-scope cache
## Description
`isSystemAdmin` caches a single admin user ID at module scope (`_adminUserId`) indefinitely. If the admin user row is deleted and recreated (common in dev seed flows, or in any future multi-tenant scenario), the old UUID remains "admin" for the entire process lifetime, granting unintended permission bypass.

**Relevant file:** `harmony-backend/src/lib/admin.utils.ts` lines 14–31

## Fix
Remove the indefinite module-scope cache. Either perform a fresh DB lookup each time (the query is trivial), use a short TTL cache, or invalidate the cache on seed/migration. If the cache is kept for performance, it must also cache negative results (non-admin) to prevent stale positives.

acabrera04:
## Summary
- remove the module-scope positive admin ID cache from isSystemAdmin
- add regression coverage proving a deleted admin row no longer stays authorized

## Verification
- npm test -- --runTestsByPath tests/admin.utils.test.ts --runInBand
- npm run lint (passes with two unrelated existing warnings)
- npm run build (passes after allowing Prisma cache writes)
- attempted full npm test -- --runInBand; sandboxed run hit listen EPERM on Supertest, approved run hung silently and was stopped for this worktree

Closes #435

### DIFF
--- a/harmony-backend/tests/admin.utils.test.ts
+++ b/harmony-backend/tests/admin.utils.test.ts
@@ -5,7 +5,12 @@
  * matches ADMIN_EMAIL cannot escalate privileges in production.
  */
 
-import { isSystemAdmin, ADMIN_EMAIL, RESERVED_EMAILS, RESERVED_USERNAMES } from '../src/lib/admin.utils';
+import {
+  isSystemAdmin,
+  ADMIN_EMAIL,
+  RESERVED_EMAILS,
+  RESERVED_USERNAMES,
+} from '../src/lib/admin.utils';
 
 const ADMIN_USER_ID = '00000000-0000-0000-0000-000000000099';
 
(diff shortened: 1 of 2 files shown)


harmony-backend/tests/admin.utils.test.ts
```diff
@@ -85,6 +90,18 @@ describe('isSystemAdmin — ENABLE_DEV_ADMIN guard (issue #425)', () => {
 
     expect(result).toBe(false);
   });
+
+  it('rechecks the database so a deleted admin row does not stay authorized', async () => {
+    Object.defineProperty(process.env, 'NODE_ENV', { value: 'development', writable: true });
```
AvanishKulkarni, line 95: Nit (pre-existing pattern, not introduced here): `Object.defineProperty` to set `NODE_ENV` is repeated across several tests in this describe block. Worth consolidating into a `beforeEach` / `afterEach` in a follow-up to keep each test focused on its assertion.

AvanishKulkarni (commented):
## Overview

Removes the module-scope `_adminUserId` cache from `isSystemAdmin`. The cached positive lookup meant a revoked or deleted admin account would remain authorized for the entire process lifetime. Removing it entirely is the right call, and the simplification is clean.

---

## Security

- **Bug fixed correctly.** The old cache was a privilege-escalation vector: demote or delete an admin and the process still recognizes them as admin until restart.
- **Production blast radius is zero.** `isSystemAdmin` short-circuits with `return false` when `isDevAdminEnabled()` is false, so this code path is unreachable in production/staging.

---

## Code Quality

- The function body now collapses to a single `return` expression — clean and readable.
- The JSDoc comment is correctly updated to drop the caching mention.
- Import formatting in the test file is a minor cosmetic improvement.

---

## Test Coverage

The new regression test is well-constructed: `mockResolvedValueOnce` is chained twice to simulate first-call admin → second-call deleted, and asserts `findUnique` was called twice, directly proving the cache is gone.

---

## Risks / Concerns

- **DB call on every invocation (dev only).** Acceptable given this guard is dev-only and infrequently called. No production impact.
- **Full test suite could not be verified end-to-end** (Supertest EPERM / hung run noted in the PR description). The targeted test file passes, and the change is isolated, but confirm the GitHub Actions CI check passes before merging.

---

**Verdict: Approve pending green CI.** The fix is correct, minimal, and well-tested. No code changes needed.

PR #641, opened 2026-05-04
fix(notifications): enlarge dismiss button and fix markAllAsRead body

declanblanc:
## Summary

- **Dismiss button too small to click**: The per-notification mark-as-read button was rendered as a bare `✓` character with `text-[10px]` and no padding, making it nearly impossible to hit. Replaced with a properly-sized SVG checkmark icon (`h-3.5 w-3.5`) inside a `p-1.5` padded button that also shows a hover background, giving a comfortable tap target.

- **"Mark all as read" not clearing notifications**: The previous fix (#602) changed `input ?? null` → `input` (sending `undefined`) to avoid serializing JSON `null`. However, passing `undefined` to Axios with a globally-set `Content-Type: application/json` header can produce an ambiguous empty body that some body-parser configurations reject. Now `markAllAsRead` explicitly passes `{}` so Axios always sends a valid JSON object body and the tRPC adapter can parse it reliably.

## Changes

- `harmony-frontend/src/components/channel/NotificationBell.tsx`: enlarged dismiss button; pass `{}` to `markAllAsRead`
- `harmony-frontend/src/__tests__/NotificationBell.test.tsx`: updated assertion to match `{}` argument

## Test plan

- [ ] Verify unread notifications show a clearly-clickable checkmark button that dismisses the individual notification and removes it from the unread count
- [ ] Verify "Mark all as read" clears the unread badge and removes the indigo highlight from all notifications
- [ ] All 13 `NotificationBell` unit tests pass

### DIFF
--- a/harmony-frontend/src/__tests__/NotificationBell.test.tsx
+++ b/harmony-frontend/src/__tests__/NotificationBell.test.tsx
@@ -179,7 +179,7 @@ describe('NotificationBell — mark-as-read flows (Issue #574)', () => {
     });
 
     await waitFor(() =>
-      expect(mockTrpcMutation).toHaveBeenCalledWith('notification.markAllAsRead'),
+      expect(mockTrpcMutation).toHaveBeenCalledWith('notification.markAllAsRead', {}),
     );
   });
 
--- a/harmony-frontend/src/components/channel/NotificationBell.tsx
+++ b/harmony-frontend/src/components/channel/NotificationBell.tsx
@@ -182,7 +182,7 @@ export function NotificationBell({ userId }: NotificationBellProps) {
 
   const markAllAsRead = async () => {
     try {
-      await apiClient.trpcMutation('notification.markAllAsRead');
+      await apiClient.trpcMutation('notification.markAllAsRead', {});
       setNotifications((prev) => prev.map((n) => ({ ...n, read: true })));
     } catch {
       // ignore — non-critical, badge will self-correct on next load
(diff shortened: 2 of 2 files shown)


AvanishKulkarni (approved):
## PR Review Checklist

- [x] **Logic over Style** — Both changes address real behavioral issues (inaccessible tap target; potential request body parsing failure), not cosmetic preferences.
- [x] **Security First** — No security surface changes. No new inputs or auth paths touched.
- [x] **Architectural Alignment** — Inline SVG pattern matches the existing `BellIcon` usage in the same file. Call-site fix stays in the component, consistent with how other mutations are invoked.
- [x] **Issue Completion** — PR description maps directly to two concrete bugs (button hit-target too small; mark-all-as-read body parsing). Both are addressed and tested.
- [x] **No Nitpicking** — Review focuses only on logic and correctness.
- [x] **Prevent CI Failures** — The test assertion is updated in lockstep with the call-site change; all 13 `NotificationBell` tests should continue passing.

---

## Review

Both fixes are correct and well-scoped. A few inline observations below.

**Dismiss button (UI fix):**
The replacement of the bare `✓` text with a padded SVG button is the right call. The added `aria-label` is a meaningful accessibility improvement, and the `p-1.5` padding gives a comfortable touch target.

**Mark-all-as-read body fix:**
The `trpcMutation` implementation in `api-client.ts` already applies `input ?? null` (line 244), so omitting the argument previously sent `null` rather than a truly empty body. That said, explicitly passing `{}` at the call site is a safer and clearer default for a no-input mutation — it removes the null/undefined ambiguity entirely and is more idiomatic for tRPC. The test update correctly mirrors this change.

No blocking issues. Approving.

acabrera04 (approved):
## Review checklist

- [x] Logic over style: The mutation call now sends an explicit `{}` payload for the no-input tRPC mutation, and the individual mark-as-read button remains scoped to its notification while stopping row navigation propagation.
- [x] Security first: No new trust boundary, credential handling, or user-controlled data exposure is introduced.
- [x] Architectural alignment: The change stays within the existing `apiClient.trpcMutation` and `NotificationBell` patterns.
- [x] Issue completion: The PR addresses the stated notification usability and mark-all-read request-body behavior.
- [x] No nitpicking: No style-only feedback added.
- [x] Avoid repetition: There were no existing comments or review threads on this PR.
- [x] Iterative reviews: No prior review from me to re-check.
- [x] Prevent CI failures: I checked the touched code for missing imports/syntax issues; GitHub reports frontend lint/build, frontend tests, backend lint/build, backend tests, and integration tests all passing.

Approved. The patch is narrow and matches the intended behavior. I attempted the focused local frontend Jest run with `rtk npm --prefix harmony-frontend test -- NotificationBell.test.tsx --runInBand`, but this worktree has no `harmony-frontend/node_modules/.bin/jest`, so I relied on the green CI test jobs for executable verification.

PR #642, opened 2026-05-04
fix(settings): use authenticated channel lookup for private server channel settings

declanblanc:
## Summary

- Channel settings pages (`/settings/:serverSlug/:channelSlug`) returned a 404 for private servers
- Root cause: `getChannel` hit the public REST endpoint `GET /api/public/servers/:serverSlug` which filters by `isPublic: true`, so private servers returned a 404 and `getChannel` returned `null` early — triggering `notFound()` in the page
- Fix: added `getChannelAuthenticated` to `channelService.ts` which goes directly to the authenticated tRPC `channel.getChannel` procedure; the settings page now uses the `server.id` already returned by the existing `requireServerSettingsAccess` call to fetch the channel without touching the public endpoint

## Changes

- `harmony-frontend/src/services/channelService.ts` — added `getChannelAuthenticated(serverId, channelSlug)` helper that uses the authenticated tRPC path
- `harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx` — swapped `getChannel` for `getChannelAuthenticated`, using the server ID from `requireServerSettingsAccess` (which was already being called and returns the server object)

## Test plan

- [ ] Visit `/settings/:serverSlug/:channelSlug` for a **private** server as the owner — page should load instead of 404
- [ ] Visit the same page for a **public** server — should still work as before
- [ ] Visit the settings page as a non-member — should redirect to `/channels/:serverSlug` (unchanged behavior from `requireServerSettingsAccess`)

### DIFF
--- a/harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx
+++ b/harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx
@@ -1,5 +1,5 @@
 import { notFound } from 'next/navigation';
-import { getChannel, getChannels } from '@/services/channelService';
+import { getChannelAuthenticated, getChannels } from '@/services/channelService';
 import { ChannelSettingsPage } from '@/components/settings/ChannelSettingsPage';
 import { requireServerSettingsAccess } from '@/app/settings/[serverSlug]/settings-access';
 import { ChannelType } from '@/types';
@@ -10,9 +10,9 @@ interface PageProps {
 
 export default async function SettingsPage({ params }: PageProps) {
   const { serverSlug, channelSlug } = await params;
-  await requireServerSettingsAccess(serverSlug);
+  const server = await requireServerSettingsAccess(serverSlug);
 
-  const channel = await getChannel(serverSlug, channelSlug);
+  const channel = await getChannelAuthenticated(server.id, channelSlug);
   if (!channel) notFound();
 
   const allChannels = await getChannels(channel.serverId);
--- a/harmony-frontend/src/__tests__/guest-server-name-sanitization.test.tsx
+++ b/harmony-frontend/src/__tests__/guest-server-name-sanitization.test.tsx
@@ -11,6 +11,7 @@ jest.mock('@/hooks/useAuth', () => ({
 
 jest.mock('next/navigation', () => ({
   usePathname: () => '/c/admin/general',
+  useRouter: () => ({ push: jest.fn(), replace: jest.fn() }),
 }));
 
 describe('guest server name sanitization', () => {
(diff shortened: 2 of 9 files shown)


harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx
```diff
@@ -10,9 +10,9 @@ interface PageProps {
 
 export default async function SettingsPage({ params }: PageProps) {
   const { serverSlug, channelSlug } = await params;
-  await requireServerSettingsAccess(serverSlug);
+  const server = await requireServerSettingsAccess(serverSlug);
 
-  const channel = await getChannel(serverSlug, channelSlug);
+  const channel = await getChannelAuthenticated(server.id, channelSlug);
```
Comment (acabrera04) [inline harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx:15]: This fixes the initial SSR channel lookup, but the rendered settings UI still calls server actions that resolve the same private server/channel through the old public helpers. `saveChannelSettings`, `fetchAuditLog`, `deleteChannelAction`, `resolveChannelAndServer`, and `updateChannelVisibility` all call `getChannel(serverSlug, channelSlug)` and/or `getServer(serverSlug)`, and those hit the public `/servers/:slug` path that returns `null` for private servers. After this page loads for a private server, saving overview changes, loading the visibility audit log, deleting, and member management still fail with `Channel not found`/`Server not found`. Please thread the authenticated server/channel resolution through the settings actions as well, or add an authenticated resolver there, so the private-server settings page is actually usable.

acabrera04 (changes requested):
- [x] Logic over style: The new authenticated page lookup is the right direction, but the settings flow is still incomplete for private servers.
- [x] Security first: The new lookup keeps authorization on `requireServerSettingsAccess` plus `channel.getChannel`; no new client-trusted ID path stood out.
- [x] Architectural alignment: Uses existing service/tRPC patterns rather than adding a parallel backend path.
- [ ] Issue completion: The private-server settings page can load, but core settings actions still resolve through public server/channel helpers and will fail for the same private-server case.
- [x] No nitpicking: Feedback is limited to a behavior regression in the submitted flow.
- [x] Avoid repetition: There were no existing comments or review threads on this PR.
- [x] Iterative reviews: No prior review from me to re-check.
- [x] Prevent CI failures: Static inspection did not show syntax/import issues in the changed files.

I attempted focused verification with `npm test -- --runTestsByPath src/__tests__/channelService.test.ts` and targeted lint, but this worktree does not have frontend dependencies installed (`jest`/`eslint` were not found).

harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx
Comment (declanblanc) [reply harmony-frontend/src/app/settings/[serverSlug]/[channelSlug]/page.tsx:15]: Fixed in 9cc3151a. All settings server actions (saveChannelSettings, fetchAuditLog, resolveChannelForSeo, deleteChannelAction, resolveChannelAndServer, fetchServerMembersForChannel, and updateChannelVisibility) now use getServerAuthenticated + getChannelAuthenticated instead of the public getChannel/getServer helpers, so the full settings UI works for private servers.

